PR #1, opened 2014-03-07
devices: LoopDevice is partitionable

fabiand:
Thus we set the attribute correctly

Signed-off-by: Fabian Deutsch <fabiand@fedoraproject.org>


vpodzime:
@dwlehman, could you please provide some more information about the "only when flags.installer_mode is False"? I'm afraid that the installation process is what @fabiand is targetting here. Or not, @fabiand?

fabiand:
@vpodzime actually it was for testing back then. We investigated using blivet, but as you know we settled with using anaconda in it's whole entirety. Means we can close this bug from my side if there are objections.

PR #16, opened 2014-12-19
Since py 2.6 doesn't support dict comprehension, I've switched to compat...

jcpunk:
Since py 2.6 doesn't support dict comprehension, I've switched to compatible syntax here.

If the deps can be met, this allows blivet to run on EL6 which is somethign I've active interest in.

Someone should double check the test suite, it worked for me but.....


vpodzime:
Looks good to me.

mulkieran:
Pushed.

PR #27, opened 2015-02-19
Move the Blivet class into its own module

vpodzime:
The __init__.py file should not contain such a big class with a lot of code. And
importing anything from the 'blivet' package shouldn't require importing all the
things that the 'Blivet' class requires.


dwlehman:
Looks good to me.

PR #28, opened 2015-02-19
Port blivet to libblockdev

vpodzime:
Replace majority of functions from blivet/devicelibs/ with calls to the *libblockdev* library. Also remove tests for such replaced functions and Requires: for tools we no longer directly need and replace calls to pyblock and volume_key-python with calls to libblockdev so that we can drop these dependencies.

``tests/pylint/runpylint.sh`` and ``make test`` are both okay with these changes.


dwlehman:
Looks good to me. Push it whenever you're ready.

PR #29, opened 2015-02-26
Switch translations to use Zanata

bcl:
Use https://fedora.zanata.org/project/view/blivet for translations. Also
check to make sure zanata-python-client is installed before po-pull


vpodzime:
Looks good to me.

PR #32, opened 2015-03-03
Master services

mulkieran:
The primary purpose of this patch set is just to remove services infrastructure from blivet devices and formats,
since it is no longer necessary.

Some services functions had been copied from some packages functions which were buggy, so I thought I might as well fix those, too, and any that were buggy like them. There weren't any other similar bugs that I could find, but there were a couple methods  that yielded to simplification. I thought I should test changes, so it seemed like a good idea to split out devices_tests.py since it was growing over long.


blivet/devices/storage.py
```diff
@@ -148,46 +148,15 @@ def __str__(self):
...
-        """
-        services = super(StorageDevice, self).services
-        services.extend(self.format.services)
-        for parent in self.parents:
-            for service in parent.format.services:
-                if service not in services:
-                    services.append(service)
-
-        return services
+        return packages + [p for p in self.format.packages if p not in packages]
```
vpodzime: Any reason for using + and list comprehension instead of extend() and generator in this particular case?

vpodzime:
Apart from the two comments/suggestions this looks good to me.

blivet/devices/storage.py
mulkieran: Not a compelling one. I was thinking about induction and emphasizing that the expression "p not in packages" did not require the updated version of packages. But I don't make use of that fact any where else. Unless anybody has an objection I'll use the extends version instead.

mulkieran:
Pushed.

PR #33, opened 2015-03-04
Allow passing fractional MiB values to vgcreate -s option

TimothyAsirJeyasing:
Currently blivet expects pesize parm of the vgcreate function
to be in MiB. However it does not accept any floating values.
This limits passing any values which are KiBs for pesize to create a vg.
This patch fixes the issue by accepting floating values also.

Bug url: https://bugzilla.redhat.com/show_bug.cgi?id=1198568

Signed-off-by: Timothy Asir Jeyasingh <tjeyasin@redhat.com>

### DIFF
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -339,7 +339,7 @@ def pvinfo(device=None):
 def vgcreate(vg_name, pv_list, pe_size):
     argv = ["vgcreate"]
     if pe_size:
-        argv.extend(["-s", "%sm" % pe_size.convertTo(spec="mib")])
+        argv.extend(["-s", "%sm" % pe_size.convertTo(spec="KiB")])
     argv.extend(_getConfigArgs())
     argv.append(vg_name)
     argv.extend(pv_list)
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -339,7 +339,7 @@ def pvinfo(device=None):
 def vgcreate(vg_name, pv_list, pe_size):
     argv = ["vgcreate"]
     if pe_size:
-        argv.extend(["-s", "%sm" % pe_size.convertTo(spec="KiB")])
+        argv.extend(["-s", "%sk" % pe_size.convertTo(spec="KiB")])
     argv.extend(_getConfigArgs())
     argv.append(vg_name)
     argv.extend(pv_list)
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -339,7 +339,7 @@ def pvinfo(device=None):
 def vgcreate(vg_name, pv_list, pe_size):
     argv = ["vgcreate"]
     if pe_size:
-        argv.extend(["-s", "%dm" % pe_size.convertTo(spec="mib")])
+        argv.extend(["-s", "%sm" % pe_size.convertTo(spec="mib")])
     argv.extend(_getConfigArgs())
     argv.append(vg_name)
     argv.extend(pv_list)


dwlehman:
Does lvm handle the Decimal values correctly? I wonder if it might be better to convert to KiB in that function as appropriate.

mulkieran:
It looks like libblockdev version of vgcreate uses bytes? So maybe converting to bytes and using "b" is the way to go.

dwlehman:
That would certainly put the issue to rest.

vpodzime:
Well, this already is resolved on master by using libblockdev. But this PR targets the f21-branch

TimothyAsirJeyasing:
Yes LVM  handles the decimal values when we pass like .25m

mulkieran:
Yes, this is for f21 branch. But my theory is that this code in libblockdev's vgcreate method:

```
argv[0] = "vgcreate";
argv[1] = "-s";
argv[2] = g_strdup_printf ("%"G_GUINT64_FORMAT"b", pe_size);
argv[3] = name;
```
indicates that lvm can accept bytes specification for pe size. Otherwise, some failure would already have been noticed. So why not just give it bytes in f21 branch? That way, it has to make an active effort to lose precision, whereas if it interprets the input
string as a floating point number it loses precision without any extra effort.

It strikes me that passing the value with the smallest units is probably often a good bet for these
external tools anyway.

mulkieran:
How about:

```
argv.extend(["-s", "%sb" % pe_size.convertTo(spec="b")])
```

dwlehman:
That looks good to me, except I might use %d.

vpodzime:
Yeah, "%d" would be nicer. For vgcreate this should work, however, it doesn't work for lvcreate because that requires a multiple of 512 if 'b' is used as a unit (even though it then rounds it to a multiple of extents).

mulkieran:
Actually, AFAICT vgcreate requires a multiple of 512 bytes, just like lvcreate, if you specify bytes.
But it's totally happy with 0.37K.

It looks to me like lvcreate and vgcreate are both better off using these fractional quantities in both libblockdev/devicelibs to get the values past the lvm censors.

The only other sensible thing is to have libblockdev/devicelibs provide the rounding unit, e.g., 512, to devices somehow, and let them do the rounding. But it would have to be done at the call site. Extending the Size functions convertTo() and roundToNearest() to make this happen would be easy, though.

dwlehman:
Good grief. I foolishly assumed "b" was for "bytes".

mulkieran:
I think "b" must be for bytes. If it meant something like MiB, then it wouldn't compalin that 2b is not a multiple of 512 bytes.

TimothyAsirJeyasing:
It looks like lvm vgcreate command accepts physical extent size can be given as kilobytes up to terabytes only (values can be from 8 KB to 16 GB in powers of 2) and the default is megabytes (4MB) if no suffix present.

vpodzime:
From ``man vgcreate`` (on my F21 machine):
```
   -s, --physicalextentsize PhysicalExtentSize[bBsSkKmMgGtTpPeE]
          Sets  the  physical extent size on physical volumes of this volume group. 
          A size suffix (k for kilobytes up to t for terabytes) is optional, megabytes
          is the default if no suffix is present. For LVM2 format, the value must be
          a power of 2 of at least 1 sector (where the sector size  is  the  largest 
          sector size  of the PVs currently used in the VG) or, if not a power of 2,
          at least 128KiB.  For the older LVM1 format, it must be a power of 2 of
          at least 8KiB. The default is 4 MiB.
```

So it's even more complicated. I guess the "multiple of 512 bytes" requirement is caused by drives usually having 512 B sectors these days. However, generally the 'b' suffix works as bytes (e.g. in libblockdev). However, using KiBs and "k" seems a little bit safer.

TimothyAsirJeyasing:
Sure, How about changing into
argv.extend(["-s", "%sk" % pe_size.convertTo(spec="KiB")]) ?

TimothyAsirJeyasing:
Sorry, I haven't noticed, I will change to %sk

bcl:
Looks good to me.

dwlehman:
Manually squashed and pushed.

PR #34, opened 2015-03-06
Makefile rc-release target and associated support

bcl:
This allows us to make scratch rpms with a version-release that allows it to be used before the next official release. Useful for nightly testing and such.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,14 @@ PKGNAME=blivet
 SPECFILE=python-blivet.spec
 VERSION=$(shell awk '/Version:/ { print $$2 }' $(SPECFILE))
 RELEASE=$(shell awk '/Release:/ { print $$2 }' $(SPECFILE) | sed -e 's|%.*$$||g')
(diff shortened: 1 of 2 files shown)


dwlehman:
My brain is too fried to conduct a thorough review, but this looks good in principle.

vpodzime:
Other than the minor commens above this looks good to me.

clumens:
This all looks solid to me, though now I am starting to wonder:  Do we want to find a way to share makebumpver and Makefile targets?

mulkieran:
Yes, we would really like to share makebumpver. Just so you know, the version in different packages have diverged _a lot_, since they migrated. So it would be kind of challenging.

bcl:
I don't think so. They're similar, but still different enough that any effort to make them common would just add a 3rd place to edit things.

clumens:
Perhaps pocketlint (which is right now just a place where I'm moving all our pylint stuff to be shared) could also function as a place to store makebumpver?  In that case, do I want to rename it to something more generic?  Though that is a very good name.  Sharing Makefile stuff - okay, maybe that's too hard.

PR #35, opened 2015-03-07
Fix dd wipe call

alexanderlaw:
When running test script:

```python
b = blivet.Blivet()
b.reset()
disk = b.devicetree.getDeviceByName('sda')
part = b.newPartition(parents=[disk], size=10*1024*1024, grow=False)
action = blivet.deviceaction.ActionCreateDevice(part)
b.devicetree.registerAction(action)
blivet.partitioning.doPartitioning(storage=b)
b.devicetree.processActions()
```
I see in the log:

INFO:program:Running... dd if=/dev/zero of=/dev/sda bs=512 Б seek=22022144 count=2048
INFO:program:stderr:
INFO:program:dd: invalid number '512 \320\221'
DEBUG:program:Return code: 1

Where 'Б' is first letter in Russian 'Байт' (Bytes).
It seems, blivet should pass sector size as a number, not as a human-readable string.

BTW, result code is not checked so you can see the error only when increasing log verbosity.

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -539,7 +539,7 @@ class PartitionDevice(StorageDevice):
         # Erase 1MiB or to end of partition
         count = min(int(Size("1 MiB") // bs), part_len)
 
-        cmd = ["dd", "if=/dev/zero", "of=%s" % device, "bs=%s" % bs,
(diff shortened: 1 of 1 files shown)


mulkieran:
Thanks for the patch. This looks exactly right to me. I'll merge it early next week unless anybody has an objection.

mulkieran:
Pushed.

PR #37, opened 2015-03-07
Make size._parseSpec a public method.

mulkieran:
This will allow it to be used by client code to prepare for a switch
to new implementation of Size which does not take an arbitrary string
specification as a constructor argument.


mulkieran:
Pushed.

PR #38, opened 2015-03-09
Robustify PartitionDevice._wipe() method.

mulkieran:
Previously, if block size is larger than 1 MiB, nothing gets erased.
Now, the smallest number of blocks that erases at least 1 MiB is wiped.


bcl:
So far I haven't seen any disks with 1MiB or greater block sizes, but may as well be prepared. While you're at it, why don't you convert all the dd args to %d and int() since they all need to be numbers anyway.

mulkieran:
Pushed.

PR #39, opened 2015-03-10
Make sure the device is setup before formatting it (#1196397)

bcl:
Reusing an existing LV wasn't working because it was not setup before
formatting.


bcl:
Pushed to f22-branch

PR #40, opened 2015-03-10
Master misc

mulkieran:
Minor fixes that I stumbled on while working on tasks branch. Since none of them are really related to tasks branch directly I'm putting them up for consideration now. Fixes two bugs, one in FS.testMount() and one in a TmpFS test case. They should all be pretty easy to ack, I would think.


bcl:
These all look good.

mulkieran:
pushed.

PR #41, opened 2015-03-10
Allow to set partition flags

alexanderlaw:
This patch allows to set partition flags supported by parted such as 'hidden', 'boot', 'bios_grub'...
('bootable' argument in fact does nothing if the partition is not mounted to '/boot'.)
Flags are passed in a dictionary:
```python
partiition = b.newPartition(parents=[disk], size=10*1024*1024, grow=False, partFlags = {'bios_grub': True})
```

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -65,7 +65,7 @@ class PartitionDevice(StorageDevice):
                  size=None, grow=False, maxsize=None, start=None, end=None,
                  major=None, minor=None, bootable=None,
                  sysfsPath='', parents=None, exists=False,
-                 partType=None, primary=False, weight=0):
+                 partType=None, primary=False, weight=0, partFlags=None):
         """
             :param name: the device name (generally a device node's basename)
             :type name: str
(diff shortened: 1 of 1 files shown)


mulkieran:
Thanks for the patch. It's not clear what is your overall goal here, i.e., what problem you are solving. It might make sense to start an Issue on blivet to get this across better.

alexanderlaw:
My goal is to create boot (bios_grub) partition on a GPT disk. Okay, I'll start an issue for it.

dwlehman:
Check out the "biosboot" format class in blivet.formats.biosboot -- creating such a format using blivet actions should set you up.

alexanderlaw:
Yes, biosboot works indeed, thanks. But then I still doesn't understand how to create boot (PARTITION_BOOT) partition. Do I need to add a new DeviceFormat-based class simlar to BIOSBoot to create one?

alexanderlaw:
It seems, that EFIFS is intented for that type of partition, but to get the 'boot' flag I need to add to it's definiton:
+    partedFlag = PARTITION_BOOT
+    partedSystem = None

dwlehman:
biosboot is for BIOS boot on gpt disk. (bios_grub flag)

efifs is for EFI boot on gpt disk. ("EFI System Partition")

dwlehman:
See Blivet.doIt for how boot flags are handled during OS installation via anaconda.

alexanderlaw:
Thanks for info. I use openlmi-storage and thus restricted by it's entities and methods in some way.
Nevertherless, I see that my original proposition is not line with the existing implementation, so the pull request can be closed.

PR #43, opened 2015-03-11
Master tasks misc

mulkieran:
These are some more trivial cleanups. They only do two things:
1) Use the class hierarchy correctly wherever possible, making it clear when superclass method is
being completely disregarded rather than just extended.
2) Make some things easier to read at a glance, hence easier to comprehend when considering the
larger picture.

They ought to be very easy to ack.


vpodzime:
These all look good to me.

mulkieran:
Pushed.

PR #44, opened 2015-03-12
Set format options for mkfs

TimothyAsirJeyasing:
Currently format options can be specified only for doFormat
function. This patch provides options to set the format options.

Usage ex:-
format = blivet.formats.getFormat("xfs", device=lv.path)
format.formatOptions = ["-f", "-K", "-i", "size=256", "-d", ...]
blivetEnv.formatDevice(lv, format)
blivetEnv.doIt()

bug-url: https://bugzilla.redhat.com/show_bug.cgi?id=1180221
Signed-off-by: Timothy Asir Jeyasingh <tjeyasin@redhat.com>

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -69,6 +69,7 @@ class FS(DeviceFormat):
     _fsck = ""                           # fs check utility
     _fsckErrors = {}                     # fs check command error codes & msgs
     _infofs = ""                         # fs info utility
+    _formatOptions = []
     _defaultFormatOptions = []           # default options passed to mkfs
     _defaultMountOptions = ["defaults"]  # default options passed to mount
     _defaultCheckOptions = []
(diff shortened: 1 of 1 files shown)


mulkieran:
Please also consider an equivalent patch for clone at: https://github.com/mulkieran/blivet/tree/master-tasks, as that refactors a good deal of the mkfs functionality into a separate task, and may be applied soon. Look at blivet/tasks/fsmkfs.py and maybe blivet/tasks/fsmount.py for an example.

dwlehman:
I would like to see this applied to all of the classes in blivet.formats so we can have a shared mechanism for specifying options to be passed to both create and setup. It could be useful for swap and luks in addition to the filesystems.

TimothyAsirJeyasing:
Yes sure, I will check it out and update the patch

clumens:
I'm currently working on this as part of adding support for letting you specify arbitrary mkfs options in kickstarts, so don't worry about it if you've not gotten started.

mulkieran:
This PR is defunct. @TimothyAsirJeyasing I've cc'd you on the relevant bzs.

PR #45, opened 2015-03-12
Split up DeviceTree

dwlehman:
DeviceTree has gotten to be a huge class. This patch set moves methods used to populate the DeviceTree into blivet.populator.Populator. It also moves action list/queue management into blivet.actionlist.ActionList.

Notes on ActionList:
- I kept processActions and findActions wrappers in DeviceTree to preserve API, but I plan to remove them
  both in blivet-2.0.
- I wanted to call it ActionQueue, but it's not a proper queue. Could/should I call it that anyway?

Notes on Populator:
- I'm not crazy about the name. Previously it was discoverer.Discoverer. Any opinions on which is better?
- It's a little goofy that Populator has a DeviceTree instance as an attribute and vice-versa. I'd be open to
  fixing that in a better way if people are uneasy about, or disgusted by, it.


mulkieran:
Generally speaking, I'm all in favor of better encapsulation. Since I was not familiar with the structure to begin with I can't do much of a critique on exactly how it's separated.

Maybe ActionList should be called something that doesn't make a statement about the data structure, in case sometime it becomes a DAG or something. Conceptually, it is a DAG, because of the ordering, so why not ActionDAG.

dwlehman:
The problem is there isn't much common language that can't also be interpreted as making a statement about the data structure.

How about this: ActionListButListInANonDataStructureSense

This is an irritating topic.

vpodzime:
I don't like the ``ActionDAG`` name because I don't know the acronym ``DAG`` (had to ask @dwlehman about it) and I think many people would have the same issue with it. ``ActionList`` feels just natural and since vast majority of people doesn't really have to bother with the implementation details, it should be good enough for everybody.

vpodzime:
Other than that one line that really could/should be dropped, this all looks good to me.

PR #46, opened 2015-03-13
Prevent issues caused by passing XFS/BTRFS max sizes to libblockdev (#1200812)

vpodzime:
These should be max 64-bit unsigned integer value which is 16 EiB - 1.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -1115,7 +1115,7 @@ class BTRFS(FS):
     _supported = True
     _packages = ["btrfs-progs"]
     _minSize = Size("256 MiB")
-    _maxSize = Size("16 EiB")
+    _maxSize = Size("15 EiB")
     # FIXME parted needs to be taught about btrfs so that we can set the
     # partition table type correctly for btrfs partitions
     # partedSystem = fileSystemType["btrfs"]
@@ -1254,7 +1254,7 @@ class XFS(FS):
     _modules = ["xfs"]
     _labelfs = fslabeling.XFSLabeling()
     _defaultFormatOptions = ["-f"]
-    _maxSize = Size("16 EiB")
+    _maxSize = Size("15 EiB")
     _formattable = True
     _linuxNative = True
     _supported = True
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -936,16 +936,15 @@ class LVRequest(Request):
         self.base = int(lv.vg.align(lv.req_size, roundup=True) // lv.vg.peSize)
 
         if lv.req_grow:
-            limits = [int(l // lv.vg.peSize) for l in
-                        (lv.vg.align(lv.req_max_size),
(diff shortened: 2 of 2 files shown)


jkonecny12:
It's working correctly for me now.

mulkieran:
Comments in bug.

vpodzime:
Any reason for not commenting here? I really don't like having to copy&paste the bug number and reading something that is not closely related to the bug in bugzilla.

vpodzime:
@mulkieran posted to bugzilla:
> In bytesize, it will not be possible, or at least it will be discouraged, to add numbers to sizes (or subtract numbers from sizes). That's not the real problem with this patch, though.
Exactly. That is to be resolved once we switch to bytesize.

> *) You haven't confirmed that xfs or btrfs are actually constrained to this number...maybe they store all values as multiples of sectors. We should make sure that we know the _actual_ limits of these two, probably by asking, before we make a patch based on their actual limits.
Or we could just not waste time by doing that since it really doesn't matter at all now and wont' matter in near or even distant future.

> *) This doesn't fix the problem of the way that libblockdev represents integers, even in this particular instance. It is true that this particular test case will no longer raise that particular exception. But, it will do the wrong thing, because the integer will overflow in bd_lvm_round_size_to_pe, since it is rounding up. A quick glance seems to indicate that the result of rounding 2^64 - 1 up to the nearest pe_size boundary will return pe_size, which is wrong.
That is a valid point that has to be addressed.

vpodzime:
New version of the patch proposed for merge. This time with the values agreed on during the meeting.

mulkieran:
Wikipedia claims that 8EiB is the de facto limit for btrfs on Linux systems and 8 EiB is a more plausible number than 15 EiB (because it is a power of 2, like all our other max sizes). It would be just as good for both.

Since 15 EiB is certainly a multiple of sector size, though, the choice is not such a big deal.

Since I think it is not the best choice, I'm not giving it an ACK. But, since I think it's not a huge deal, I'm not NACKing it neither. If someone else wants it to go in as is, I'll be OK with it. (Still working out how to use ACKs and NACKs, but this approach makes sense to me.)

vpodzime:
Added one more commit, that implements:
> 2) protect this call by limiting the size to the maximum size for an LV

mulkieran:
It looks to me like: https://github.com/rhinstaller/blivet/pull/50 makes this PR entirely unnecessary.

vpodzime:
> It looks to me like: #50 makes this PR entirely unnecessary.

No it doesn't. At least the the second commit that limits the check for maximum LV growth to maximum LV size is still valid.

dwlehman:
I definitely like the max lv size being factored into the growth limit calculation.

mulkieran:
Without commit 1 and with commit 2 message amended a bit this looks fine to me.

vpodzime:
Deprecated by PR #50

PR #47, opened 2015-03-16
Dynamic mountpoint detection

vojtechtrefny:
Dynamic detection of mountpoints for existing devices.

blivet.mounts.MountCache parses /proc/mounts and /proc/self/mountinfo to get information about mounted devices. See https://github.com/rhinstaller/blivet/issues/13 for more information.

### DIFF
--- a/blivet/mounts.py
+++ b/blivet/mounts.py
@@ -1,0 +1,137 @@
+# mounts.py
+# Active mountpoints cache.
+#
+# Copyright (C) 2015  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+# Red Hat Author(s): Vojtech Trefny <vtrefny@redhat.com>
+#
+
+from . import util
+
+import logging
+log = logging.getLogger("blivet")
+
+class MountsCache(object):
+
+    def __init__(self):
+        self.mountsHash = 0
+        self.mountpoints = {}
+
+        # new device manually added to cache since last check
+        self.newDevice = False
+
+    def add(self, devspec, subvolspec=None):
+        """ Add device to cache
+
+            :param devscpec: device specification, eg. "/dev/vda1"
+            :type devspec: str
+            :param subvolspec: btrfs subvolume specification, eg. ID or name
+            :type subvolspec: str
+
+        """
+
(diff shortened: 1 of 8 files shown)


mulkieran:
A couple things that strike me at first glance are:
* I think mountCache should be in FS not DeviceFormat. AFAICT, the only thing about mounting in DeviceFormat is mountable, and there is a ton of stuff about mounting in FS.
* When tasks infrastructure goes in, MountsCache should be factored out into an always available Task and the rest of it. Please take a look at https://github.com/mulkieran/blivet/tree/master-tasks and work out how you would do that, and send me a link for that branch. A PR would probably just make the list too busy.
* Can you come up with a way to test these changes?

- mulhern

vpodzime:
> I think mountCache should be in FS not DeviceFormat. AFAICT, the only thing about mounting in DeviceFormat is mountable, and there is a ton of stuff about mounting in FS.

If I understand it correctly, having mountCache in FS not in DeviceFormat would require all the code that tries to get/set it on a ``device.format`` check if that format is a ``FS`` first. I think not having to do this checks is a nice benefit of having it one level higher in the class hierarchy.

> When tasks infrastructure goes in, MountsCache should be factored out into an always available Task and the rest of it. Please take a look at https://github.com/mulkieran/blivet/tree/master-tasks and work out how you would do that, and send me a link for that branch. A PR would probably just make the list too busy.

I think the priority of this depends on in which order we want to merge this and the tasks infrastructure. If this is the first thing to go in, we should focus on merging this and then on "rebasing" the tasks infrastructure.

vpodzime:
This looks good to me. I'm not giving ACK right now because I think this needs some more attention of other members of the team.

blivet/util.py
```diff
@@ -357,6 +358,18 @@ def insert_colons(a_string):
     else:
         return suffix
 
+def md5_file(filename):
+
+    f = open(filename, "rb")
```
mulkieran: We usually use the file as a context manager these days:

```
md5 = hashlib.md5()
with open(filename, "rb") as f:
    # do your stuff
```

blivet/mounts.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def _resolveFormat(self, devspec, subvolspec=None):
+
+        for fmt in self.mountpoints.keys():
+            if fmt[0] == devspec:
+                if not fmt[1]:
+                    return fmt
+
+                elif fmt[1] == subvolspec:
+                    return fmt
```
mulkieran:
```
return next((f for (d, s) in self.mountpoints.keys() if d == devspec and (not s or s == subvolspec)), None)
```
should do the same thing and also explicitly returns None.

This method is probably worth a doc string.

mulkieran:
There's a method, util.get_mount_paths() which blivet does not use, but anaconda does use. Would it be possible to use your MountsCache to re-implement that method as well? It seems like that method is duplicating the reading and using the information that MountsCache should have.

blivet/mounts.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def getMountpoint(self, devspec, subvolspec=None):
+        """ Get mountpoint for selected device
+
+            :param devscpec: device specification, eg. "/dev/vda1"
+            :type devspec: str
+            :param subvolspec: btrfs subvolume specification, eg. ID or name
+            :type subvolspec: str
+            :returns: mountpoint (path)
+            :rtype: str
```
mulkieran:
```
:rtype: str or NoneType
```

blivet/mounts.py
```diff
@@ -0,0 +1,137 @@
...
+
+        """
+
+        self._cacheCheck()
+
+        if (devspec, subvolspec) not in self.mountpoints.keys():
+            return None
+
+        else:
+            return self.mountpoints[(devspec, subvolspec)]
```
mulkieran:
```
return self.mountpoints.get((devspec, subvolspec))
```
can replace the if statement entirely.

PR #48, opened 2015-03-16
Master pointless

mulkieran:


clumens:
Can you please commit the first two patches to pocketlint as well?  You should have commit access there, but if not I'll add it.  Eventually I'm going to have patches to get blivet using pocketlint, but I'm not quite there yet.

mulkieran:
pushed...also to pocketlint.

PR #50, opened 2015-03-17
Master round

mulkieran:


dwlehman:
I like this because it saves us from having to artificially change the fs max size values.

mulkieran:
Pushed.

vpodzime:
Just for the record:
I still think this is not a good thing to do because the logic of such calculations should stay in the code that is related. If we want to do this, we should do the same with MD RAID superblock size calculations, LVM thin provisioning calculations, etc. These all have the same potential for similar issues. But that doesn't make much sense, right?

One thing as an example -- I think rounding size smaller than PE size to a multiple of extents and trying to round it down shouldn't result in 0 because we still need to put that bytes somewhere. So combination of size smaller than PE size and roundup=False should result in rounding up, I think. Does this logic fit into the ``size.roundToNearest``? No. Does it fit into the ``blockdev.lvm_round_size_to_pe``? Yes.

And there may be (and possibly will be) more glitches like this in the future.

PR #51, opened 2015-03-17
Miscellaneous patches from uevents-branch

dwlehman:
These are mostly very small, focused changes that either right old wrongs (eg: re-raising DeviceCreateError) or clean things up in various ways.


bcl:
Looks good other than those few comments.

mulkieran:
This seems fine to me.

PR #52, opened 2015-03-18
Catch GLib.GError in places where we catch StorageError (#1202505)

vpodzime:
Previously, all errors that could have been raised from running storage actions
from blivet were inherited from the StorageError class. This no longer holds
with libblockdev being used because libblockdev raises exceptions through GI
introspection and thus as GLib.GError. In places we catch the all-covering
StorageError we should also catch GLib.GError and in some places we should only
catch GLib.GError if a libblockdev call is the only thing in the try-except
block.

**This is a temporary fix until libblockdev adds some better exception hierarchy into it's GI overrides.**

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -33,6 +33,7 @@ from .partitioning import TotalSizeSet
 from .partitioning import doPartitioning
 from .size import Size
 
+from gi.repository import GLib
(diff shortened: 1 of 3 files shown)


vpodzime:
For master all the added lines should be changed to GLib.Error instead of GLib.GError

vpodzime:
Pushed

PR #53, opened 2015-03-18
F22 swap prio

vpodzime:
Don't specify priority in fstab if -1 (default) is used

For some reason, such fstab line doesn't work and it's better not to rely on -1
meaning the default forever.

Related: rhbz#1202505


AdamWill:
Tested that this does fix the live installer crash case of RHBZ#1202505.

vpodzime:
Pushed.

PR #54, opened 2015-03-18
Dynamic mountpoint detection -- new version

vojtechtrefny:
New version of previous pull request -- https://github.com/rhinstaller/blivet/pull/47

Changes:
 - MountsCache is now a singleton
 - It is not neccessary to add devices to cache from devicetree
 - Opening files using "with"
 - One line return in MountsCache.getMountpoint
 - tests

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -39,6 +39,7 @@ from ..size import Size, ROUND_UP, ROUND_DOWN, unitStr
 from ..size import B, KiB, MiB, GiB, KB, MB, GB
 from ..i18n import _, N_
 from .. import udev
+from ..mounts import mountsCache
 
 import logging
 log = logging.getLogger("blivet")
@@ -110,7 +111,6 @@ class FS(DeviceFormat):
         # filesystem size does not necessarily equal device size
         self._size = kwargs.get("size", Size(0))
         self._minInstanceSize = Size(0)    # min size of this FS instance
-        self._mountpoint = None     # the current mountpoint when mounted
 
         # Resize operations are limited to error-free filesystems whose current
         # size is known.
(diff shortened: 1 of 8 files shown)


blivet/formats/fs.py
```diff
@@ -1604,14 +1615,14 @@ def _setOptions(self, options):
 
     @property
     def free(self):
-        if self._mountpoint:
-            # If self._mountpoint is defined, it means this tmpfs mount
+        if self.systeMountpoint:
```
mulkieran: Watch out for spelling mistakes!..."systeMountpoint". pylint will definitely catch these kinds of things.

blivet/formats/fs.py
```diff
@@ -582,6 +582,13 @@ def loadModule(self):
         # also need to update the list of supported filesystems.
         update_kernel_filesystems()
 
+    @property
+    def systemMountpoint(self):
+        if not self.exists:
+            return None
+
+        return mountsCache.getMountpoint(self.device)
```
mulkieran: I would like it slightly better if this had a comment/docstring, if the final line was:

```
return mountsCache.getMountpoint(self.device, getattr(self, "subvolspec", None))
```
and the overriding method in BTRFS class went away.

mulkieran:
Other than minor inline comments, that looks good to me. And a very nice improvement to our filesystem code, too.

vojtechtrefny:
I made few changes based on comments by mulkieran and bcl but after force pushing them most of the comments disappeared...

mulkieran:
Still looks good to me. FWIW, I ran a quick profile on the md5_file() method on my /proc/mounts and it's the read() method that chews up the bulk of the time, with updating the hash a distant second. And if we check whether there has been a change this way we just can't get away from reading those bits. Reading it in smaller chunks makes no difference...reading it in one big step doubles the time it takes.
Hoping that under normal use this change makes no perceptible difference (fingers crossed).

vpodzime:
Other than the neat-pick above this still looks good to me.

vojtechtrefny:
Docstring added. Pushed.

PR #57, opened 2015-03-19
Fix name of the keyword argument for luks_add_key

vpodzime:
It should be 'npass' not 'new_passphrase'.


vpodzime:
I think this should also go to the f22-branch althought it is basically impossible to hit it from the installer (we don't allow adding multiple LUKS passphrases).

vpodzime:
Added one more similar change to the commit that is needed.

jsafrane:
It looks fine from OpenLMI point of view, our LUKS tests pass with the patch.

bcl:
Looks good.

vpodzime:
Rebased to the current master.

PR #58, opened 2015-03-19
A few miscellaneous patches

mulkieran:
I would wait until I had more, but until the first patch gets in all the blivet devices tests won't be run.


bcl:
These all look ok.

mulkieran:
Put an __init__.py... pushed. Other two withdrawn.

PR #59, opened 2015-03-20
Implement the __dir__ method of the LazyImportObject class

vpodzime:
This makes the 'TAB-TAB' completion in ipython and dir() calls work as if the
lazy-imported object was imported in a normal way.


M4rtinK:
Looks good to me, ACK!

clumens:
Looks good to jenkins, too.  Cool!

PR #61, opened 2015-03-31
Use LVname not VG-LVname when adding LVs

vpodzime:
The VG-LVname string is not a key in the dictionary constructed from the 'lvm
lvs' output and if by some chance (really high on Python 3) e.g. a thin pool is
requested to be added by a thin LV, using VG-LVname as the key raises a KeyError
while the info about the pool LV is in the dictionary as expected.

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -145,7 +145,7 @@ class DeviceTree(object):
     def lvInfo(self):
         if self._lvs_cache is None:
             lvs = blockdev.lvm_lvs()
-            self._lvs_cache = dict((lv.lv_name, lv) for lv in lvs) # pylint: disable=attribute-defined-outside-init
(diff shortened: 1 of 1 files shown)


vpodzime:
I think this is also worth porting to f22-branch since it is only based on coincidence whether it works or not without this change (the order of ``dict.values()`` to be more precise)

dwlehman:
Did this change with the move to libblockdev? It seems dangerous since it's fairly common for people (testers) to have multiple VGs with the same set of LV names.

vpodzime:
> Did this change with the move to libblockdev? It seems dangerous since it's fairly common for people (testers) to have multiple VGs with the same set of LV names.

Yes and yes. Updating the PR with a better fix -- using VGname-LVname for keys in the ``lvInfo`` dictionary as it was before the move to libblockdev.

vpodzime:
Pushed.

PR #62, opened 2015-04-01
Add back DeviceTree's methods and properties used from the outside

vpodzime:
These methods/properties were moved to Populator and ActionList classes, but to
keep the API we need to provide them in the DeviceTree class too.

Also make the 'dasd' property settable because that's what Anaconda does.


vpodzime:
Pushed.

PR #63, opened 2015-04-01
Bits needed to enforce sane disk selection in anaconda

dwlehman:
This goes with rhinstaller/anaconda#60

- move getRelatedDisks into blivet as suggested by both @vpodzime and @mulkieran
- tweak getDependentDevices to (optionally) work with hidden devices

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -361,23 +361,30 @@ class DeviceTree(object):
                              dryRun=dryRun,
                              callbacks=callbacks)
 
-    def getDependentDevices(self, dep):
+    def getDependentDevices(self, dep, hidden=False):
         """ Return a list of devices that depend on dep.
 
             The list includes both direct and indirect dependents.
 
             :param dep: the device whose dependents we are looking for
             :type dep: :class:`~.devices.StorageDevice`
+            :keyword bool hidden: include hidden devices in search
         """
         dependents = []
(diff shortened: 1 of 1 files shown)


blivet/devicetree.py
```diff
@@ -361,28 +361,48 @@ def processActions(self, dryRun=False, callbacks=None):
...
+
+        for device in devices:
+            log.debug("checking if %s depends on %s", device.name, dep.name)
             if device.dependsOn(dep):
                 dependents.append(device)
 
         return dependents
 
+    def getRelatedDisks(self, disk):
+        """ Return a list of disks related to disk by container membership.
```
mulkieran: Literally, it returns a set, not a list.

mulkieran: Also the docstring talks about "container membership" but given the definition of getDependentDevices it feels like getRelatedDisks is doing something far more general than anything to do with containers, per se.

mulkieran:
Other than the docstring, it looks fine to me.

blivet/devicetree.py
dwlehman: Unless I'm forgetting something, the only way for two devices to share a child is through container membership.

vpodzime:
Looks good to me. The ``:rtype: set of :class:`~.devices.StorageDevice``` would be nice, though.

dwlehman:
Improved docstring for getRelatedDisks

PR #64, opened 2015-04-01
Wrap keys() with a list so that the dictionary can be changed

M4rtinK:
In Python 3 the keys() dictionary method returns and iterator,
which we can't use if we want to modify the dictionary while iterating over it 
(it raises a RuntimeError). So wrap keys() with a list to prevent this from
happening.


M4rtinK:
Applied manually and pushed.

PR #66, opened 2015-04-02
Fix some Python 3 compatibility issues and add the python3 subpackage

vpodzime:
Fixes for Python 3 compatibility issues discovered while running tests with Python 3 including hundreds of installations with Python 3 Anaconda. With all that passing in the end, I decided to add the pieces that make sure the python3 subpackage is built too (tested in [Copr builds] (https://copr.fedoraproject.org/coprs/bkabrda/py3anaconda/builds/)).

**Disclaimer:** 
These patches don't try to fix anything else than Python 3 incompatibilities in the most straightforward way. Many changes are subject for future improvements and nicer solutions.


M4rtinK:
Rebased on Blivet master branch.

PR #67, opened 2015-04-02
F22 1208536

mulkieran:
Also for master.


vpodzime:
Other than the two neatpicks above this looks good to me. And I keep my fingers crossed in a hope this is going to be the last piece in the *F22 FW RAID Saga*

mulkieran:
The last two or three are new, but the first two are the original.

AdamWill:
FWIW I've tested this, with an updates.img containing these changes I can complete install to a RAID-0 set and boot to the installed system.

vpodzime:
This all look good to me.

PR #68, opened 2015-04-02
Use /sys to obtain sizes of existing devices

dwlehman:
parted.Device is quite heavy for just tracking device size. Due to its intended purpose being quite different from how blivet uses it, it becomes problematic in certain situations related to uevents, udev rules, and things like md arrays. Specifically, deleting a parted.Device involves closing a read/write file descriptor, which generates a "change" uevent, which triggers all sorts of udev rules. If the device whose parted.Device you just deleted happened to be a member of an md array, the result would be that the array would get activated. One way this can happen is by simply calling Blivet.reset on a system with an inactive md array. Obviously, this is unacceptable on the basis that storage libraries (or modules) must not activate system devices without having been instructed to do so.

This business of closing rw fds gets even less convenient when blivet begins listening for and reacting to uevents, so this is a prerequisite for that.


vpodzime:
Thanks for deeper investigation and testing. With the leftover removed this looks good to me.

dwlehman:
Added the following to the first patch:
```patch
commit e193a85e4b28899952c22440381f16819ea346ad
Author: David Lehman <dlehman@redhat.com>
Date:   Tue Apr 7 09:29:23 2015 -0500
 
    Update size info as needed after setting up any device.
 
diff --git a/blivet/devices/storage.py b/blivet/devices/storage.py
index 0897065..052b397 100644
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -379,9 +379,10 @@ class StorageDevice(Device):
     def _postSetup(self):
         """ Perform post-setup operations. """
         udev.settle()
-        # we always probe since the device may not be set up when we want
-        # information about it
-        self._size = self.currentSize
+        self.updateSysfsPath()
+        # the device may not be set up when we want information about it
+        if self._size == Size(0):
+            self.updateSize()
 
     #
     # teardown
```

vpodzime:
Just for the record -- still looks good to me.

PR #69, opened 2015-04-06
Master multiple mounts

bcl:
Fix dynamic mountpoint code to allow multiple mounts when the user specifies the mountpoint.

### DIFF
--- a/blivet/mounts.py
+++ b/blivet/mounts.py
@@ -19,7 +19,7 @@
 #
 # Red Hat Author(s): Vojtech Trefny <vtrefny@redhat.com>
 #
-
+from collections import defaultdict
 from . import util
 
 import logging
@@ -32,23 +32,25 @@ class MountsCache(object):
 
     def __init__(self):
         self.mountsHash = 0
-        self.mountpoints = {}
+        self.mountpoints = defaultdict(list)
 
-    def getMountpoint(self, devspec, subvolspec=None):
-        """ Get mountpoint for selected device
+    def getMountpoints(self, devspec, subvolspec=None):
+        """ Get mountpoints for selected device
 
             :param devscpec: device specification, eg. "/dev/vda1"
             :type devspec: str
             :param subvolspec: btrfs subvolume specification, eg. ID or name
             :type subvolspec: str
-            :returns: mountpoint (path)
-            :rtype: str or NoneType
+            :returns: list of mountpoints (path)
+            :rtype: list of str or empty list
 
+            .. note::
+                Devices can be mounted on multiple paths, and paths can have multiple
+                devices mounted to them (hiding previous mounts). Callers should take this into account.
         """
-
         self._cacheCheck()
 
-        return self.mountpoints.get((devspec, subvolspec))
+        return self.mountpoints[(devspec, subvolspec)]
 
     def isMountpoint(self, path):
         """ Check to see if a path is already mounted
@@ -57,13 +59,15 @@ class MountsCache(object):
         """
         self._cacheCheck()
 
-        return path in self.mountpoints.values()
+        return any(path in p for p in self.mountpoints.values())
 
     def _getActiveMounts(self):
         """ Get information about mounted devices from /proc/mounts and
             /proc/self/mountinfo
-        """
 
+            Refreshes self.mountpoints with current moutpoint information
+        """
+        self.mountpoints = defaultdict(list)
         for line in open("/proc/mounts").readlines():
             try:
                 (devspec, mountpoint, fstype, _options, _rest) = line.split(None, 4)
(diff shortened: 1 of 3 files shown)


mulkieran:
Could you piggyback a little patch on this that gets rid of the FIXME in FS.status definition? Cause it is definitely obsolete now.

mulkieran:
There is a disagreement between the MountsCache "one mountpoint per ..." and the util.get_mount_paths() method which clearly allows for multiple mountpoints. blivet does not use get_mount_paths(), but anaconda does.

I think that:
* MountsCache.getMountpoint() should be changed to MountsCache.getMountpoints(), since, as you say there may be multiple mountpoints. MountsCache.mountpoints should be a tuple to list map, instead of a tuple to str map, or something like that.
* FS.systemMountpoint should do something smart so it knows what mountpoint to prefer if it gets multiple mountpoints. I _believe_ that /proc/mounts entries are ordered so that the last mountpoint found is the most recently mounted...that would be most convenient.
* util.get_mount_paths can become:
```
def get_mount_paths(dev):
    return mounts.MountsCache.getMountpoints(dev, None)
```
or something similar.

I believe that it is also possible for two different devices to be mounted at a single mountpoint, so that one obscures the other. So, maybe checking isMountpoint() is not necessary unless you absolutely want to prevent two devices being mounted at the same point.

I think mount() should return when something like 

```
mountpoint in MountCache.getMountpoints() 
```
is True (becuase it is already mounted at mountpoint), but should raise an exception if

```
mountsCache.isMountpoint(mountpoint) and mountpoint not in MountCache.getMountpoints()
```
because trying to mount over a device that is already mounted at mountpoint (I'm assuming that we
don't want this).

bcl:
That all sounds like a good idea, but that's probably better handled by @vojtechtrefny as an enhancement to the mount cache. What I'm trying to do with this patch set is make sure that rawhide stays usable.

mulkieran:
Ok, I'm glad you like those suggestions in principle.

I'm not clear to me what the bug was that manifested originally...I'm guessing it was a problem in anaconda, since your blivet patch does not use the extra unmount parameter anywhere. So, I am somewhat in the dark.

I'm worried about these things:

* The additional parameter passed to unmount(). Nothing in your patch set actually invokes unmount() with a value for that parameter. I wish I understood the use case better.
* It seems like, if my suggestions above were implemented, systemMountpoint would always contain the last mountpoint used, so the necessity for a mountpoint parameter to unmount() would go away entirely.
* I'm not sure about the change in FS.mount(), because...

I _think_ that the reason the method just returns if status is True is because status is used as a (not very good) proxy for "is mounted at the given mountpoint" and if that is True then there is nothing to do.

isMountpoint(chrootedMountpoint) covers two possible cases, which I think should be handled differently, something like:

```
if mountsCache.isMountpoint(chrootedMountpoint): # something is mounted here
    if self.systemMountpoint == chrootedMountpoint:  # self mounted here
        return        # nothing to be done
    else:  # different filesystem mounted here
        raise FSError("Collision!")                                   
```

The way systemMountpoint is currently implemented it uses the last value read, overwriting any previous, so that should represent the most recently used mountpoint (under the assumption that /proc/mounts lists mountpoints in order).

bcl:
Gah. that didn't push like I expected. Ignore the first set, the second set are the correct ones.

I've converted things to handle multiple mounts, pjones says that /proc/mounts is in order,  so the last one is the visible one.

I think it's ok to request multiple mounts, I don't think that should raise an error.

blivet/mounts.py
```diff
@@ -32,29 +32,42 @@ class MountsCache(object):
...
+        """ Get mountpoints for selected device
 
             :param devscpec: device specification, eg. "/dev/vda1"
             :type devspec: str
             :param subvolspec: btrfs subvolume specification, eg. ID or name
             :type subvolspec: str
-            :returns: mountpoint (path)
-            :rtype: str or NoneType
+            :returns: list of mountpoints (path)
+            :rtype: list of str or empty list
```
mulkieran: Probably useful to include a note about the order that the mountpoints occur in the result list.

blivet/mounts.py
```diff
@@ -32,29 +32,42 @@ class MountsCache(object):
...
         """ Get information about mounted devices from /proc/mounts and
             /proc/self/mountinfo
-        """
 
+            Refreshes self.mountpoints with current moutpoint information
+        """
+        self.mountpoints = defaultdict(list)
         for line in open("/proc/mounts").readlines():
             try:
                 (devspec, mountpoint, fstype, _options, _rest) = line.split(None, 4)
```
mulkieran: If you could piggy-back another patch changing above line to:

```(devspec, mountpoint, fstype, _rest) = line.split(None, 3)```

that would be great.

mulkieran:
There are a couple of places in comments where you say "first mountpoint listed by the system" and I'm not sure what that means. I feel it's the last one listed in /proc/mounts, so I figure it's some other aspect of the system you're referring to. Other than that, and the inline comments above, it looks good to me.

That'll be the first use of defaultdict in blivet. It's a nice convenience.

bcl:
Thanks, I'll review those before pushing -- it is actually the most recent mount, so I'm not sure why I called it first, unless I just didn't change them.

bcl:
pushed. comments fixed, they were leftover from the 1st pass at this.

PR #71, opened 2015-04-07
Consolidate common code in DeviceFormat class methods.

dwlehman:
This is the same model used within the device classes, which has worked pretty well.

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -367,6 +367,12 @@ class DeviceFormat(ObjectID):
         """
         log_method_call(self, device=self.device,
                         type=self.type, status=self.status)
+        self._preCreate(**kwargs)
+        self._create(**kwargs)
+        self._postCreate(**kwargs)
+
+    def _preCreate(self, **kwargs):
+        """ Perform checks and setup prior to creating the format. """
         # allow late specification of device path
         device = kwargs.get("device")
         if device:
@@ -375,21 +381,53 @@ class DeviceFormat(ObjectID):
         if not os.path.exists(self.device):
             raise FormatCreateError("invalid device specification", self.device)
 
+        if self.exists:
+            raise DeviceFormatError("format already exists")
+
+        if self.status:
+            raise DeviceFormatError("device exists and is active")
+
+    # pylint: disable=unused-argument
+    def _create(self, **kwargs):
+        """ Type-specific create method. """
+        pass
+
+    # pylint: disable=unused-argument
+    def _postCreate(self, **kwargs):
+        self.exists = True
+        self.notifyKernel()
+
     def destroy(self, **kwargs):
         """ Remove the formatting from the associated block device.
 
             :raises: FormatDestroyError
             :returns: None.
         """
-        # pylint: disable=unused-argument
         log_method_call(self, device=self.device,
                         type=self.type, status=self.status)
+        self._preDestroy(**kwargs)
+        self._destroy(**kwargs)
+        self._postDestroy(**kwargs)
+
+    # pylint: disable=unused-argument
+    def _preDestroy(self, **kwargs):
+        if not self.exists:
+            raise DeviceFormatError("format has not been created")
+
+        if self.status:
+            raise DeviceFormatError("device is active")
+
+        if not os.access(self.device, os.W_OK):
+            raise DeviceFormatError("device path does not exist or is not writable")
+
+    def _destroy(self, **kwargs):
+        rc = 0
+        err = ""
         try:
             rc = run_program(["wipefs", "-f", "-a", self.device])
         except OSError as e:
             err = str(e)
         else:
-            err = ""
             if rc:
                 err = str(rc)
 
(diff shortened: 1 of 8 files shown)


blivet/formats/__init__.py
```diff
@@ -367,6 +367,15 @@ def create(self, **kwargs):
         """
         log_method_call(self, device=self.device,
                         type=self.type, status=self.status)
+        self._preCreate(**kwargs)
+        try:
+            self._create(**kwargs)
+        except Exception:
+            raise
```
vpodzime: I think the try-except doesn't make sense here. Maybe it had ``finally:`` or ``else:`` where it was taken from (?), but in the form it has here it doesn't make a difference to a plain ``self._create(**kwargs)`` call.

blivet/formats/__init__.py
```diff
@@ -375,21 +384,56 @@ def create(self, **kwargs):
...
             :returns: None.
         """
-        # pylint: disable=unused-argument
         log_method_call(self, device=self.device,
                         type=self.type, status=self.status)
+        self._preDestroy(**kwargs)
+        try:
+            self._destroy(**kwargs)
+        except Exception:
+            raise
```
vpodzime: Same here.

blivet/formats/__init__.py
```diff
@@ -375,21 +384,56 @@ def create(self, **kwargs):
...
+    # pylint: disable=unused-argument
+    def _preDestroy(self, **kwargs):
+        if not self.exists:
+            raise DeviceFormatError("format has not been created")
+
+        if self.status:
+            raise DeviceFormatError("device is active")
+
+        if not os.access(self.device, os.W_OK):
+            raise DeviceFormatError("device path does not exist")
```
vpodzime: The error message could be more precise -- *"...not exist or is not writable"*

blivet/formats/fs.py
```diff
@@ -386,7 +386,12 @@ def _getFormatOptions(self, options=None, do_labeling=False):
         argv.append(self.device)
         return argv
 
-    def doFormat(self, options=None):
+    def _preCreate(self, **kwargs):
+        super(FS, self)._preCreate(**kwargs)
+        if not self.mkfsProg:
+            return
```
vpodzime: It would be nice if all ``_pre*`` functions returned bool values. If that's already the intention, the above should return False.

blivet/formats/lvmpv.py
```diff
@@ -96,55 +95,29 @@ def dict(self):
...
-            # Consider use of -Z|--zero
-            # -f|--force or -y|--yes may be required
+        # Consider use of -Z|--zero
+        # -f|--force or -y|--yes may be required
 
-            # lvm has issues with persistence of metadata, so here comes the
-            # hammer...
+        # lvm has issues with persistence of metadata, so here comes the
+        # hammer...
+        self.exists = True
```
vpodzime: The logic of setting ``self.exists`` to True here and changing it to False if something fails is quite weird. Could we please set it to False (or leave it as it is) here and set it to True if we successfully get to the end?

vpodzime:
Other than the minor modification requests above this looks good to me.

blivet/formats/fs.py
dwlehman: I agree. The reason I didn't do that is the Device classes don't and I wanted to keep them the same. The main difference is that it's okay to call setup on an already-active device or format, but it's not okay to call create on an already-existing device or format. That's why setup/teardown have a way to exit early while create/destroy do not.

blivet/formats/lvmpv.py
dwlehman: It is done to satisfy the precondition checking in DeviceFormat.destroy. I think I can just call DeviceFormat._destroy instead and avoid manipulating self.exists there.

blivet/formats/__init__.py
dwlehman: That is there because a later commit will be adding code to the except clause, but it's a good idea to add the try/except later since the additional complexity achieves nothing at this point.

blivet/formats/fs.py
vpodzime: Okay, thanks for the explanation.

blivet/formats/lvmpv.py
vpodzime: I think that would be better especially with a comment explaining why we need to call ``DeviceFormat._destroy()`` here (both why we need a destroy and why ``_destroy()``)

dwlehman: It's really hard to imagine that destroy call being necessary at this point. The original commit (from 2009) did not reference any bug, so it's impossible to verify for certain.

vpodzime: Isn't it there to prevent ``pvcreate`` from having issues with existing metadata/format on the device? libblockdev's ``pvcreate`` doesn't use the ``-f`` option to force PV creation and so didn't blivet's ``pvcreate`` IIRC.

dwlehman: Yes, that's why it was originally put there. I just suspect that the other measures put in place since then obviate this one.

- we dd zeros into the first MiB of new partitions
- we run wipefs -a on all new partitions (soon to be all new devices with a patch I have waiting)
- lvm zeros the first 2KiB by default in pvcreate and lvcreate
- lvm wipes all signatures using libblkid by default in lvcreate

vpodzime: Well, we can always try removing it and see what it breaks :)

PR #72, opened 2015-04-08
Pickling support for Size

vojtechtrefny:
I need to pickle size instances in blivet-gui for multiprocess communication and \__reduce__ function inherited from decimal causes precision loss and other problems.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -269,6 +269,11 @@ class Size(Decimal):
     def __deepcopy__(self, memo):
         return Size(self.convertTo())
 
+    # pickling support for Size
+    # see https://docs.python.org/3/library/pickle.html#object.__reduce__
+    def __reduce__(self):
+        return (self.__class__, (self.convertTo(),))
+
     def __add__(self, other, context=None):
         return Size(Decimal.__add__(self, other, context=context))
 
--- a/tests/size_test.py
+++ b/tests/size_test.py
@@ -25,6 +25,8 @@ import locale
 import os
 import unittest
 
+from six.moves import cPickle
+
 from decimal import Decimal
 
 from blivet.i18n import _
@@ -193,6 +195,10 @@ class SizeTestCase(unittest.TestCase):
         self.assertEqual(size.parseSpec("1e-1 KB"), Decimal(100))
         self.assertEqual(size.parseSpec("1E-4KB"), Decimal("0.1"))
         self.assertEqual(Size("1E-10KB"), Size(0))
+
+    def testPickling(self):
+        s = Size("10 MiB")
+        self.assertEqual(s, cPickle.loads(cPickle.dumps(s)))
 
 class TranslationTestCase(unittest.TestCase):
 


blivet/size.py
```diff
@@ -269,6 +269,9 @@ def __repr__(self):
     def __deepcopy__(self, memo):
         return Size(self.convertTo())
 
+    def __reduce__(self):
+        return (self.__class__, (str(self.convertTo()),))
```
vpodzime: I think this is worth a comment about the meaning  of the ``__reduce__`` method and it's return type/format. Without it I think many people would have to study the docs like I did which is quite cumbersome. Also the returned value looks really weird if the return type is not mentioned/described.

vpodzime:
Other than the missing comment this looks good to me.

vpodzime:
One more thing I forgot -- please make sure this works with Python 3.

mulkieran:
I do not think you need to convert result of convertTo() to a str. Size.__new__ will handle Decimal value correctly.

Please add a unit test for the new method.

vojtechtrefny:
New version -- comment and test added. Tested with python3 (it works!).

vpodzime:
Nice! Looks even better to me now.

tests/size_test.py
```diff
@@ -194,6 +196,10 @@ def testScientificNotation(self):
         self.assertEqual(size.parseSpec("1E-4KB"), Decimal("0.1"))
         self.assertEqual(Size("1E-10KB"), Size(0))
 
+    def testPickling(self):
+        s = Size("10 MiB")
+        self.assertEqual(s, cPickle.loads(cPickle.dumps(s)))
```
mulkieran: It's reassuring that the pickling does work, but...

```
c, a = s.__reduce__()
self.assertEqual(s, c(*a))
```

is more properly unit-testy, and you don't have to actually import the pickler that way.

vpodzime: But we care about pickling working not our implementation of some method working as it is written. Maybe not so unit-testy, but what @vojtechtrefny has in the test is what we actually want to test.

mulkieran: When we're unit-testing size...we are testing that size's methods work correctly.
```__reduce__``` is used for other things than pickling, like implementing copying.

I agree that it is important that the pickling works...but there is some better place to test that.
blivet-gui is the app that needs the ```__reduce__``` method for pickling...it should have tests to make sure that the pickling is successful for its purposes. At some point, blivet-gui should have _some_ tests, perhaps this is a good place to begin.

vpodzime:
This pull request is titled **Pickling support for Size** and as such adds code that makes pickling work and test that tests it. I have no idea about copying or whatever else using ``__reduce__``, but we don't care about it (here). Nobody should use the ``__reduce__`` method from the outside, however after this patch we will expect pickling to work on Size instances. If pickle suddenly started using some other method than ``__reduce__`` the test of the ``__reduce__`` method would work, but pickling wouldn't. And our tests wouldn't find out. Under the same logic we have tests for adding, subtracting, division, multiplication, exponentiation,... not for ``__add__`` method, ``__sub__`` method etc.

Why should *blivet-gui* tests if Size supports pickling? With the same logic blivet should tests if e.g. *Parted*'s objects work as they should. That doesn't make sense to me.

This PR adds support for pickling in Size and a test for such functionality. That's absolutely correct for me. If anybody wants to add some more tests, that could be a subject for a separate PR/patch, if we really want to test the ``__reduce__`` method.

dwlehman:
Do all other classes in blivet support pickling? Is it a requirement that they do? If not all, which ones?

My knee-jerk reaction to the cPickle/reduce argument is "test them both and move along", but I'm curious about people's answers to the above questions.

vojtechtrefny:
> Do all other classes in blivet support pickling? Is it a requirement that they do? If not all, which ones?

No and no. I actually don't need pickling support for size pickling neither. I have proxy objects for "objects too complex to be pickled" and I just can use those for size too.
But picklablable would make my life easier: 
a) it is already picklable because decimal i picklable but \__reduce__ inherited from decimal doesn't work with size; 
b) size is the only object I actually create on client side of blivet-gui and send back to "server" (my "object proxy" now works only for objects sent from server to client);

mulkieran:
I think that pickling and deepcopy are closely related and probably roughly as tricky. So, I'ld hesitate to say that blivet supports pickling all types of objects.

I am all in favor of having Size provide a ```__reduce__``` method, and it looks like this one is a good one.

But it also seems that Vojtech ran into a complicated problem with pickling Size objects as Decimals. It seems that blivet-gui is the client for that, and the one that needs it to work according to some specification in some specific situation that it is in that blivet knows nothing about. So, I believe that blivet-gui should be testing that part of the behaviour.

vpodzime:
> But it also seems that Vojtech ran into a complicated problem with pickling Size objects as Decimals. 

It's not a complicated problem, it's just a missing override for a method of the class (``Decimal``) we inherit the ``Size`` class from. It's the same type of issue as when ``rmul`` (or whatever we had to fix recently) returned ``Decimal`` instances instead of ``Size`` instances. The result here is that pickling doesn't work on ``Size`` objects without the change that @vojtechtrefny proposed in this PR. Just like anaconda discovered many issues in blivet and the tests didn't go to anaconda, tests for this issue discovered in blivet-gui shouldn't go to blivet-gui.

dwlehman:
If I am correct that by testing pickling we also test ```__reduce__```, we can just take this patch as it is.

PR #73, opened 2015-04-10
F22 unicode

dashea:
The second commit might need to be ported to master in some form


vpodzime:
This looks good to me for the *f22-branch*. I think for master we should do what I mentioned in the previous commit.

dashea:
Pushed to f22-branch.

PR #76, opened 2015-04-10
Do not continue when presented with unusable storage configurations

dwlehman:
There are some issues that must be resolved before blivet can do anything useful. The two that this PR addresses are multiple LVM VGs with the same name and disklabel corruption that prevents parted from acknowledging a disklabel when partitions are present on the disk.

Until the lvm tools accept UUID in place of VG name for at least lvremove, vgchange, and vgremove we have no way to work with multiple VGs with the same name. This related bug has been sitting since 2008: https://bugzilla.redhat.com/show_bug.cgi?id=449832

The first patch enables parted to proceed with a partially-corrupt GPT disklabel, but I suspect it will not cover all such cases, hence the exception handling.

There will soon be an accompanying anaconda PR that handles these exceptions by showing the error and the suggested course of action to the user.

### DIFF
--- a/blivet/errors.py
+++ b/blivet/errors.py
@@ -20,6 +20,8 @@
 # Red Hat Author(s): Dave Lehman <dlehman@redhat.com>
 #
 
+from .i18n import N_
+
 class StorageError(Exception):
     def __init__(self, *args, **kwargs):
         self.hardware_fault = kwargs.pop("hardware_fault", False)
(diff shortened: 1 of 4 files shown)


dwlehman:
Related: rhinstaller/anaconda#66

clumens:
You should bump the pyparted requirement in the spec file up to something more modern that has the exn handling support.  3.10.3 was the first version after I committed those patches.

vpodzime:
Looks good to me (with the version bump as @clumens suggested).

PR #77, opened 2015-04-14
Master fixups

bcl:


vpodzime:
My version of the "Make sure the file exists..." patch is sitting on the anaconda-patches list since Friday. :) The "Drop self.status from..." looks good to me.

bcl:
> My version of the "Make sure the file exists..." patch is sitting on the anaconda-patches list since Friday.

Could have sworn I ack'd that or someone else did. I do remember reading it. Consider this an ACK since I don't have that email anymore.

> This should probably return mountpoint != self.systemMountpoint or whatever that would look like with all the recent changes. The idea is to return False if you should not proceed with the actual mount command.

Ok.

dwlehman:
Setting ack under the assumption that the file size patch is dropped.

bcl:
Pushed just that 1 patch.

PR #78, opened 2015-04-14
Fix "anaconda hangs while trying to discover iscsi..." (#1166652)

jkonecny12:
This was 2 errors in one.

First problem: anaconda freeze for some time.
Second problem: libiscsi was run on thread and libiscsi the library
is not thread safe.

Anaconda freeze is because of GIL is taken on library thread which is
trying to connect to server.
The thread safe problem is because libiscsi is using signals internally
which were delivered to another thread (this is hard to solve in python
threads...). Solution is to start discover method on new process.

So new process is now created to call libiscsi discover_sendtargets
method. This should fix both problems.

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -23,6 +23,7 @@ from . import util
 from .flags import flags
 from .i18n import _
 from .storage_log import log_exception_info
+from multiprocessing import Process, Pipe
 import os
 import logging
 import shutil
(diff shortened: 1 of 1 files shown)


blivet/iscsi.py
```diff
@@ -277,10 +319,41 @@ def discover(self, ipaddr, port="3260", username=None, password=None,
...
+                          "_call_discover_targets")
+
+            p.join()
+
+            # convert dictionary back to iscsi nodes object
+            if found_nodes is not None:
+                for node in data:
+                    found_nodes.append(libiscsi.node(**node))
+
+
```
vpodzime: Looks good to me. The only minor concern for me is the amount of empty space in the new code. I think you can get rid of some blank lines -- before ``p.start()``, before ``if not ok``, before ``except EOFError`` and at the very end.

blivet/iscsi.py
```diff
@@ -277,10 +319,41 @@ def discover(self, ipaddr, port="3260", username=None, password=None,
...
+                if not ok:
+                    found_nodes =  None
+                    log.debug("iSCSI: error raised when "
+                              "discover_sendtargets process called: %s",
+                              str(data))
+
+            except EOFError:
+                found_nodes = None
+                log.debug("iSCSI: can't receive response from "
+                          "_call_discover_targets")
```
vpodzime: This should be ``log.error()`` since that's quite an issue, I think.

jkonecny12: I see, my bad.
I'll fix this locally.

blivet/iscsi.py
jkonecny12: Ok. Thank you for response.
Will be changed locally.

jkonecny12:
I found an error with returned None object from libiscsi.

Closing this pull request -- new will be created.

PR #80, opened 2015-04-15
Add libselinux-python to package dependencies (#1211834)

vojtechtrefny:
Blivet imports selinux module but doen't depend on libselinux-python (only anaconda-core depends on that). Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1211834

### DIFF
--- a/python-blivet.spec
+++ b/python-blivet.spec
@@ -34,6 +34,7 @@ Requires: e2fsprogs >= %{e2fsver}
 Requires: lsof
 Requires: libblockdev >= %{libblockdevver}
 Requires: libblockdev-plugins-all >= %{libblockdevver}
+Requires: libselinux-python
 
 %description
 The python-blivet package is a python module for examining and modifying


mulkieran:
This seems like the right decision to me, but we should at least consider whether there wouldn't be a reason to prefer isolating the very minor dependencies on the package that occur in our code.

It's funny this hasn't come up until now, since it is an import in util.py that has been there since the split, AFAICT.

dwlehman:
We don't use selinux except when installer_mode is enabled. It makes more sense to conditionalize the import of selinux so that a missing libselinux-python doesn't cause trouble unless/until we try to use it. We could log the fact that it's missing.

mulkieran:
There is also an selinux_test.py file in tests directory that would require the same treatment.

vpodzime:
I'm not quite sure it makes much sense to complicate the code and everything for **717 KB** of dependencies and all that only in the case where firewalld, system-config-users, system-config-date, setroubleshoot and of course anaconda are not installed.

bcl:
I mostly agree with @vpodzime, but I also think it's overkill to drag in the whole package for a simple check to see if selinux is enabled. May as well check for /sys/fs/selinux/enforcing instead.

vpodzime:
> I mostly agree with @vpodzime, but I also think it's overkill to drag in the whole package for a simple check to see if selinux is enabled. May as well check for /sys/fs/selinux/enforcing instead.

I had no idea there's such nice way to find it out. Thanks for the suggestion, @bcl! However, blivet also calls ``selinux.matchpathcon()`` and ``selinux.lsetfilecon()`` and those are a bit more tricky, I think.

vojtechtrefny:
Ok, we probably should find a better solution for master. But I'd like to push this simple fix at least to f21 and f22 branches to fix this for blivet-gui -- because if you don't have libpython-selinux it will just crash with an Import Error.

vpodzime:
Giving an ACK for at least f21 and f22 branches.

bcl:
> I had no idea there's such nice way to find it out. Thanks for the suggestion, @bcl! However, blivet also calls selinux.matchpathcon() and selinux.lsetfilecon() and those are a bit more tricky, I think.

somehow I missed those calls in my grep attempts. In that case just put the Requires in there.

dwlehman:
Go ahead and push the ```Requires``` addition to master in addition to whichever fedora branches are appropriate. libselinux-python only requires python and libselinux, neither of which should be hard to come by. We don't need to spend any more time on this one.

vojtechtrefny:
Pushed.

PR #82, opened 2015-04-16
Include LUKSDevice information in kickstart data (#1139222)

mulkieran:
Resolves: rhbz#1139222

Previously, if a device had a child that was a LUKS device that child was
ignored when generating ksdata. What this meant was that that device's
format, which was luks, was used in generating the kickstart data.

Now, the LUKSDevice, which contains the format that should appear in the ks
output is used. This means that the LUKS format is used to set some ks data,
but the mountpoint and so forth are overwritten with the correct values
when the format on the LUKSDevice is used to populate those values.

The --passphrase option with passphrase continues to not appear in the output
anaconda-ks.cfg file.


mulkieran:
also for master.

vpodzime:
Same suggestion as @bcl. Otherwise this looks good to me.

mulkieran:
Superseded by PR#98.

Issue #65, opened 2015-04-02
Make use of new libblockdev API with better exception reporting


vpodzime:
*libblockdev* is soon going to provide a new API (as described in [rhinstaller/libblockdev/pull/#4](https://github.com/rhinstaller/libblockdev/pull/4)) which provides better exception reporting and thus allows *blivet* to work better with exceptions coming from libblockdev.

vpodzime:
Should be resolved now.

PR #85, opened 2015-04-17
Make use of the new libblockdev error reporting

vpodzime:
As of version 0.10 libblockdev provides better error reporting if its ErrorProxy
instances are used (e.g. blockdev.lvm.lvcreate instead of lvm.lvm_lvcreate where
blockdev.lvm is an ErrorProxy instance). Let's make use of this feature and
catch native python exceptions defined in libblockdev's python GI overrides
instead of GLib.GError.


vpodzime:
this should also go to master

vpodzime:
Forgot to mention that this resolves the issue #65

PR #87, opened 2015-04-21
Fix "anaconda hangs while trying to discover iscsi..." (#1166652)

jkonecny12:
This was 2 errors in one.

First problem: anaconda freeze for some time.
Second problem: libiscsi was run on thread and libiscsi the library
is not thread safe.

Anaconda freeze is because of GIL is taken on library thread which is
trying to connect to server.
The thread safe problem is because libiscsi is using signals internally
which were delivered to another thread (this is hard to solve in python
threads...). Solution is to start discover method on new process.

So new process is now created to call libiscsi discover_sendtargets
method. This should fix both problems.

Note: Will be applied to master too.

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -23,6 +23,7 @@ from . import util
 from .flags import flags
 from .i18n import _
 from .storage_log import log_exception_info
+from multiprocessing import Process, Pipe
 import os
 import logging
 import shutil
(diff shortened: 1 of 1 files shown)


jkonecny12:
I did some changes like you said bcl. Please look on it now.

Thank you.

blivet/iscsi.py
```diff
@@ -277,12 +332,37 @@ def discover(self, ipaddr, port="3260", username=None, password=None,
...
+                              str(data))
+            except EOFError:
+                ok = False
+                log.error("iSCSI: can't receive response from "
+                          "_call_discover_targets")
+
+            p.join()
+
+            # convert dictionary back to iscsi nodes object
+            if ok is True:
```
vpodzime: ``if ok`` would be sufficient here

vpodzime:
Other than the neatpick above this looks good to me.

blivet/iscsi.py
jkonecny12: I will change it before push

PR #88, opened 2015-04-21
If the parent volume has a label, use it in subvol's kickstart (#1072060...

mulkieran:
...)

Resolves: rhbz#1072060

The LABEL= entry identifies the parent of the specified subvolume by its
label. If the parent volume does not have a label, then no entry is made
that identifies the parent volume and the generated kickstart
file will be rejected by ksvalidator.


vpodzime:
Looks good to me.

PR #90, opened 2015-04-22
Master tasks 20150422

mulkieran:
Now with fewer commits.


vpodzime:
Other that the (rather minor) comments above this looks good to me. With the exception of the dependecies being removed out out python-blivet.spec's scope where they belong (as I tried to explain in [the related Anaconda's PR] (https://github.com/rhinstaller/anaconda/pull/83))

PR #93, opened 2015-04-23
Fix adding new partitions on disks with active devices

vojtechtrefny:
See https://bugzilla.redhat.com/show_bug.cgi?id=1212841

Currently is not possible to add new partitions to disk with an existing LVM VG. 'findActiveDevicesOnActionDisks' should disallow only actions involving changes on the disklabel not adding new partitions.

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -247,18 +247,21 @@ class ActionList(object):
         disks = []
         for action in self._actions:
             disk = None
-            if action.isDevice and isinstance(action.device, PartitionDevice):
-                disk = action.device.disk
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
New version. I've changed the check to include both active formats (except disks) and active devices (except partitions and disks). Unfortunately dwlehman's comment on this seems to be lost after force pushing the change.

vpodzime:
Looks good to me too with the same comment about the ``if`` as @dwlehman has above.

vojtechtrefny:
I actually had it this way but it looked too complicated and hard to understand so I changed it.
It's merged now so if you think it should be changed, you'll have to open new pull request...

PR #94, opened 2015-04-23
Move definition of _level attribute above super-constructor call.

mulkieran:
The basic problem is that level is a very important concept for md
so a lot of log messages log this information. Some of these
log messages are executed as a result of running the parent constructor.
The particular problem I saw is that self.updateSize() is called by
StorageDevice.__init__, which understandably logs the size,
and the size() methods logs its result, along with the level, leading to
an AttributeError. Now there will be no AttributeError, but the RAID
level will be displayed as None when this happens.


vpodzime:
Sure thing!

PR #98, opened 2015-04-28
Include LUKSDevice information in kickstart data (#1139222)

mulkieran:
Resolves: rhbz#1139222

Previously, if a device had a child that was a LUKS device that child was
ignored when generating ksdata. What this meant was that that device's
format, which was luks, was used in generating the kickstart data.

Now, the LUKSDevice, which contains the format that should appear in the ks
output is used. This means that the LUKS format is used to set some ks data,
but the mountpoint and so forth are overwritten with the correct values
when the format on the LUKSDevice is used to populate those values.

The --passphrase option with passphrase continues to not appear in the
output anaconda-ks.cfg file.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -72,7 +72,7 @@ from pykickstart.constants import AUTOPART_TYPE_LVM, CLEARPART_TYPE_ALL, CLEARPA
 
 from .storage_log import log_exception_info, log_method_call
 from .errors import DeviceError, DirtyFSError, FSResizeError, FSTabTypeMismatchError, LUKSDeviceWithoutKeyError, UnknownSourceDeviceError, SanityError, SanityWarning, StorageError, UnrecognizedFSTabEntryError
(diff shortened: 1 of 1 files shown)


mulkieran:
This is the second try. The only difference from the previous is that I use the ```raw_device``` property instead of using ```instanceof(d, LUKSDevice)``` and the LUKSDevice's ```slave``` property since this makes the change more internally consistant.

blivet/__init__.py
```diff
@@ -2170,6 +2179,10 @@ def updateKSData(self):
             cls = getattr(self.ksdata, class_attr)
             data = cls()    # all defaults
 
+            # if this is the raw device of a complementary device use the
+            # complementary device instead.
+            device = next((d for d in complementary_devices if d.raw_device is device), device)
```
vpodzime: I'd prefer here a list comprehension assigned to a properly named variable with an ``if`` instead of the combinatio of ``next()`` and a generator. If would be much better readable to me and better matching the comment above it.

vpodzime:
Except of the comment above this looks good to me.

PR #99, opened 2015-04-28
Do not update size information on LVM snapshots

vojtechtrefny:
Current updateSize method implementation checks the size of devices in /sys where the size of LVM snapshots is incorrect (it reports the size of the LV instead of the actual size of the snapshot).

My fix actually prevents blivet from updating size of the snapshot. The question is if we want to update the size because it would require to run the devicetree.lvInfo() function again (because only "lvs" reports the correct size for snapshots).

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -966,6 +966,11 @@ class LVMSnapShotDevice(LVMSnapShotBase, LVMLogicalVolumeDevice):
         return (self.origin == dep or
                 super(LVMSnapShotBase, self).dependsOn(dep))
 
+    def readCurrentSize(self):
+        # override StorageDevice.readCurrentSize because /sys reports wrong size
+        # for snapshots
+        return self._size
+
 class LVMThinPoolDevice(LVMLogicalVolumeDevice):
     """ An LVM Thin Pool """
     _type = "lvmthinpool"


vpodzime:
I think this should be safe because AFAICT the size of the snapshot cannot change without the update of its ``self._size`` from blivet. And if it changes externally, we are doomed anyway (until the uevent support gets merged).

vpodzime:
Not giving ACK yet though as more eyes could reveal some "gotcha" I'm missing here.

dwlehman:
It would be better to override ```StorageDevice.readCurrentSize``` in ```LVMSnapShotBase``` if the problem is reading of size from /sys.

vojtechtrefny:
New version with readCurrentSize.

dwlehman:
Is this for old-style snapshots or thin snapshots or both?

For the old-style snapshots there is always a <snapshot_name>-cow lv that will give the correct size, eg:

```
[root@localhost blivet]# lvs blivet
  LV            VG     Attr       LSize   Pool Origin  Data%  Meta%  Move Log Cpy%Sync Convert
  private       blivet owi-a-s--- 256.00m                                                     
  private-snap1 blivet swi-a-s--- 200.00m      private 0.00                                   
[root@localhost blivet]# ls -l /dev/mapper/blivet-*
lrwxrwxrwx. 1 root root 7 May  7 11:39 /dev/mapper/blivet-private -> ../dm-3
lrwxrwxrwx. 1 root root 7 May  7 11:39 /dev/mapper/blivet-private-real -> ../dm-2
lrwxrwxrwx. 1 root root 7 May  7 11:39 /dev/mapper/blivet-private--snap1 -> ../dm-5
lrwxrwxrwx. 1 root root 7 May  7 11:39 /dev/mapper/blivet-private--snap1-cow -> ../dm-4
[root@localhost blivet]# cat /sys/block/dm-3/size
524288
[root@localhost blivet]# cat /sys/block/dm-4/size
409600
[root@localhost blivet]# cat /sys/block/dm-5/size
524288
```

I think the size is reported correctly for thin snapshots using the normal methods (not the trick above) .

vojtechtrefny:
> Is this for old-style snapshots or thin snapshots or both?

Only for old-style snapshots. There is no problem with thin snapshots.

> For the old-style snapshots there is always a -cow lv that will give the correct size, eg:

Oh, I didn't know about this. I'll prepare new version next week (with readCurrentSize using the cow lv to read the correct sys file).

PR #100, opened 2015-04-28
Un-escape '-'s in names or paths for _all_ lvm lv or vgs.

mulkieran:
Assume that whenever lvm presents blivet w/ a name with two consecutive
dashes this represents an escaped single dash.

Previously only doing this for lvmvg's and lvmlv's, which turned out
to be a problem when looking up subclass instances, like lvm thin lvs.


mulkieran:
New version uses instanceof and has a correct commit message.

vpodzime:
Looks good to me.

PR #103, opened 2015-05-01
Allow passing KiB values to vgcreate -s option

mulkieran:
Resolves: rhbz#1206526

Currently blivet exepects pesize parm of the vgcreate function
should be in MiB. Any given value will be converted to MiB format
and omits any fractional part. That means, if any KiB values are
passed to pesize parm, it becomes 0 and the actual lvm lvcreate
faild to create the vg and returns an error.
This limits passing any KiB values for pesize to create a vg.
This patch fixes the issue by accepting the given value in KiB.

Submitted-by: Timothy Asir Jeyasingh <tjeyasin@redhat.com>


mulkieran:
Acked, see #33.

PR #105, opened 2015-05-05
Master btrfs

mulkieran:
Fixes a bug where mountpoint of btrfs top-level volume was not found in mountsCache.

Two parts of fix are:
* Make MAIN_VOLUME_ID default subvolspec for btrfs format. If making a subvol it should always be
overridden with subvol's subvolspec.
* Make mountsCache use consistent types for subvolspec in keys.

Symptom of bug was uncaught RuntimeError "btrfs subvol create requires mounted volume" while executing storage actions.


vpodzime:
Other than the minor comments above this looks good to me.

PR #106, opened 2015-05-05
Master tasks 20150503

mulkieran:
Changes are:
* quick way to make (un)available resources in availability.py
* all external resources are now constants (more uniform, makes testing easier)
* have available and availabilityErrors in ExternalResource and Task, define available in terms of availabilityErrors, makes use in fs a bit less wordy
* caching is done at the external resource level and at the PackageMethod level
* make whether or not lvm actions in devicelibs are actually done dependent on whether lvm is available
* track availability information in remaining bits of devicelibs files
* add check to get_supported_raid_levels to take into account availability information
* add a device type supported check in devicefactory
* guard some test against device support being missing


vpodzime:
Looks good to me other than the minor comments above.

PR #108, opened 2015-05-07
Add progress report callback for action processing

vojtechtrefny:
Add a simple callback for all actions to report progress when running blivet.doIt() in blivet-gui


vpodzime:
Looks good to me. I think this provides nice and generic interface for reporting and/or logging progress.

PR #109, opened 2015-05-07
Master btrfs c

mulkieran:
Supersedes PR#107.

The final patch, "Cache data..." is the only interesting one.

The main change from previous PR is that since mountinfo cache is only needed in _getActiveMounts(), it is only instantiated there. In #107 is was an attribute of the mounts cache, which was excessive. This way, the cache needs to be managed less, which is good.


mulkieran:
The change here is that instead of setting the subvolspec in the BTRFS format constructor I set the subvolspec for a volume in the newBTRFS method. I believe that subvolume devices created by newBTRFS will not have their subvolspecs set properly, but I believe that that will prove significantly less serious than not having the parent subvolspec set properly.

It is not absolutely clear to me how to set the subvolspec for newly created subvolumes properly.

vpodzime:
Looks good to me too.

PR #110, opened 2015-05-08
Drop use of md metadata version 1.0

dwlehman:
We began using this metadata version because grub1 could not read arrays with v1.1 or newer metadata. Now that we're using grub2 we can stop using v1.0.

See https://bugzilla.redhat.com/show_bug.cgi?id=1061711#c6

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -22,7 +22,7 @@
 
 import copy
 
-from .deviceaction import ActionCreateDevice, ActionDestroyDevice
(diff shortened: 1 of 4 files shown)


mulkieran:
This looks good to me.

vpodzime:
Looks good to me too.

dwlehman:
I overlooked the fact that this will break mirrored /boot/efi. I have to rework this to stop using v1.0 for /boot but keep it for /boot/efi. The second commit will have to get dropped.

PR #111, opened 2015-05-08
Store vendor/model information for DiskDevice instances.

### LINKED ISSUE #84: DiskDevice instances report no model
Probably since the commit 73d9c995c8d46f1aa248d1ade3bbb1838653b6b0 all *DiskDevice* instances report no model since they have ``self._model`` unset and it is no longer updated from their ``self.partedDevice`` attribute. I tried looking at the udev info we get at the point where we create *DiskDevice* instances, but the model is not there and ``udev.device_get_model(info)`` returns ``None``. So I don't really have any idea how to get the model back.

dwlehman:
Also make sure _model and _vendor attrs are never set to None in the
StorageDevice constructor.

Fixes #84

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -105,7 +105,7 @@ class DiskDevice(StorageDevice):
 
     @property
     def description(self):
-        return self.model
+        return " ".join(s for s in (self.vendor, self.model) if s)
 
     def _preDestroy(self):
         """ Destroy the device. """
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -476,12 +476,10 @@ class Populator(object):
         serial = udev.device_get_serial(info)
         bus = udev.device_get_bus(info)
 
-        # udev doesn't always provide a vendor.
-        vendor = udev.device_get_vendor(info)
-        if not vendor:
-            vendor = ""
+        vendor = util.get_sysfs_attr(sysfs_path, "device/vendor")
+        model = util.get_sysfs_attr(sysfs_path, "device/model")
 
-        kwargs = { "serial": serial, "vendor": vendor, "bus": bus }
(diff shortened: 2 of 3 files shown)


blivet/devices/disk.py
```diff
@@ -105,7 +105,11 @@ def mediaPresent(self):
 
     @property
     def description(self):
-        return self.model
+        desc = self.vendor
+        if desc:
+            desc += " "
+        desc += self.model
+        return desc
```
mulkieran:
```
" ".join(s for s in (self.vendor, self.model) if s)
```
seems clearer to me.

blivet/populator.py
```diff
@@ -476,12 +476,10 @@ def addUdevDiskDevice(self, info):
         serial = udev.device_get_serial(info)
         bus = udev.device_get_bus(info)
 
-        # udev doesn't always provide a vendor.
-        vendor = udev.device_get_vendor(info)
-        if not vendor:
-            vendor = ""
+        vendor = util.get_sysfs_attr(sysfs_path, "device/vendor")
+        model = util.get_sysfs_attr(sysfs_path, "device/model")
```
mulkieran: Given that ```get_sysfs_attr()``` can, in theory, return None, appending ```or ""``` to the above two lines is necessary to be consistent with goal of always using empty str to represent no value for vendor and model.

dwlehman: I did so in StorageDevice.__init__ instead so it would be covered for other callers as well.

blivet/devices/disk.py
dwlehman: Agreed.

mulkieran:
Looks good, then.

vpodzime:
Looks good to me too.

PR #113, opened 2015-05-12
Master tasks 20150513

mulkieran:
Acked with changes in #106. Leaves out spec file dependency removals which will be handled in a separate commit.


vpodzime:
NICE! We should open a bottle of champagne! :)

mulkieran:
Farewell to massive rebases!

- mulhern

----- Original Message -----

PR #114, opened 2015-05-13
Use COW device to get actual size of LVM snapshots

vojtechtrefny:
New version of #99 

-----
Override readCurrentSize of LVM old-style snapshots to check size
on COW device instead of snapshot device. (The latter one reports
snapshot to be same size as the LV.)

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -24,6 +24,7 @@ from six import add_metaclass
 import abc
 import pprint
 import re
+import os
 from gi.repository import BlockDev as blockdev
 
 # device backend modules
@@ -39,6 +40,7 @@ from ..size import Size, KiB, MiB, ROUND_UP, ROUND_DOWN
 import logging
 log = logging.getLogger("blivet")
 
+from .lib import LINUX_SECTOR_SIZE
 from .device import Device
 from .storage import StorageDevice
 from .container import ContainerDevice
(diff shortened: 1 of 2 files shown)


blivet/util.py
```diff
@@ -226,6 +226,18 @@ def get_sysfs_path_by_name(dev_node, class_name="block"):
...
+    """ Return sysfs path of cow device for a given lvm snapshot.
+    """
+
+    cow_path = dev_path + "-cow"
+    if not os.path.islink(cow_path):
+        raise RuntimeError("get_cow_sysfs_path: Could not find cow device for" %
+                            (dev_path))
+
+    dm_name = os.path.basename(os.path.realpath(cow_path))
+    return os.path.join(os.path.split(dev_sysfsPath)[0], dm_name)
```
vpodzime: I think these two lines are worth a comment about what's going on there as I find them quite cryptic.

blivet/util.py
```diff
@@ -226,6 +226,18 @@ def get_sysfs_path_by_name(dev_node, class_name="block"):
         raise RuntimeError("get_sysfs_path_by_name: Could not find sysfs path "
                            "for '%s' (it is not at '%s')" % (dev_node, dev_path))
 
+def get_cow_sysfs_path(dev_path, dev_sysfsPath):
+    """ Return sysfs path of cow device for a given lvm snapshot.
```
vpodzime: *"for a given device"* would be better as it is more generic (as this function potentially is) and doesn't create a confusion by *"lvm snapshot"* suddenly appearing here for no obvious reason.

vpodzime:
Other than the minor things above this looks good to me.

blivet/devices/lvm.py
```diff
@@ -966,6 +968,20 @@ def dependsOn(self, dep):
...
 
+    def readCurrentSize(self):
+        log_method_call(self, exists=self.exists, path=self.path,
+                        sysfsPath=self.sysfsPath)
+        size = Size(0)
+        if self.exists and os.path.exists(self.path) and \
+           os.path.isdir(self.sysfsPath):
+            cowSysfsPath = util.get_cow_sysfs_path(self.path, self.sysfsPath)
+
+            if os.path.exists(cowSysfsPath) and os.path.isdir(cowSysfsPath):
```
mulkieran: You should be able to omit ```os.path.exists(cowSysfsPath)``` since ```os.path.isdir(cowSysfsPath)``` will evaluate to False if ```cowSysfsPath``` does not exist.

vojtechtrefny: Makes sense, fixed locally.

PR #115, opened 2015-05-13
Workaround for chrooted mountpoints  (#1217578)

vojtechtrefny:
This solution is definitely not the ideal one but at least it fixes the problem with failing atomic builds. It is neccessary to come up with a better solution for all nodev devices and its mountpoints.

-----
These are currently failing because of mounts cache. The problem is mostly with chrooted mounts and bind mounts of 'nodev' devices. This solution is definitely not ideal one but at least it fixes the problem with failing atomic builds. It is neccessaty to come up with a better solution for all nodev devices and its mountpoints.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -138,6 +138,8 @@ class FS(DeviceFormat):
 
         self._targetSize = self._size
 
+        self._chrootedMountpoint = None
+
         if self.supported:
             self.loadModule()
 
@@ -496,6 +498,9 @@ class FS(DeviceFormat):
         if not self.exists:
             return None
 
+        if self._chrootedMountpoint:
+            return self._chrootedMountpoint
+
         # It is possible to have multiple mountpoints, return the last one
         try:
             return mountsCache.getMountpoints(self.device,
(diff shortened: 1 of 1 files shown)


mulkieran:
Is it really impossible to handle this problem in mountsCache itself?

vpodzime:
> Is it really impossible to handle this problem in mountsCache itself?

It's not entirely easy and no way we could come up with looked safe enough to be applied fast. As @vojtechtrefny mentioned above *"This solution is definitely not the ideal one but at least it fixes the problem with failing atomic builds."* which we should do ASAP.

vojtechtrefny:
Updated version based on @vpodzime comments.

@mulkieran as @vpodzime already mentioned, this is only temporary workaround. I tried to fix the mountsCache to be able to work with nodev filesystems and bind mounts but there is just too many problems and it will need some time and more testing.

mulkieran:
ok, so long as we all explicitly agree that it's a _temporary_ fix that will be reverted soon, and it does fix the atomic builds, I'm fine w/ it.

vojtechtrefny:
Pushed.

PR #116, opened 2015-05-13
Filter GLib.GError instead of GLib.Error in pylint false positives.

mulkieran:
GError and Error are just aliases of each other, and GError is more
distinctive, thus a better choice for filtering so long as it is
used always in place of Error.

Should have gone in with commit 3f0aa449e42bc64da78e2ff8c0553b4526deed00.

### DIFF
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -9,7 +9,7 @@ class BlivetLintConfig(PocketLintConfig):
         PocketLintConfig.__init__(self)
 
         self.falsePositives = [ FalsePositive(r"BTRFSVolumeDevice._create: Instance of 'DeviceFormat' has no 'label' member"),
-                                FalsePositive(r"Catching an exception which doesn't inherit from BaseException: (BlockDev|DM|Crypto|Swap|LVM|Btrfs|MDRaid|)Error$"),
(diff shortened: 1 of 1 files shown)


mulkieran:
There is a sense in which the code that initializes ```libblockdev``` in ```blivet/__init__.py``` that is causing the pylint error shouldn't be like it is and that the interface of the methods ```try_init()``` and ```reinit()``` should be changed. At first I thought that
the change should be that more work should be done on ErrorProxy so that these methods actually raised an exception in some sublcass of ```BlockDevError```. That would require a little work in the ```ErrorProxy``` definition, AFAICT, because these methods don't belong to a particular plugin, and so their prefix is really the empty string (and other problems related to that).

But now I think that a plugin not getting loaded should not be something that raises an exception and that the idea that each of these methods should return a list of the plugins loaded is actually still a very good one. I think that implementing this would require a bit of work in blockdev.c to enrich the information returned by the various methods, but that it would probably be worth the effort. I also think that the meaning of the static variable ```initialized``` might have to be modified...I think that now it means all plugins were succesfully loaded. What that means is that if not all plugins are loaded in ```blivet/__init__.py``` ```initialized``` will always be false. ```initialized``` seems more like ```complete``` to me.

mulkieran:
clumens acked this trivial change.

vpodzime:
I understand the point. The discussion and tracking should go on in some libblockdev-related place, so I created [a separate issue] (https://github.com/rhinstaller/libblockdev/issues/15).

mulkieran:
Great, thanks!

- mulhern

----- Original Message -----

PR #118, opened 2015-05-18
Master hawkey

mulkieran:


vpodzime:
Looks good to me.

mulkieran:
Had to update this slightly, catching an IOError when loading the system repo.

vpodzime:
> Had to update this slightly, catching an IOError when loading the system repo.

It seems almost like a bug in *hawkey* that they just pass the ``IOError`` further from their calls. In which cases is this exception raised? They should probably "wrap it" in their own exception class. And we should at least add a comment describing the weird case when it gets raised. With that added this still looks good to me.

mulkieran:
Yes, it is a bit weird. Comment added, bz posted.

vpodzime:
Thanks!

PR #120, opened 2015-05-21
Run the test suite with both Python 2 and Python 3

atodorov:
Since blivet is supposed to be compatible with both python 2 and 3 IMO it makes sense to run the test suite with both versions. (not to mention build the entire package with 2 or 3) At present only python2 is used to execute the test suite while python3 is used for pylint (due to pocketlint being python3 only). 

This PR introduces the following changes:

- adds Makefile targets to execute the test suite under python 2 and 3 separately
- make test now runs the test suite with *both* python 2 and 3
- adds 2 separate targets to execute test suite with coverage report for python 2 and 3
- adds new check-requires target which checks if Requires:/BuildRequires and a few manual added ones are present

- replaces `/usr/bin/python` with `/usr/bin/env python` inside test cases. NOTE: I don't see any executable tests/*.py files but maybe this is a good measure. Otherwise this line doesn't have any effect and can be removed. 

- updates the test cases to take care of several deprecation warnings coming from unittest. I've introduces a BlivetTestCase class which adds some aliases in case we're running with python2. 

TODO:

There are still a few places which use hard-coded python version:
- Makefile uses PYTHON=python2 which is used in a few places calling setup.py
- some scripts/ use /usr/bin/python explicitly
- runpylint.py uses /usr/bin/python3 explicitly

make check uses pocketlint which is Python3 only. I have not looked into how hard it will be to build it for Python2. Will do it tomorrow.

Test results:

Running the test suite with python2 gives me 4 errors:
```
ERROR: testAvailabilityMDRAIDplugin (devices_test.dependencies_test.MockingDeviceDependenciesTestCase)
  File "/root/blivet/tests/devices_test/dependencies_test.py", line 64, in testAvailabilityMDRAIDplugin
    self.assertIsNotNone(ActionCreateFormat(self.luks, fmt=getFormat("macefi")))
  File "blivet/deviceaction.py", line 530, in __init__
    raise ValueError("resource to create this format %s is unavailable" % fmt)
ValueError: resource to create this format non-existent macefi filesystem is unavailable

ERROR: testGetLabelArgs (formats_test.labeling_test.MethodsTestCase)
  File "/root/blivet/tests/formats_test/labeling_test.py", line 73, in testGetLabelArgs
    reiserfs = self.fs["reiserfs"]
KeyError: 'reiserfs'

ERROR: testActionDependencies (action_test.DeviceActionTestCase)
Verify correct functioning of action dependencies.
  File "/root/blivet/tests/action_test.py", line 944, in testActionDependencies
    format_lv = ActionCreateFormat(testlv, fmt)
  File "blivet/deviceaction.py", line 530, in __init__
    raise ValueError("resource to create this format %s is unavailable" % fmt)
ValueError: resource to create this format non-existent None is unavailable

ERROR: testContainerActions (action_test.DeviceActionTestCase)
  File "/root/blivet/tests/action_test.py", line 1122, in testContainerActions
    fmt=new_lv_format)
  File "tests/storagetestcase.py", line 245, in scheduleCreateFormat
    action = blivet.deviceaction.ActionCreateFormat(device, fmt)
  File "blivet/deviceaction.py", line 530, in __init__
    raise ValueError("resource to create this format %s is unavailable" % fmt)
ValueError: resource to create this format non-existent xfs filesystem is unavailable

Ran 467 tests in 704.377s

FAILED (errors=4, skipped=319)
```

Running with python3 gives me much more of the type:
```
ERROR: testMountingXFS (formats_test.selinux_test.SELinuxContextTestCase)
XFS does not have a lost+found directory.
  File "/root/blivet/tests/formats_test/selinux_test.py", line 67, in testMountingXFS
    an_fs = fs.XFS(device=self.loopDevices[0], label="test")
  File "/root/blivet/blivet/formats/fs.py", line 101, in __init__
    DeviceFormat.__init__(self, **kwargs)
  File "/root/blivet/blivet/formats/__init__.py", line 186, in __init__
    self.device = kwargs.get("device")
  File "/root/blivet/blivet/formats/__init__.py", line 315, in <lambda>
    lambda f,d: f._setDevice(d),
  File "/root/blivet/blivet/formats/__init__.py", line 306, in _setDevice
    error_msg = self._deviceCheck(devspec)
  File "/root/blivet/blivet/formats/__init__.py", line 301, in _deviceCheck
    if devspec and not devspec.startswith("/"):
TypeError: startswith first arg must be bytes or a tuple of bytes, not str

I'm running on a recent F22 system upgraded to packages from the Rawhide repo. Let me know how would you like to proceed. 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,11 @@ RELEASE_TAG=$(PKGNAME)-$(VERSION)-$(RELEASE)
 VERSION_TAG=$(PKGNAME)-$(VERSION)
 
 PYTHON=python2
+COVERAGE=coverage
+ifeq ($(PYTHON),python3)
+  COVERAGE=coverage3
+endif
+
 ZANATA_PULL_ARGS = --transdir ./po/
 ZANATA_PUSH_ARGS = --srcdir ./po/ --push-type source --force
 
@@ -26,17 +31,26 @@ po-empty:
 		exit 1 ; \
 	done
 
-test:
-	@echo "*** Running unittests ***"
+check-requires:
+	@echo "*** Checking if the dependencies required for testing and analysis are available ***"
+	@grep "^Requires:" python-blivet.spec | cut -f2 -d: | cut -f1 -d">" | xargs rpm -q
+	@rpm -q python-mock python3-mock
+	@rpm -q cryptsetup-python cryptsetup-python3
+	@rpm -q python3-gobject
+	@rpm -q python-coverage python3-coverage
+	@rpm -q xfsprogs hfsplus-tools
+	@rpm -q python3-pocketlint
+
+test: check-requires
+	@echo "*** Running unittests with $(PYTHON) ***"
 	PYTHONPATH=.:tests/ $(PYTHON) -m unittest discover -v -s tests/ -p '*_test.py'
 
-coverage:
-	@which coverage || (echo "*** Please install python-coverage ***"; exit 2)
-	@echo "*** Running unittests with coverage ***"
-	PYTHONPATH=.:tests/ coverage run --branch -m unittest discover -v -s tests/ -p '*_test.py'
-	coverage report --include="blivet/*"
+coverage: check-requires
+	@echo "*** Running unittests with $(COVERAGE) for $(PYTHON) ***"
+	PYTHONPATH=.:tests/ $(COVERAGE) run --branch -m unittest discover -v -s tests/ -p '*_test.py'
+	$(COVERAGE) report --include="blivet/*" --show-missing
 
-check:
+check: check-requires
 	PYTHONPATH=. tests/pylint/runpylint.py
 
 clean:


jkonecny12:
I don't like to see 200+ changes in 27 files in only 1 commit. I think it would be better to split this to more commits.

atodorov:
@jkonecny12 ok, splitting them is easy, how about the general approach and the TODO items ?

dwlehman:
Splitting into Makefile changes, #! changes, and BlivetTestCase changes would be very helpful indeed.

mulkieran:
Please use "with self.assertRaises" everywhere. It's just easier on the eyes.

mulkieran:
I think you should be able to avoid modifying the class hierarchy to deal with the versioning problem in commit 3. You might consider modifying unittest.TestCase instead, for example.

Also, we always use six.PY2 and six.PY3 to check the python version for this kind of thing, not sys.version_info, so it's probably best to stick with that.

mulkieran:
It might be a good idea to make a separate issue for the python 2 tests that are failing and another for the python3 tests that are failing. Feel free to assign the Python 2 tests one to me.

atodorov:
@mulkieran - I can't assign issues to people. See https://github.com/rhinstaller/blivet/issues/130

atodorov:
Everyone, I've split this PR into 3 parts: 
PR #132 is for #! changes
PR #133 is for deprecation warnings
and this one (PR #120) is left with the Makefile changes. 

This should make it easier for reviewers and easier to merge b/c the changes are fairly self contained. Sorry for the long thread above.

vpodzime:
This looks good to me.

mulkieran:
The Makefile already has a PYTHON variable to distinguish between python versions. Is it possible to make the Makefile parameterized on the PYTHON version instead of having duplicate targets for each version? I noticed that you have a tests target that depends on tests for version 2 and tests for version 3, but it would seem nicer to do something like:
```
make PYTHON_VERSION=2 test
make PYTHON_VERSION=3 test
```
in test harness then have these duplicate targets in Makefile.

Unsetting ACK because this feels like an important question to address.

atodorov:
This should take care of it. Example usage:

```
make test
make PYTHON=python2 test
make PYTHON=python3 test

make coverage
make PYTHON=python2 coverage
make PYTHON=python3 coverage
```

atodorov:
@mulkieran - I did a small update and removed python-bugzilla and python-pyblock, these seem unnecessary at the moment and we can always add more dependency checking later if necessary.

mulkieran:
Thanks! I just posted a slightly amended version...would you be just as happy with #151?

mulkieran:
It looks like everyone is OK with #151, so closing this one.

PR #121, opened 2015-05-21
lvm: Add missing argument to libblockdev API for thinpool

cgwalters:
It takes a boolean:
```
 * @discard: whether discard/TRIM is required to be supported or not
```

So I don't know offhand whether anything in anaconda has support for
forcing TRIM support on or not.  `git grep -i trim` in both blivet and
anaconda has no obvious hits, so I'm going to go ahead and assume no.
In that case, let's pass `False` here to avoid blowing chunks.

NOTE: Not actually tested, but I was trying to do thinp via kickstart
with a rawhide installer and saw a traceback for this, and it looks
right to me.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -1016,7 +1016,7 @@ class LVMThinPoolDevice(LVMLogicalVolumeDevice):
             raise ValueError("invalid metadatasize value")
 
         if chunksize is not None and \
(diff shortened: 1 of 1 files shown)


vpodzime:
This looks good to me, but I think a better solution would be to add a default value of ``False`` to the ``@discard`` parameter in libblockdev's Python overrides as that's what we do for many other functions.

PR #123, opened 2015-05-22
Master keepers

mulkieran:
The first commit is a bug fix. The rest are a bunch of useful fixes, i.e., keepers from a separate branch that made me notice the bug in the first place.

PR #124, opened 2015-05-22
Master keepers b

mulkieran:
This is an alternative to #123. I think it's better because the regular expression is simpler and its purpose is clearer. Passes all the same tests. The only difference is in the "Fix a bug..." commit.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -178,7 +178,29 @@ def parseSpec(spec):
     if radix != '.':
         spec = spec.replace(radix, '.')
 
-    m = re.match(r'(?P<numeric>(-|\+)?\s*(?P<base>([0-9]+)|([0-9]*\.[0-9]+)|([0-9]+\.[0-9]*))(?P<exp>(e|E)(-|\+)[0-9]+)?)\s*(?P<rest>[^\s]*)$', spec.strip())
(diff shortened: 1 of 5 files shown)


vpodzime:
The RE in this version is better, but I don't understand why do don't just split the numeric and "unit" parts and then try to create a ``Decimal()`` instance from the numeric part. If that works, we are good, if not, we are doomed anyway. I think we are adding artificial restrictions caused by our custom, buggy-or-not RE what we care for is already covered somewhere else -- by the ``Decimal()`` constructor.

mulkieran:
That's exactly what I'm trying to do. It's what I explained in the comment. The problem is in making sure that the numeric part doesn't get confused with a unit spec, since they both can contain non-numbers.

The best design initially would have been to require whitespace between magnitude and unit specification. Then [^\s]+\s+[^\s]*, or something similar, would have covered all situations and Decimal constructor, etc. could have done all the remaining heavy lifting. Unfortunately, because we allow specs like "1KiB" (no space to separate magnitude from unit spec), that is just not an option, AFAIK.

If it is an option, I'm delighted, and happy to ditch the current regular expression for a much simpler one.

mulkieran:
Superseded by #129.

PR #125, opened 2015-05-25
Ignore and clean files generated by Sphinx

atodorov:
Just cleaning up a bit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,9 @@
 .*.swp
 blivet*.tar.gz
 blivet*.tar.bz2
-doc/blivet.devicelibs.rst
-doc/blivet.formats.rst
-doc/blivet.rst
-doc/modules.rst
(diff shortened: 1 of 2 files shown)


mulkieran:
AFAICT, almost all the .rst files are ignored. Might be better to ignore all .rst files and use
! to leave out the ones that should not be ignored, like intro.rst and index.rst in .gitignore file.

Might be better to autogenerate most of the module names that are deleted in Makefile, too.

atodorov:
.gitignore should look better now but I have no idea how would you like to autogenerate the module names in the Makefile. Any suggestions ?

mulkieran:
using wildcards and things. It is probably not worth the trouble at this time. I'll ack it as is...

atodorov:
With #144 in the works I guess this can be closed. Reopen if you think otherwise.

PR #126, opened 2015-05-25
Put translations into a subpackage shared by Python 2 and 3

vpodzime:
With providing both Python 2 and Python 3 packages we need to make sure both
have the translations available. This commit makes the translations shared by
both Python 2 and 3 subpackages by putting them into a separate subpackage
required by both.

### DIFF
--- a/python-blivet.spec
+++ b/python-blivet.spec
@@ -45,10 +45,18 @@ Requires: libblockdev-plugins-all >= %{libblockdevver}
 Requires: libselinux-python
 Requires: python-hawkey
 Requires: pygobject3-base
+Requires: %{realname}-data = %{epoch}:%{version}-%{release}
 
 %description
 The python-blivet package is a python module for examining and modifying
 storage configuration.
+
+%package -n %{realname}-data
+Summary: Data for the %{realname} python module.
+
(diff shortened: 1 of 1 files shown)


dashea:
Shouldn't the Requires for blivet-data be matched by version?

This is kind of one of those things where it would be nice if the Fedora package committee came up with some recommendations, because all the solutions are haphazard and gross. I'm not real keen on the idea of a new package to begin with, and if you match the python3?-blivet version to the blivet-data version, which I think it should be, then if you have python-blivet and python3-blivet on the same system then those versions would need to match too, which isn't terrible but is kind of weird.

The solution a lot of packages seem to be using, including pykickstart, is to install the .mo files into the python directory, something like site-packages/blivet/locale.

vpodzime:
Yeah, I think the version should be matched. I'll fix this as I think it's a "lesser evil" when compared to the version mismatch between the data package and the code.

Generally, the idea is to have these files in the system only once. And if we need some more "non-code" stuff in the future (like a config file or something), it should go into this subpackage.

vpodzime:
A new version rebased to master and correctly requiring the exact same version (epoch:version-release) of the blivet-data package.

PR #127, opened 2015-05-25
Adapt to the new libblockdev initialization API

vpodzime:
In version 1.0 the libblockdev library changes the initialization functions a
bit and thanks to that we may simplify our code.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -67,20 +67,13 @@ from gi.repository import GLib
 from gi.repository import BlockDev as blockdev
 _REQUESTED_PLUGIN_NAMES = set(("lvm", "btrfs", "swap", "crypto", "loop", "mdraid", "mpath", "dm"))
 _requested_plugins = blockdev.plugin_specs_from_names(_REQUESTED_PLUGIN_NAMES)
-if not blockdev.is_initialized():
-    try:
-        blockdev.try_init(require_plugins=_requested_plugins, log_func=log_bd_message)
-    except GLib.GError:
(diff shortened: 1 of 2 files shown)


blivet/__init__.py
```diff
@@ -67,20 +67,13 @@ def cb(self, exn):
 from gi.repository import BlockDev as blockdev
 _REQUESTED_PLUGIN_NAMES = set(("lvm", "btrfs", "swap", "crypto", "loop", "mdraid", "mpath", "dm"))
 _requested_plugins = blockdev.plugin_specs_from_names(_REQUESTED_PLUGIN_NAMES)
-if not blockdev.is_initialized():
-    try:
-        blockdev.try_init(require_plugins=_requested_plugins, log_func=log_bd_message)
-    except GLib.GError:
-        pass
+try:
+    succ, avail_plugs = blockdev.try_reinit(require_plugins=_requested_plugins, reload=False, log_func=log_bd_message)
```
mulkieran: succ is ignored, so it should be underscored.

mulkieran:
Other than comment about "succ", it looks fine to me.

blivet/__init__.py
vpodzime: Good catch, thanks! I originally meant to use it to skip the set operations and checks, but I think it's clearer to just compare the sets. Fixing.

vpodzime:
Added the underscore to the ``succ`` variable and bumped the required version of libblockdev to 1.0.

PR #128, opened 2015-05-25
Include automatic and manual test documentation.

atodorov:
This is my first attempt to bring some test documentation to blivet before actually trying to write any more tests. Summary of changes:

- modify existing Makefile target to include test documentation in the main document
- add dependency on Graphviz which is used to create inheritance graphs for test cases (used to outline which are the base test classes in Blivet). WARNING: may break Red The Docs if they don't have graphviz installed. Possibly needs a BuildRequires change for the spec file as well.
- update .gitignore
- add tests/README.rst with basic description of blivet's test suite.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -43,8 +43,12 @@ help:
 
 clean:
 	-rm -rf $(BUILDDIR)/*
+	-rm -f $(addprefix ./, tests.*)
 
 apidoc:
+	-rm -f $(addprefix ./, tests.* modules.rst)
+	$(SPHINXAPIDOC) -o . ../tests/
+	-cp modules.rst tests.modules.rst
 	-rm $(addprefix ./, $(MODULE_NAMES))
 	$(SPHINXAPIDOC) -o . $(SOURCEDIR)
 
--- a/tests/README.rst
+++ b/tests/README.rst
@@ -1,0 +1,88 @@
+Testing Blivet
+==============
+
+Note: The test suite documented here is available only from the git repository
+not as part of any installable packages.
+
+In order to execute blivet's test suite from inside the source directory execute
+the command::
+
+    make test
+
+Tests descending from :class:`~.imagebackedtestcase.ImageBackedTestCase` or
+:class:`~.loopbackedtestcase.LoopBackedTestCase` require root access on the
+system and will be skipped if you're running as non-root user.
+Tests descending from :class:`~.imagebackedtestcase.ImageBackedTestCase` will
+also be skipped if the environment variable JENKINS_HOME is not defined. If
+you'd like to execute them use the following commands (as root)::
+
+    # export JENKINS_HOME=`pwd`
+    # make test
+
+To execute the Pylint code analysis tool run::
+
+    make check
+
+Running Pylint doesn't require root privileges but requires Python3 due to usage
+of pocket-lint.
+
+It is also possible to generate test coverage reports using the Python coverage
+tool. To do that execute::
+
+    make coverage
+
+It is also possible to check all external links in the documentation for
+integrity. To do this::
+
+    cd doc/
+    make linkcheck
+
+Test Suite Architecture
+------------------------
+
+Blivet's test suite relies on several base classes listed below. All test cases
+inherit from them.
+
+- :class:`unittest.TestCase` - the standard unit test class in Python.
+  Used for tests which don't touch disk space;
+
+
+- :class:`~tests.storagetestcase.StorageTestCase` - intended as a base class for
+  higher-level tests. Most of what it does is stub out operations that touch
+  disks. Currently it is only used in
+  :class:`~tests.action_test.DeviceActionTestCase`;
+
+
+- :class:`~tests.loopbackedtestcase.LoopBackedTestCase` and
+  :class:`~tests.imagebackedtestcase.ImageBackedTestCase` - both classes
+  represent actual storage space.
+  :class:`~tests.imagebackedtestcase.ImageBackedTestCase` uses the same stacks
+  as anaconda disk image installations. These mimic normal disks more closely
+  than using loop devices directly. Usually
+  :class:`~tests.loopbackedtestcase.LoopBackedTestCase` is used for stacks of
+  limited depth (eg: md array with two loop members) and
+  :class:`~tests.imagebackedtestcase.ImageBackedTestCase` for stacks of greater
+  or arbitrary depth.
+
+
+In order to get a high level view of how test classes inherit from each other
+you can generate an inheritance diagram::
+
+    PYTHONPATH=.:tests/ python3-pyreverse -p "Blivet_Tests" -o svg -SkAmy tests/
+
+
+Alternatively you may try::
+
+    cd doc/
+    make clean html \
+    SPHINXOPTS="                                                         \
+        -D extensions=sphinx.ext.inheritance_diagram,sphinx.ext.graphviz \
+        -D inheritance_graph_attrs.rankdir=LR                            \
+        -D inheritance_graph_attrs.size=''"
+
+
+and view the `_build/html/inheritance.html` document.
+
+Note: pyreverse is part of pylint while sphinx.ext.inheritance_diagram is part
+of Sphinx itself. Both should be installed on your system if you're building
+Blivet from source.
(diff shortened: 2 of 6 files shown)


tests/README.rst
```diff
@@ -0,0 +1,45 @@
+Testing Blivet
+==============
+
+Note: The test suite documented here is available only from the git repository
+not as part of any installable packages.
+
+In order to execute blivet's test suite you need root access on the system.
```
mulkieran: This is not really true. Only some tests require root access. Many fairly useful ones don't.

tests/README.rst
```diff
@@ -0,0 +1,45 @@
...
+From inside the source directory execute the commands::
+
+    make test
+    make check
```
mulkieran: I never use the check target, because I have to extend my PYTHONPATH with a bunch of directories, including pocketlint. But YMMV.

Also, you absolutely don't need to be root to run pylint.

mulkieran:
Thanks for the work!

The "Testing Blivet" section is buried way down below the list of blivet packages. Is there a way to organize that list in the main contents pane so that it is a little more visible or maybe to organize the table of contents on the left pane better so it is clear that it is there?

I'm not sure that putting sphinx code in tests/README.rst is absolutely the right way to go, unless that's already an established sort of thing to do.

I'm not absolutely sure that having those dependence graphs included is the correct way to go.
I'm not sure they add much information and they cause a bit of trouble. It is neat that it is possible, though, but maybe it's something that should be left for later.

atodorov:
README.rst inline comments fixed with new version.

atodorov:
> The "Testing Blivet" section is buried way down below the list of blivet packages. Is there a way to organize that list in the main contents pane so that it is a little more visible or maybe to organize  the table of contents on the left pane better so it is clear that it is there?

I have no idea about the left pane but we could either alter  toctree maxdepth in index.rst or change the order of topics in the document. Updating maxdepth to 2 makes the testing section more visible but hides all the submodules from the main page. The alternative is to move Testing Blivet above the generated API docs.


> I'm not sure that putting sphinx code in tests/README.rst is absolutely the right way to go, unless that's already an established sort of thing to do.

I'm not quire sure either but see below to understand my concerns. Given that this is test documentation I'm OK with the trade-offs.

> I'm not absolutely sure that having those dependence graphs included is the correct way to go.
I'm not sure they add much information and they cause a bit of trouble. It is neat that it is possible, though, but maybe it's something that should be left for later.

I first started with the idea of writing new tests for blivet and quickly realized I have no idea how to execute the test suite (and all of its requirements) and how the tests were organized. The class diagram gave me a good overview of how things are related to each other and which were the base classes to look at if I want to create new tests.  And like most Linux users I'm not using an IDE which provides this info for me.  The Sphynx diagrams are the easiest way to provide an overview of how the test suite is structured internally. This also revealed some possible issues (see https://www.redhat.com/archives/anaconda-devel-list/2015-May/msg00006.html).

Also see - https://github.com/sphinx-doc/sphinx/pull/1902. Once accepted we should be able to generate smaller inheritance diagrams with better visibility which will result in better documentation.

mulkieran:
If you include testing.rst in intro.rst and leave testing out of index.rst the arrangement in the main pane and TOC in the left pane is a bit better, I think.

Note that pyreverse, which is part of pylint package, will generate class inheritance diagram for you. That's what I generally use.

It would actually be nice to have class inheritance diagrams in read the docs generally, but all of blivet is a pretty monstrous diagram, so including them would require some judgement. It's nice that it is available.

atodorov:
I've tried pyreverse, it has some advantages but I'm not satisfied with the results. 

The SVG graphics are rendered in horrible order and take too much space on the screen. And the graph seems to be incorrect. I've found at least one class, which inherits from another one and these two are displayed as stand-alone nodes in the graph. 

My command was (am I doing it wrong):
```
[ tests]# python3-pyreverse -p "Blivet_Tests" -o svg -m y -A -k .
```

Then check the graph for StorageTestCase and DeviceActionTestCase. There should be a link between them.

One thing we can do is separate these diagrams from the rest of the docs and let people build them by hand if required.

mulkieran:
I usually output pyreverse to dot format, then I can edit the file to get the layout I want. pyreverse's default layout seems to be BT. sphinx's seems to be LR, which is better for these purposes.

I also find missing links in the graph, though. In my graph, LoopBackedTestCase appears to have no children, apparently because all its children reside in subdirectories. This feels wrong, definitely. pyreverse is very lightweight, mostly because it uses astng to understand the python source.

Making generating the graphs separately might be the best idea, for now.

atodorov:
> pyreverse's default layout seems to be BT. sphinx's seems to be LR, which is better for these purposes.

For the record: indeed sphinx uses LR by default but that is configurable. 

Let me see if I can come with a clean solution of generating the inheritance graph manually.

atodorov:
Several notes related to the latest update:

* If we pass the correct PYTHONPATH to pyreverse it is able to properly display the links between inherited classes (the problem was bacisally that we do from tests.xxx import yyy). However pyreverse now fails with https://bitbucket.org/logilab/pylint/issue/543/

* pyreverse images continue to look worse compared to Sphinx generated ones.

* I've added notes how to build the inheritance diagram is necessary.

* Due to limitations in Sphinx I need to add a new config file and rst file displaying the inheritance diagrams. This rst file is built with make html as well (I don't see any way to exclude it or ignore it) which produces (as expected) warnings that inheritance-diagram is not recognized as valid directive.

IMO we can't make this any better at the moment without resorting to additional tools or heavily modifying the existing ones. I also think that despite the warnings this change will not break Read The Docs or any other docs builder so it should be safe to merge as is.

Let me know what you think.

doc/Makefile
```diff
@@ -43,11 +43,16 @@ help:
...
+	-rm -f $(addprefix ./, tests.*)
 
 apidoc:
+	-rm -f $(addprefix ./, tests.* modules.rst)
+	$(SPHINXAPIDOC) -o . ../tests/
+	-cp modules.rst tests.modules.rst
 	-rm $(addprefix ./, $(MODULE_NAMES))
 	$(SPHINXAPIDOC) -o . $(SOURCEDIR)
 
+
```
mulkieran: accidental extra line.

doc/inheritance/conf.py
```diff
@@ -0,0 +1,12 @@
+import os
+import sys
+sys.path.append(os.path.abspath('..'))
+
+from ..conf import *
```
mulkieran: We always reject wildcard imports.

doc/inheritance/conf.py
```diff
@@ -0,0 +1,12 @@
...
+
+sys.path.insert(0, os.path.abspath('../..'))
+
+for ext in ['sphinx.ext.inheritance_diagram', 'sphinx.ext.graphviz']:
+    extensions.append(ext)
```
mulkieran: ```extensions.extend(('sphinx.ext.inheritance_diagram', 'sphinx.ext.graphviz'))``` is probably better.

doc/inheritance/conf.py
atodorov: Make an exception. I don't see any other way to import all of the config variables and override only the extensions.

mulkieran:
You could override extensions with the sphinx-build -D flag and a Makefile variable if it weren't for this bug: https://github.com/sphinx-doc/sphinx/pull/1911.

Given that this extra conf.py file is a workaround for a bug in sphinx that may be fixed...could you separate the changes that are part of the workaround into a separate commit, and explain all this in a separate commit message?

mulkieran:
Or maybe it would be better just to drop the inheritance subdirectory entirely, really.

The real problem is just that, if dot is missing, sphinx will put a big ugly glob of dot output in place of the nice image that we would like to see. It doesn't raise an error, AFAICT on my system.
I've posted a sphinx issue about this: https://github.com/sphinx-doc/sphinx/issues/1912 and they may do something to change what they do now.

But for now, it seems better to me not to complicate our code by working around what is essentially a sphinx bug in this fashion.

atodorov:
@mulkieran please see the latest changes. I'm assuming those sphinx issues will be fixed sooner rather than later.

mulkieran:
Suggest #144 instead.

mulkieran:
Going with #144 instead.

PR #132, opened 2015-05-27
Remove /usr/bin/python shebang

atodorov:
*  replaces /usr/bin/python with /usr/bin/env python inside test cases. NOTE: I don't see any executable tests/*.py files but maybe this is a good measure. Otherwise this line doesn't have any effect and can be removed.

There are still a few places which use hard-coded python version:

*    Makefile uses PYTHON=python2 which is used in a few places calling setup.py
*    some scripts/ use /usr/bin/python explicitly
*    runpylint.py uses /usr/bin/python3 explicitly

### DIFF
--- a/tests/action_test.py
+++ b/tests/action_test.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 
 import unittest
 
@@ -1162,7 +1161,4 @@ class DeviceActionTestCase(StorageTestCase):
     def testActionSorting(self, *args, **kwargs):
         """ Verify correct functioning of action sorting. """
         pass
-
-if __name__ == "__main__":
-    unittest.main()
 
--- a/tests/devicefactory_test.py
+++ b/tests/devicefactory_test.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 
 import unittest
 
@@ -59,5 +58,3 @@ class MDFactoryTestCase(unittest.TestCase):
 
         self.assertIsNone(self.factory2.get_container())
 
-if __name__ == "__main__":
-    unittest.main()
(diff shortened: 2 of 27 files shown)


mulkieran:
This is fine with me. I don't mind keeping the shebang line...but I'ld be just as happy to get rid of it, I guess.

What's the purpose of that extra chunk of code in raid_test.py?

atodorov:
> What's the purpose of that extra chunk of code in raid_test.py?

So you can do
```
PYTHONPATH=.:tests/ $(PYTHON) tests/devicelibs_test/raid_test.py
```
instead of running all tests together. It is present in other files as well, I just found raid_test.py was missing it while I was working on it.

mulkieran:
Aaaah. Now I understand your commit message better. I didn't realize that chunk of code was present in the other files.

This is related to the change I suggesed for #133 just now.

I think we should actually get rid of the shebangs and ```if __name__ == "__main__"``` chunks in all the *_test.py files (and all others if there are any).

The reason for this is that we always want to run the tests with the 2/3 compatible TestCase functions.
Therefore, we always want to run them in a context where a compatibility module has been loaded. Running them directly from the command line is likely to skip that, so we should get rid of code that makes it easier to run them from the command line.

mulkieran:
I'm assuming you got them all, ACK.

vpodzime:
The commit message shouldn't end with ';' and it should be updated to better describe the current set of changes.

mulkieran:
Please rebase on master and remove that added line in raid_test.py and I'll go ahead and merge this PR.

atodorov:
Should be good to go

mulkieran:
PR #148 is just a minor cleanup of this pull request.

PR #134, opened 2015-05-27
Small tests/ fixes

atodorov:
Following up with a PR after the IRC discussion.


mulkieran:
Except please extend commit message to mention that tests/lib was previously used for pylint and now pocketlint is used instead.

atodorov:
@mulkieran - I've just noticed a tests/testenv.sh file which also appears to be obsolete. Should I go ahead and remove it too ?

mulkieran:
Yes! Thanks for noticing.

mulkieran:
Superseded by #143.

PR #136, opened 2015-05-28
Drop check from the release build target

bcl:
It has become impossible to do builds for rawhide from previous releases
because the system is expected to match what it's being built for. Run
make check manually when making commits instead.


dwlehman:
Could we add all the deps as `BuildRequires` and run `make check` as part of the rpm build?

dwlehman:
As a short term solution this patch is fine.

PR #137, opened 2015-05-28
Use default md metadata version for everything except /boot/efi

dwlehman:
We began using this metadata version because grub1 could not read arrays with v1.1 or newer metadata. Now that we're using grub2 we can stop using v1.0. The UEFI bootloader still needs us to use v1.0 for /boot/efi arrays.

See https://bugzilla.redhat.com/show_bug.cgi?id=1061711#c6

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -22,7 +22,7 @@
 
 import copy
 
-from .deviceaction import ActionCreateDevice, ActionDestroyDevice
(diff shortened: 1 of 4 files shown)


dwlehman:
I see that github has reversed the order of the commits. "Use the default..." should be first.

mulkieran:
Looked at this before and it seems like all my questions were answered back then, so ack.

PR #138, opened 2015-05-28
Allow partitioning of md arrays whose members are all disks

dwlehman:
The motivation here is to enable hardware vendors to use regular md arrays for fwraid without the need for mdmonitor. The difficulty is that there will be nothing to distinguish these fwraid arrays from any other user-created array whose members are disks.

Existing arrays whose members/parents are all partitionable will be partitionable.
Existing arrays whose members/parents are all disks will be treated as disks.

It shouldn't be possible to delete such arrays in anaconda. They should be treated like fwraid arrays. It will be possible to delete them in blivet by scheduling an `ActionDestroyFormat`/`ActionDestroyDevice` action pair, but calling `DeviceTree.recursiveRemove` on them will not destroy the array.

### DIFF
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -78,6 +78,19 @@ class MDRaidArrayDevice(ContainerDevice, RaidDevice):
             :type metadataVersion: str (eg: "0.90")
             :keyword minor: the device minor (obsolete?)
             :type minor: int
+
+            .. note::
+
+                An instance of this class whose :attr:`exists` attribute is
+                True and whose parent/member devices are all partitionable is
+                also considered to be partitionable.
+
+            .. note::
+
+                An instance of this class whose :attr:`exists` attribute is
+                True and whose parent/member devices are all disks is also
+                treated like a disk.
+
         """
         # pylint: disable=unused-argument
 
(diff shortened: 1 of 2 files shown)


bcl:
The code looks ok to me. I think you should add docstrings on the new methods/attrs.

Also, is it possible that it could end up not being partitionable and not a disk? eg. if some parents are disks and some are partitions both methods will return false. Is there protection to make sure all the parents are consistent?

dwlehman:
Having parents of mixed types is allowed, but precludes treating the array as partitionable or as a disk.

vpodzime:
Other than the nitpick abovet this looks good to me.

vathpela:
I think you've got this right - if we preserve the capabilities that all of the constituents share, we should get a reasonable thing out. If any constituent doesn't have some capability, the resulting device doesn't have it.

dwlehman:
I found what I think is a better implementation of the second patch. What do you think?

  https://gist.github.com/dwlehman/d9263e8b180c189d9362

I would like to replace the second patch ("Do format handling when...") with the equivalent of the above gist (ported to populator.py instead of devicetree.py).

vpodzime:
> I found what I think is a better implementation of the second patch. What do you think?
> 
> https://gist.github.com/dwlehman/d9263e8b180c189d9362
> 
> I would like to replace the second patch ("Do format handling when...") with the equivalent of the above gist (ported to populator.py instead of devicetree.py).

The latter approach looks better to me. However, I'd change the name of the parameter to ``updateOrigFmt`` as that's what it causes/does.

mulkieran:
When is it wrong to update the origFormat from the format when building a device tree? In other words, why do you need that parameter at all?

dwlehman:
It is possible to call `DeviceTree.populate` at any point, including after scheduling a reformat. It is important that originalFormat actually be the original format and not just the format it had when `addUdevDevice` ran last. The most obvious example of this is with network-attached block devices that get added one at a time from anaconda.

PR #140, opened 2015-05-28
Rhel7 rebase mybugs

bcl:
A few patches that needed to be cherry-picked.


dwlehman:
I think I just pushed the zanata one.

bcl:
pushed

PR #144, opened 2015-06-02
Include automatic and manual test documentation.

mulkieran:
* No longer explicitly remove files in apidoc target, all handled in clean.
* Put autogenerated .rst files in subdirectories and gitignore them.
* Create a hidden file which includes some inheritance diagrams.
* Add commented out option for inheritance graph in conf.py.
* Include documentation for test modules at top level.
* Include documentation about testing in the introduction.
* Get rid of unnecessary MODULE_NAMES variable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,8 @@
 .*.swp
 blivet*.tar.gz
 blivet*.tar.bz2
-doc/blivet.devicelibs.rst
-doc/blivet.formats.rst
-doc/blivet.rst
-doc/modules.rst
+doc/blivet
+doc/tests
 doc/_build
 po/*.gmo
 po/*.po
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -8,7 +8,7 @@ SPHINXAPIDOC  = sphinx-apidoc
 PAPER         =
 BUILDDIR      = _build
 SOURCEDIR     = ../blivet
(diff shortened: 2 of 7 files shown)


mulkieran:
An alternative to #128.

@atodorov, I'ld like to push this instead of #128. I'ld rather not tell users that the -D flag works for ```extensions``` when it doesn't. I feel that the inheritance diagrams serve mostly as notes for you, so hiding it, and making it easy for you to generate if you want to seems like the right decision. I did not include ```inheritance_graph_attrs.*``` specs in conf.py file because they were unrecognized.

After noticing how inheritance-diagram notations don't mix well with figures I'm beginning to feel that the decision to include graphics in what is, after all, restructured _text_ was a bad idea and should not be exploited. If we ever want to produce lovely documentation with beautiful figures, latex and LyX are really what we should be using.

atodorov:
ack, let's go ahead with this one. 

> After noticing how inheritance-diagram notations don't mix well with figures I'm beginning to feel that the decision to include graphics in what is, after all, restructured text was a bad idea and should not be exploited. If we ever want to produce lovely documentation with beautiful figures, latex and LyX are really what we should be using.

I would love to see a text based diagram with links to each class but that will likely require a different tool. I will examine once more as I go along updating the test docs in the future.

mulkieran:
ok, great. I will merge this tomorrow unless somebody brings up a further point.

PR #145, opened 2015-06-02
rhel7 cherry picks

mulkieran:
Original bz's are:  1019685, 1090009.

bz#1151649 is already in as bz#1150147
bz#1155151 is already in as bz#1156202.
bz#1090009 required some extra follow up patches


vpodzime:
These all look good to me.

mulkieran:
Fix a typo...BTRFSError should have been BTRFSValueError in "Add BTRFSValueError...". That's all.

PR #146, opened 2015-06-03
RHEL 7.2 cherry-picks

vpodzime:
I think it's easier to review the whole resulting diff as things move back and forth in the commits. One question -- is it correct that all the commits have 'Related: rhbz#1202877' instead of the original bugs referenced?


vpodzime:
Added the original bug IDs back.

PR #147, opened 2015-06-03
Fix setting original size for format resize action. (#1225352)

vojtechtrefny:
Simple fix for https://bugzilla.redhat.com/show_bug.cgi?id=1225352 Cancelling of ActionResizeFormat fails because the origSize is set to format.targetSize that can be 0.


vpodzime:
Looks good to me.

mulkieran:
Please add an inline comment explaining what makes those new lines correct.

vojtechtrefny:
Comment added. Manually pushed to f22-branch too.

PR #148, opened 2015-06-04
Remove facilities for running tests as standalone modules.

mulkieran:
* Remove #!/usr/bin/python shebang from tests.
* Remove unittest.main() from tests.
* Remove a definition of suite method, which we no longer use.

The tests should be run via the unittest discovery mechanism.
They are best run by invoking the "test" target in the Makefile.

This allows any necessary setup to occur, including overrides
for Python2/3 compatibility.

### DIFF
--- a/tests/action_test.py
+++ b/tests/action_test.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python
 
 import unittest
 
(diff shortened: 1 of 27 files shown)


mulkieran:
Supercedes #132.

Only changes are:
* Removes a bunch of trailing newlines.
* Removes a definition of ```suite``` method in parentlist_test.

vpodzime:
Looks good to me.

PR #149, opened 2015-06-04
Make blivet aware of internal LVs

vpodzime:
This is really more a request for comments than a real pull request as I'd like to know your opinion on the chosen approach. Making blivet aware of internal LVs is one of the crucial steps towards creation and manipulation of thin/cache pools' metadata volumes and other things.

The internal LVs are not part of the DeviceTree, they are only referenced by their "parent" LVs as they should not be manipulated with directly and because there may be many of them (tens on quite simple setups). However, these LVs are still just LVs like any other as far as things like name, UUID, size, etc. go.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -48,6 +48,8 @@ from .container import ContainerDevice
 from .dm import DMDevice
 from .md import MDRaidArrayDevice
 
+INTERNAL_LV_CLASSES = []
+
 class LVMVolumeGroupDevice(ContainerDevice):
     """ An LVM Volume Group """
     _type = "lvmvg"
@@ -495,9 +497,6 @@ class LVMLogicalVolumeDevice(DMDevice):
                           exists=exists)
 
         self.uuid = uuid
-        self.copies = copies
-        self.logSize = logSize or Size(0)
-        self.metaDataSize = Size(0)
         self.segType = segType or "linear"
         self.snapshots = []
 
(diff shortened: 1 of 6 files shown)


dwlehman:
I think this is generally a good approach. The internal LVs are represented, but because they aren't in the devicetree they do not complicate things on its level.

vpodzime:
Pushed a version that should be close to the final revision of this change. I did some manual testing on a VM with quite nice setup (4 LVs + ~20 internal LVs) and all seems to work as expected.

vpodzime:
Changed order of the last two patches and improved the second one. Should be ready for a review now.

dwlehman:
Have you run pylint on the tree with your changes applied?

vpodzime:
> Have you run pylint on the tree with your changes applied?

Yes, but I only run it from time to time because on my Fedora 21 I get quite many false positives and pylint runs for a long time. I'll definitely run it and take care of all real issues before I merge this. The review of the idea, approach and implementation strategy is what the first required step, I think.

vpodzime:
New version with the suggestions and comments addressed. Internal LVs now have their ``format`` attribute set to ``DeviceFormat`` instances and I also added some local disables for pylint so that it doesn't complain about bad ``super()`` calls (intentional, explained in the comments) etc.

Also rebased to current master.

blivet/devices/lvm.py
```diff
@@ -32,7 +32,7 @@
 
 from .. import errors
 from .. import util
-from ..formats import getFormat
+from ..formats import getFormat, DeviceFormat
```
dwlehman: This import is unused.

dwlehman:
Looks good to me aside from the one unused import.

PR #151, opened 2015-06-08
Parameterize Makefile test targets on Python version.

mulkieran:
Check that dependencies for testing are available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,10 +7,24 @@ RELEASE_TAG=$(PKGNAME)-$(VERSION)-$(RELEASE)
 VERSION_TAG=$(PKGNAME)-$(VERSION)
 
 PYTHON=python2
+COVERAGE=coverage
+ifeq ($(PYTHON),python3)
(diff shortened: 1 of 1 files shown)


mulkieran:
This is a slightly amended version of #120.

The only difference is that test dependencies are stored in variable TEST_DEPENDENCIES and that sort and uniq are used to drop duplicates while non-recursive assignment is used to calculate value of TEST_DEPENDENCIES. This is marginally better because it reports _all_ uninstalled packages in an orderly way. The upside of storing this all in a variable is that the test dependencies only need to be calculated once; the downside is that they need to be calculated at least once always.

Note that even if running tests for python 2, python 3 dependencies are checked. It's hard to get around that w/out doing something really different than grepping the spec file for '^Requires:' lines, and I'm not sure what that really different thing would be.

atodorov:
ack on my side.

vpodzime:
Looks good to me too.

PR #155, opened 2015-06-10
Sort the lists before comparing because coversion to set reverses order in Python3

### LINKED ISSUE #154: Hard to reproduce failure in tests/devices_test/device_packages_test.py - DevicePackagesTestCase
With Python3 I sometimes get this error:
```
FAIL: testPackages (device_packages_test.DevicePackagesTestCase)
  File "/var/www/html/blivet/tests/devices_test/device_packages_test.py", line 26, in testPackages
    self.assertListEqual(packages, list(set(packages)))
AssertionError: Lists differ: ['cryptsetup', 'mdadm'] != ['mdadm', 'cryptsetup']

First differing element 0:
cryptsetup
mdadm

- ['cryptsetup', 'mdadm']
+ ['mdadm', 'cryptsetup']
```
I've managed to isolate the issue down to the conversion from list to set and then back to list.

Initially packages is ['cryptsetup', 'mdadm'], then set(packages) becomes {'mdadm', 'cryptsetup'} which later becomes the list ['mdadm', 'cryptsetup'] and the assert fails. Converting to set under Python3 sometimes reverses the order of the elements but not always. Any ideas how to reproduce reliably or why that happens ?

atodorov:
Closes #154

### DIFF
--- a/tests/devices_test/device_packages_test.py
+++ b/tests/devices_test/device_packages_test.py
@@ -20,7 +20,7 @@ class DevicePackagesTestCase(unittest.TestCase):
         packages = luks.packages
 
         # no duplicates in list of packages
-        self.assertListEqual(packages, list(set(packages)))
+        self.assertEqual(len(packages), len(set(packages)))
 
         # several packages that ought to be included are
         for package in dev1.packages + dev2.packages + dev.packages:


tests/devices_test/device_packages_test.py
```diff
@@ -18,9 +18,12 @@ def testPackages(self):
         dev = MDRaidArrayDevice("dev", level="raid1", parents=[dev1,dev2])
         luks = LUKSDevice("luks", parents=[dev])
         packages = luks.packages
+        packages.sort() # sort due to blivet/issues/154
 
         # no duplicates in list of packages
-        self.assertListEqual(packages, list(set(packages)))
+        l = list(set(packages))
+        l.sort() # sort due to blivet/issues/154
+        self.assertListEqual(packages, l)
```
vpodzime: Why don't we just compare the sets? The order of the packages doesn't matter to us, AFAICT.

atodorov: We have a list, not set. We convert it to set to remove all duplicates, then back to list to compare it to the original. Now this can be done in other ways as #154 suggests, but there aren't two sets which you can compare to begin with.

vpodzime: > We have a list, not set. We convert it to set to remove all duplicates, then back to list to compare it to the original. Now this can be done in other ways as #154 suggests, but there aren't two sets which you can compare to begin with.

We can convert both to sets and compare the sets.

atodorov: > We can convert both to sets and compare the sets.

I fail to see how this can be done. 
```
>>> set([1, 2, 3, 2, 1]) == set([1, 2, 3])
True
```

How do you figure out if the first set had duplicate elements in in, please explain ?

mulkieran:
The only purpose of that line is to check that there are no dups. We can trust, hopefully, that Python set() works. So,

```self.assertEqual(len(packages), len(set(packages)))```

should do the trick. I have no idea why I used ```assertListEqual``` in the first place.

**Comment by mulkieran** [comment]
We try to use a pretty standard format for our commit messages. So, something like:

```
Use len of set to check for duplicates in list of packages.

Resolves: #154.

Checking for equality of the two lists was a mistake, since the order of the list generated
from the set is undefined.
```

is better. Please update the commit message, and I can push right away.

PR #156, opened 2015-06-12
LVM cache detection, manipulation and reporting

vpodzime:
These commits add generic classes for Cache monitoring, manipulation and reporting, the implementation for the LVM Cache and pieces of code needed for the ``LVMLogicalVolumeDevice`` class to support cached LVs.

The reason why cached LV is not a new, separate class is that when the cache is detached, such (formerly) cached LV becomes a normal LV like any other and its not possible to change the type/class of an existing object. Thus the ``LVMCache`` class is added instead the instances of which are referenced by cached LVs and provide the cache-related functionality to them. The same approach will be later used for ``BcacheDevice`` and ``BcacheCache`` classes (a device can be formatted as a ``BcacheDevice`` with no cache attached to it).

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -532,6 +532,7 @@ class LVMLogicalVolumeDevice(DMDevice):
 
         self._metaDataSize = Size(0)
         self._internal_lvs = []
+        self._cache = None
 
     def _check_parents(self):
         """Check that this device has parents as expected"""
@@ -622,8 +623,12 @@ class LVMLogicalVolumeDevice(DMDevice):
     @property
     def vgSpaceUsed(self):
         """ Space occupied by this LV, not including snapshots. """
+        if self.cached:
+            cache_size = self.cache.size
+        else:
+            cache_size = Size(0)
         return (self.vg.align(self.size, roundup=True) * self.copies
-                + self.logSize + self.metaDataSize)
+                + self.logSize + self.metaDataSize + cache_size)
 
     @property
     def vg(self):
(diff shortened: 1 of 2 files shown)


jkonecny12:
I think this is ok.
I have one question wouldn't be better if you move some code to that abstract base classes? Now they are empty and you will end up with the parts of the same code in all child classes I think.

vpodzime:
> I have one question wouldn't be better if you move some code to that abstract base classes? Now they are empty and you will end up with the parts of the same code in all child classes I think.

Any example? The problem is that e.g. the [stats reported for Bcache] (http://rhinstaller.github.io/libblockdev/libblockdev-KernelBlockDevices.html#BDKBDBcacheStats) look quite different from [stats reported for the LVM cache] (http://rhinstaller.github.io/libblockdev/libblockdev-LVM.html#BDLVMCacheStats).

jkonecny12:
I'm not so familiar with the problematic to tell you. 
I think the best will be to wait and you will see that when you will implement the other classes.

jkonecny12:
I'm no blivet expert but I think this is ok :)

blivet/devices/lvm.py
```diff
@@ -847,6 +859,28 @@ def removeInternalLV(self, int_lv):
...
+        if self.exists and not self._cache:
+            # check if we have a cache pool internal LV
+            pool = None
+            for lv in self._internal_lvs:
+                if isinstance(lv, LVMCachePoolLogicalVolumeDevice):
+                    pool = lv
+
+            self._cache = LVMCache(self, size=pool.size, exists=True)
+
+        return self._cache
```
dwlehman: This always creates an `LVMCache`, no? So then the `cached` property above will always be True for existing LVs. I think it just needs an `if pool is not None:` guarding the assignment to `self._cache`.

dwlehman:
Other than the minor problem in `LVMLogicalVolumeDevice.cache` this looks good to me.

blivet/devices/lvm.py
vpodzime: Wow, good catch, thanks!

PR #160, opened 2015-06-17
Deprecate some little-used device lookup methods

dwlehman:
`getDevicesBySerial` has no callers and the other two don't offer any value beyond their list comprehension equivalents `[d for d in dt.devices if isinstance(d, cls)]` (`getDevicesByInstance`) and `[d for d in dt.devices if d.type == devtype]` (`getDevicesByType`).

The first patch establishes a basic procedure for deprecating things in blivet. The warnings will only be logged if someone has called `blivet.util.set_up_logging`.

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -518,6 +518,11 @@ def set_up_logging(log_file='/tmp/blivet.log'):
     handler.setFormatter(formatter)
     log.addHandler(handler)
     program_log.addHandler(handler)
+
+    # capture python warnings in our logs
+    warning_log = logging.getLogger("py.warnings")
+    warning_log.addHandler(handler)
+
     log.info("sys.argv = %s", sys.argv)
 
 def create_sparse_tempfile(name, size):
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -11,9 +11,11 @@ import sys
 import tempfile
 import uuid
 import hashlib
+import warnings
 from decimal import Decimal
 from contextlib import contextmanager
 from gi.repository import BlockDev as blockdev
+from functools import wraps
 
 import six
 
@@ -24,7 +26,6 @@ program_log = logging.getLogger("program")
 from threading import Lock
 # this will get set to anaconda's program_log_lock in enable_installer_mode
 program_log_lock = Lock()
-
 
 def _run_program(argv, root='/', stdin=None, env_prune=None, stderr_to_stdout=False, binary_output=False):
     if env_prune is None:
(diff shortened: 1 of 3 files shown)


vpodzime:
This is much nicer than I expected. ACK.

snbueno:
I agree, this is nice and simple. Do you think it makes sense to change the comment from "$function will be removed in a future version" to specify which future version? (If it's always going to be the next major version, at least; and it might be easier for us to keep track of as well when we go back through and do the actual deletions.)

dwlehman:
I didn't think we made a final decision regarding how long to wait before removing deprecated methods, so I intentionally left it vague. We talked about removing it in the next major version, but also about leaving it in until the second time we increment the major version (to make sure people have a chance to see the deprecation warnings). I also noticed that most other python code that does this does not identify a specific version for the removal.

vpodzime:
I'm definitely for removing stuff in the second future major version from where it got deprecated. Just like @dwlehman says, we should give people a chance to see the warrnings.

dwlehman:
I considered a decorator but since it doesn't handle the sphinx part I decided against it.

mulkieran:
Twisted's deprecated decorator handles the docstring part as well, by mangling the ```__doc__``` attribute. But that kind of thing is ugly, so I didn't want to suggest it.

But it still seems like a decorator that handled the warnings would be helpful for encouraging uniformity and recognizability and probably checkability.

I think that you could deprecate a class by deprecating its constructor using the same method as for deprecating methods.

mulkieran:
And, while munging the docstring may be ugly, at least sphinx will generate pretty documentation from the munged doc string. So, it's a better idea than it seemed at first.

dwlehman:
Yeah, I was looking into having the decorator munge the docstring. That would make it worth doing IMO, and I don't really care if it's ugly or not.

dwlehman:
New version handles issuing the warning and munging the doc text for sphinx. Usage can be seen in dwlehman@ed736ab.

blivet/util.py
```diff
@@ -637,3 +643,105 @@ def eintr_retry_call(func, *args):
...
             raise
+
+#
+# Deprecation decorator.
+#
+_DEPRECATION_MESSAGE = "will be removed in a future version."
+def _default_deprecation_msg(func):
+    return "%s %s" % (func.__name__, _DEPRECATION_MESSAGE)
+
+def indent(text, spaces=4):
```
bcl: You can use textwrap.indent for this (on py3)

blivet/util.py
```diff
@@ -637,3 +643,105 @@ def eintr_retry_call(func, *args):
...
+        base_text = " " # They contain leading and trailing spaces. *shrug*
+    else:
+        base_text = base_text[:-1] # Trim the trailing space.
+
+    if ".. deprecated::" in base_text:
+        # Don't add multiple deprecation directives.
+        return
+
+    # Figure out the number of spaces to indent docstring text.
+    indent_spaces = None
```
bcl: Why don't you just set this to 0 and skip all the None checks?

bcl:
I'm a bit worried about the extra overhead added to the calls, but since they're deprecated I suppose that doesn't matter.

blivet/util.py
dwlehman: I had myself turned around at some point, clearly, thinking that would prevent it from ever getting set to a non-default value. I'll change it.

blivet/util.py
dwlehman: Looks like I'll have to use the TextWrapper class if I want it to be portable to python2. I'll look into that.

dwlehman:
Note that the decorator is only applied once, so the only part that runs on every call to a deprecated method is the warning.warn call. The docstring hack runs once per method.

blivet/util.py
dwlehman: I was right to begin with. If indent_spaces starts at 0 that if statement will never be True and indent_spaces will always end up being 0. (The idea there is to get the minimum indentation in the docstring and use that for the sphinx directive.)

blivet/util.py
```diff
@@ -637,3 +643,105 @@ def eintr_retry_call(func, *args):
...
+        if not l.strip():
+            continue
+
+        spaces = 0
+        _l = l[:]
+        while _l and _l.startswith(" "):
+            spaces += 1
+            _l = _l[1:]
+
+        if indent_spaces is None or indent_spaces > spaces:
```
bcl: Ah! ok, maybe a note here that it's finding the minumum amount of indent.

PR #162, opened 2015-06-18
Enforce and document the fact that StorageDevice.format is always a DeviceFormat.

dwlehman:
StorageDevice.format should never be anything other than an instance
of DeviceFormat or one of its subclasses. All block devices contain data
that is arranged according to some format. The case of empty or
uninitialized or unrecognized formatting is represented by the base class
DeviceFormat.

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -29,7 +29,7 @@ from .. import util
 from ..flags import flags
 from ..storage_log import log_method_call
 from .. import udev
-from ..formats import getFormat
+from ..formats import getFormat, DeviceFormat
 from ..size import Size
 
 import logging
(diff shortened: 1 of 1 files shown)


blivet/devices/storage.py
```diff
@@ -623,6 +643,14 @@ def _updateNetDevMountOption(self):
             self._format.options = ",".join(option_list)
 
     def _getFormat(self):
+        """ Get the device's format instance.
+
```
mulkieran: Might as well docstring it all up with ```:rtype:``` and ```:return:``` while you are here.

mulkieran:
Other than neatpick is fine with me.

blivet/devices/storage.py
```diff
@@ -592,9 +592,29 @@ def status(self):
...
+            A value of None will effectively mark the device as unformatted,
+            but this is accomplished by setting it to an instance of the base
+            :class:`~.formats.DeviceFormat` class.
+
+            .. note::
+                :attr:`format` should always be an instance of
+                :class:`~.formats.DeviceFormat`. To ensure this continues to be
+                the case, all subclasses that define their own :attr:`format`
+                getter should call :meth:`StorageDevice._setFormat` from their
+                getter.
```
vpodzime: Should this talk about **setter** rather than **getter**? If not, I'm a bit confused here.

dwlehman: Indeed it should. Thanks.

PR #163, opened 2015-06-18
LUKS resize

mulkieran:
This is a request for comment, because it is incomplete. I am trying to figure out better ways to test the higher-level functionality but am having problems obtaining a cooperating virtual machine.

There are a few points:
* What makes this a bit tricky is that LUKS is handled as both a device and a format. There is a sense
in which it is both.
* I decided to start from the bottom up this time, working on resizing the format first, rather than worrying about the Actions.
* If the resize of a LUKSDevice is handled as just a resize of its slave's LUKS format then blivet.resizeDevice() does not need to change at all to handle LUKS devices AFAICT.
* However, it does not make much sense to resize a LUKS format outside the LUKS device...since the format knows nothing about the size of the, e.g., filesystem that it encrypts. And yet, it is still possible. Not sure whether it is necessary or desirable to do anything about that.
* I have not handled the dependencies between the actions.

### DIFF
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -778,7 +778,7 @@ class ActionResizeFormat(DeviceAction):
             callbacks.resize_format_pre(ResizeFormatPreData(msg))
 
         self.device.setup(orig=True)
-        self.device.format.doResize()
+        self.device.format.resize()
 
         if callbacks and callbacks.resize_format_post:
             msg = _("Resized filesystem on %(device)s") % {"device": self.device.path}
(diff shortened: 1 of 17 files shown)


mulkieran:
I could make the commit message for aaaeede more explicit, something like:

anaconda's pyananaconda.errors.ErrorHandler class defines an error handler
_fsResizeHandler(), and a lookup function, cb(), which does lookup on this precise error type.
The error handler just provides a little more information, and then re-raises the exception.
The errorHandler, and the lookup for FSResizeError should both be removed at some point,
as they will have become dead as a result of this patch.
However, no anaconda error should result from this patch,
and the anaconda spec does not need to be updated for this patch.

Also, since we're moving into deprecation for real, it might make sense to deprecate FSResizeError class.

vojtechtrefny:
Luks resize implemented in PR #308

PR #165, opened 2015-06-24
Master format copy

mulkieran:
A shallow copy of an FS object leaves all the new object's task objects referring to the copied object.
This patch set should fix that.

The first two patches enforce that an FSTask has only one instance attribute, its FS object, which is important for the fix, and also gets rid of some boilerplate initializers that existed in the expectation that there might someday be more instance attributes in a filesystem task.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -100,19 +100,7 @@ class FS(DeviceFormat):
 
         DeviceFormat.__init__(self, **kwargs)
 
-        # Create task objects
-        self._info = self._infoClass(self)
-        self._fsck = self._fsckClass(self)
-        self._mkfs = self._mkfsClass(self)
-        self._mount = self._mountClass(self)
-        self._readlabel = self._readlabelClass(self)
-        self._resize = self._resizeClass(self)
-        self._sync = self._syncClass(self)
-        self._writelabel = self._writelabelClass(self)
-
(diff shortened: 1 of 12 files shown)


dwlehman:
I know we discussed this before you did it, but I wonder if the correct change is to instead change all places where we do a shallow copy on a format to do a deep copy and leave the formats and tasks themselves as they are. After all, this would make a shallow copy of a format more like a deep copy, wouldn't it?

mulkieran:
The tasks objects are somewhat special, because they refer back to their FS object. So, there could be some reason for treating them especially.

In favor of using deepcopy everywhere:
Doing so may correct some subtle logical bugs in our code having nothing to do with tasks.
It avoids implementing a ```__copy__``` method, which may itself be buggy.

Against using deepcopy everywhere:
It is expensive and probably a bit buggy itself.
We may find that we have to implement a version of ```__deepcopy__``` and then use variable_copy(), like in some of the Device classes. This would happen if traversal reaches some kind of object that can  not itself be deepcopied.

I think it is a difficult call. But the conservative approach is to try using deepcopy everywhere, knowing that the other approach is a viable alternative. This way, no new code is added to solve the problem.

The first two patches are probably worth considering separately. They discard a bunch of boilerplate which I think has been revealed as quite pointless.

mulkieran:
Superseded by #167.

PR #167, opened 2015-06-24
Master format copy b

mulkieran:
This is an alternative to #165. The approach is to deepcopy all format objects when copying, instead of to define an ```__copy__``` method that handles tasks specially.

The first two patches are the same as in #165.

It's tested, in that I tried an experiment whereby I made a deep copy of all formats objects in fstesting.py and ran all tests on the deepcopy, rather than the original format object, and they all passed. Using copy, instead of deepcopy, they all failed. Using copy, and overriding ```__copy__``` in FS, they all passed.

It should be noted that format objects are already deep copied, rather than shallow copied, in actionlist.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -1121,7 +1121,7 @@ class Blivet(object):
             self.devicetree.cancelAction(action)
 
         # make sure any random overridden attributes are reset
-        device.format = copy.copy(device.originalFormat)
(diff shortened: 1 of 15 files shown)


vpodzime:
Looks good to me and I like it more than PR #165

dwlehman:
Looks good.

PR #168, opened 2015-06-26
iscsi and multiprocessing

dashea:
Anaconda got stuck because the libiscsi module wasn't installed on my boot.iso and we can fail better than that.


vpodzime:
Looks good to me too.

dashea:
Pushed.

PR #169, opened 2015-06-29
Make check-requires errors more readable.

dashea:
Only output the parts about packages that are not installed.


vpodzime:
Looks good to me too.

PR #170, opened 2015-07-01
Clean up device size properties a bit

dwlehman:
This fixes several minor problems and adds some unit tests to ensure the correct behaviors are preserved in the future.

We shouldn't be setting the target size from the size setter under any circumstances. The size reflects reality for existing devices and the planned reality for non-existent devices. Target size is only for resize of existing devices. The setter for one should not modify the other.


vpodzime:
Apart from the comment above this looks good to me. It's definitely an improvement.

PR #171, opened 2015-07-02
x-initrd.mount should only be set for /var (#1238603)

bcl:
Only add it to /var and only when /var is on a _netdev.

Resolves: rhbz#1238603


vpodzime:
Looks good to me.

bcl:
pushed.

PR #172, opened 2015-07-08
Add a bunch more gi.require_version calls

dashea:
Choo choo here comes the crud train


clumens:
All aboard!

dashea:
chugga chugga chugga chugga wooooo wooooo

PR #174, opened 2015-07-14
Translate vendor id 0x1af4 to Virtio Block Device (#1242117)

bcl:
The vendor information used to come from parted which reports virtio blk
devices as 'Virtio Block Device'. The sysfs vendor entry has always been
0x1af4 which isn't very descriptive. So in disk.description check for
the exception and return "Virtio Block Device". disk.vendor is left
as-is, only the user friendly description is modified.

Resolves: rhbz#1242117

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -111,7 +111,11 @@ class DiskDevice(StorageDevice):
 
     @property
     def description(self):
-        return " ".join(s for s in (self.vendor, self.model) if s)
+        # On Virtio block devices the vendor is 0x1af4, make it more friendly
+        if self.vendor == "0x1af4":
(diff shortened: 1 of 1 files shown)


dwlehman:
But why should we do this for virtio instead of having them report a sane vendor? How many blocks like this are we going to end up with for various drivers/vendors?

bcl:
> But why should we do this for virtio instead of having them report a sane vendor? How many blocks like this are we going to end up with for various drivers/vendors?

Good point, but note that we already override description for subclasses so it isn't a unique situation.

vpodzime:
I agree with @dwlehman that this should be fixed in the kernel. I'm okay with this as a workaround as long as the bug report is filed for the kernel to fix this nonsense.

PR #178, opened 2015-07-23
Add method to list primary partitions

vojtechtrefny:
We already have extendedPartition and logicalPartitions, having also primaryPartitions would make blivet-gui code a little nicer and shorter.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -324,6 +324,15 @@ class DiskLabel(DeviceFormat):
         return logicals
 
     @property
+    def primaryPartitions(self):
+        try:
+            primaries = self.partedDisk.getPrimaryPartitions()
+        except Exception: # pylint: disable=broad-except
+            log_exception_info()
+            primaries = []
+        return primaries
+
+    @property
     def firstPartition(self):
         try:
             part = self.partedDisk.getFirstPartition()


blivet/formats/disklabel.py
```diff
@@ -324,6 +324,15 @@ def logicalPartitions(self):
         return logicals
 
     @property
+    def primaryPartitions(self):
+        try:
+            primaries = self.partedDisk.getPrimaryPartitions()
```
vpodzime: I think this could be simplified to ``return self.partedDisk.getPrimaryPartitions()`` with the ``primaries`` variable being dropped.

blivet/formats/disklabel.py
```diff
@@ -324,6 +324,15 @@ def logicalPartitions(self):
...
+        except Exception: # pylint: disable=broad-except
```
vpodzime: Do we really need this broad exception? It seems to me like it will be hiding real bugs in the code.

vojtechtrefny: I donť know, I basically took the logicalPartitions function and just changed logical to primary.

dwlehman: He was just following the style of the existing, related properties. We should probably drop the try/except on all three of them, but I have some concerns about doing that in a minor version release.

vpodzime: > We should probably drop the try/except on all three of them, but I have some concerns about doing that in a minor version release.

Works for me.

vojtechtrefny:
Pushed.

PR #179, opened 2015-07-24
More followups for size property changes

dwlehman:
@kellinm The ones you're interested in are a5000d4, a800128, and b8b2c48


vpodzime:
These all look good to me.

PR #180, opened 2015-07-27
Require pygobject3-base instead of pygobject (#1246842)

dashea:
The package with the normal looking name is now the one with the cairo
crud for some reason.


dashea:
Pushed.

PR #181, opened 2015-07-28
Create md.py

cinquecento:
The EXTLINUX MD RAID1 metadata 1.0 support, complements
 the EXTLINUX MD RAID1 support in the Anaconda installer.
For those running the machines on the BIOSes, baremetal and virtual.
Resolves: rhbz#1246829


dwlehman:
I think something went wrong when you were preparing your pull request. Check the commits. There should be one commit in this case with a small patch to blivet/devices/md.py, while your commits completely remove that file and then re-add it.

PR #182, opened 2015-07-28
device factory bug fixes

dwlehman:
These were turned up in the course of writing unit tests for `DeviceFactory` (PR forthcoming).

PR #183, opened 2015-07-28
Various small test changes

dwlehman:
Improvement of loop device cleanup and disabling of image-backed tests are so we can start running the test suite in jenkins again, @clumens.

The device factory tests fail without the changes from #182. They increase coverage of devicefactory.py from 19% to 74% (IIRC).


dwlehman:
It's expected that several devicefactory tests fail without #182.

vpodzime:
Looks good to me otherwise.

PR #184, opened 2015-07-28
Misc 20150728

dwlehman:
These are just some random things I had laying around.


vpodzime:
These all look good to me.

PR #187, opened 2015-07-29
Fixes needed for livemedia-creator and anaconda --dirinstall

bcl:
Skipping device teardown is a bit of a hack, but it solves the problem for the moment. We need to look at how mountpoints are being handled and fix that at some point soon. It is nuts that repeated calls to teardown can result in different behavior.


vpodzime:
These both look good to me.

bcl:
pushed

PR #188, opened 2015-07-30
Pass a sysfs path to MultipathDevice constructor (#1245201)

rvykydal:
As dlehman advised, I

Cherry-picked master
commit d2a27294a05b0c34bb6a69d3393cda1d17ae0a7

And tested that it works.

Resolves: rhbz#1245201


rvykydal:
Pushed

PR #190, opened 2015-07-31
LVM cache creation

vpodzime:
Commits that add support for creation of cached LVs plus two minor changes that make things a bit nicer. Please note that I'm intentionally leaving the changes in the last commit as a standalone commit as it needs some attention.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -917,10 +917,12 @@ class Blivet(object):
         """
         thin_volume = kwargs.pop("thin_volume", False)
         thin_pool = kwargs.pop("thin_pool", False)
-        vg = kwargs.get("parents", [None])[0]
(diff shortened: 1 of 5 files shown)


vpodzime:
Fixed a pylint warning.

vpodzime:
New version using PV devices instead of just their names.

vojtechtrefny:
Looks good to me.

dwlehman:
Some unit tests to go along with this would be really nice.

vpodzime:
> Some unit tests to go along with this would be really nice.

I'll definitely implement some once this gets merged and has some "final" state. I first wanted to catch up with rhel7-branch and its support for LVM cache. Now I can work on both unit tests and kickstart tests for anaconda.

PR #191, opened 2015-07-31
PartitionDevice may not have a disk set (#1248973)

bcl:
When it is an implicit partition and hasn't been assigned a disk yet
self.disk may be None, so skip checking the size against the disk when
there isn't one.

Resolves: rhbz#1248973


vpodzime:
Looks good to me.

bcl:
pushed.

PR #192, opened 2015-08-03
Fix UnboundLocalError in FSMinSize (#1249304)

vojtechtrefny:


blivet/tasks/fsminsize.py
```diff
@@ -54,7 +54,7 @@ def _getResizeInfo(self):
         except OSError as e:
             error_msg = "failed to gather info from resize program: %s" % e
         if rc:
-            error_msg = "failed to gather info from resize program: %s" % e
+            error_msg = "failed to gather info from resize program: %s" % rc
```
vpodzime: This should probably rather use ``%d``, but it's not a big deal.

vpodzime:
Looks good to me either way.

blivet/tasks/fsminsize.py
```diff
@@ -54,7 +54,7 @@ def _getResizeInfo(self):
         except OSError as e:
             error_msg = "failed to gather info from resize program: %s" % e
         if rc:
-            error_msg = "failed to gather info from resize program: %s" % e
+            error_msg = "failed to gather info from resize program: %d" % rc
```
bcl: Seems like this would be better inside the try: just so it's clear it only runs when it succeeds.

vojtechtrefny: Makes sense. Fixed.

bcl:
Looks good to me.

vojtechtrefny:
Pushed.

PR #193, opened 2015-08-04
md fwraid detection

dwlehman:
This resolves various problems encountered trying to install f23 alpha on md fwraid, beginning with [this bug](https://bugzilla.redhat.com/show_bug.cgi?id=1245191). "Minor cleanup..." doesn't need to go on f23-branch since it doesn't really fix anything.

### DIFF
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -628,6 +628,14 @@ class MDContainerDevice(MDRaidArrayDevice):
         # (the device node does not allow read / write calls)
         return False
 
+    @property
+    def isDisk(self):
+        return False
+
+    @property
+    def partitionable(self):
+        return False
+
 class MDBiosRaidArrayDevice(MDRaidArrayDevice):
 
     _type = "mdbiosraidarray"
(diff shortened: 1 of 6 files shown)


blivet/devices/md.py
```diff
@@ -640,6 +651,15 @@ def __init__(self, name, **kwargs):
         self._size = self.currentSize
 
     @property
+    def isDisk(self):
+        # pylint: disable=bad-super-call
+        return super(MDRaidArrayDevice, self).isDisk
+
+    def partitionable(self):
```
bcl: Forgot @property here. But I also don't see why they are needed, it'll just call super w/o them defined in the subclass.

vpodzime:
Same opinion as @bcl about the properties, but otherwise this looks good to me.

dwlehman:
Those super calls bypass `MDRaidArrayDevice` and use the properties from `StorageDevice` instead, which is good since the ones in `MDRaidArrayDevice` should not be used for the fwraid arrays. It might be clearer if I just made them return `True` explicitly since that's what they'll do:

```python
@property
def isDisk(self):
    return True

@property
def partitionable(self):
    return True
```

vpodzime:
I'm more for adding a comment explaining what's going on there.

PR #194, opened 2015-08-04
Fix mount cache for devices with symlinks

dwlehman:
This makes it possible to use any valid device spec when accessing the cache. All device specs are resolved to the basename canonical device node name (eg: `dm-1`, `md127p2`) before being stored or search for.

It turns out that `/proc/mounts` contains `/dev/mapper/` symlinks for device-mapper devices, but does not contain `/dev/md/` symlinks for md devices. We avoid this circus by using only the canonical device node name in the cache itself. We use udev to resolve device specs to the canonical names.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -68,12 +68,14 @@ def trigger(subsystem=None, action="add", name=None):
     util.run_program(["udevadm"] + argv)
     settle()
 
-def resolve_devspec(devspec):
+def resolve_devspec(devspec, sysname=False):
     if not devspec:
         return None
 
     # import devices locally to avoid cyclic import (devices <-> udev)
     from . import devices
+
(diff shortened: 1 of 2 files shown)


dwlehman:
This resolves fedora bug [1247803](https://bugzilla.redhat.com/show_bug.cgi?id=1247803).

blivet/udev.py
```diff
@@ -99,7 +101,7 @@ def resolve_devspec(devspec):
                     break
 
     if ret:
-        return device_get_name(ret)
+        return ret.sys_name if sysname else device_get_name(ret)
```
bcl: I still don't like embedding the if like this. I've gotten used to X or Y syntax, but I think this form makes what is happening less clear.

bcl:
Looks good other than the 1 nitpick.

PR #196, opened 2015-08-07
Fix _unalignedMaxPartSize for logical partitions (#1250890)

vojtechtrefny:
This fixes two problems: free space inside extended partition was
not recognized and free space outside of extended partition was
considered as valid for resizing.

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -755,14 +755,22 @@ class PartitionDevice(StorageDevice):
         #     partition on disk. We don't care about leading free space --
         #     a filesystem cannot be relocated, so if you want to use space
         #     before and after your partition, remove it and create a new one.
-        sector = self.partedPartition.geometry.end + 1
         maxPartSize = self.size
+
+        if self.isLogical:
+            # logical partition is at the very end of the extended partition
+            extended = self.partedPartition.disk.getExtendedPartition()
+            if self.partedPartition.geometry.end == extended.geometry.end:
+                return maxPartSize
+
+        sector = self.partedPartition.geometry.end + 1
         try:
             partition = self.partedPartition.disk.getPartitionBySector(sector)
         except _ped.PartitionException:
             pass
         else:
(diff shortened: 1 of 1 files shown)


blivet/devices/partition.py
```diff
@@ -755,14 +755,23 @@ def _unalignedMaxPartSize(self):
...
+        sector = self.partedPartition.geometry.end + 1
         try:
             partition = self.partedPartition.disk.getPartitionBySector(sector)
         except _ped.PartitionException:
             pass
         else:
-            if partition.type == parted.PARTITION_FREESPACE:
+            # next partition is free space or 'logical' free space
+            if partition.type in (parted.PARTITION_FREESPACE,
+                                  parted.PARTITION_FREESPACE | parted.PARTITION_LOGICAL):
```
vpodzime: I'd consider ``if partition.type & parted.PARTITION_FREESPACE`` better readable, but it could be only me. :)

vpodzime:
Other than the nitpick above, this looks good to me. But I'm sure @dwlehman will suggest you to add some tests ;)

blivet/devices/partition.py
dwlehman: Seconded.

vojtechtrefny: Ok

vojtechtrefny:
Fixed the "if" to be simpler. Not sure about the test, I guess it would require too many Mocks and patches and it would probably test mock not blivet :) I think @KosiehBarter could add some resizing to his blivet tests to test resizing of (logical) partitions.

vpodzime:
> Not sure about the test, I guess it would require too many Mocks and patches and it would probably test mock not blivet :) I think @KosiehBarter could add some resizing to his blivet tests to test resizing of (logical) partitions.

Works for me. Thanks for changing that if!

vojtechtrefny:
Pushed to master and f23-branch.

PR #198, opened 2015-08-10
Partition requests may not have partedPartition (#1248973)

bcl:
Check for it and return, assuming the new size will be assigned to a
disk with enough space.

Resolves: rhbz#1248973


vpodzime:
Looks good to me.

PR #199, opened 2015-08-11
Add unit tests for the LVM cache support

vpodzime:
So that we can catch possible regressions easily in the future.


dwlehman:
Please also add assert statements to some of the other lvm tests to ensure they correctly report themselves as not being cached -- maybe device_properties_test is a good place for that.

vpodzime:
That's a good point, will do.

vpodzime:
Added one more check to the LVM cache support unit test (``vgSpaceUsed``), one fix and tests that make sure both cached and not-cached LVs report expected values of required properties.

vojtechtrefny:
Looks good to me.

PR #200, opened 2015-08-11
fcoe: replace fipvlan with fcoemon (#1085325)

rvykydal:
Same as dracut (#1129888), we are moving to use fcoemon which provides more
robust approach than fipvlan.

Resolves: rhbz#1085325

### DIFF
--- a/blivet/fcoe.py
+++ b/blivet/fcoe.py
@@ -23,6 +23,7 @@ from . import util
 #from pyanaconda import isys
 import logging
 import time
+import shutil
 from .i18n import _
 log = logging.getLogger("blivet")
 
@@ -96,7 +97,7 @@ class fcoe(object):
         if self.lldpadStarted:
             return
 
-        util.run_program(["lldpad", "-d"])
+        util.run_program(["systemctl", "start", "lldpad.service"])
         self.lldpadStarted = True
 
     def addSan(self, nic, dcb=False, auto_vlan=True):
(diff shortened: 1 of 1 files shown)


blivet/fcoe.py
```diff
@@ -110,27 +111,69 @@ def addSan(self, nic, dcb=False, auto_vlan=True):
...
             self._startLldpad()
-            util.run_program(["dcbtool", "sc", nic, "dcb", "on"])
-            util.run_program(["dcbtool", "sc", nic, "app:fcoe",
-                                                "e:1", "a:1", "w:1"])
-            rc, out = util.run_program_and_capture_output(["fipvlan", "-c", "-s", "-f",
-                                               "-fcoe", nic])
+            while timeout > 0:
+                rc, out = util.run_program_and_capture_output(
+                    ["lldptool", "-p"])
+                if not rc:
```
vpodzime: I prefer using ``if rc != 0`` in these checks as I always has to think about it otherwise. Unless you're strongly against that, please replace these conditions with ``if rc != 0``.

blivet/fcoe.py
```diff
@@ -110,27 +111,69 @@ def addSan(self, nic, dcb=False, auto_vlan=True):
...
+                timeout -= 1
+                time.sleep(1)
+
+            while timeout > 0:
+                rc, out = util.run_program_and_capture_output(
+                    ["dcbtool", "sc", nic, "app:fcoe", "e:1", "a:1", "w:1"])
+                if not rc:
+                    break
+                time.sleep(1)
+                log.info("Retrying to set up dcb for fcoe")
```
vpodzime: Looks like this "wait until utility exits successfully or time out" could be a useful function probably in ``iutil.py`` as it may be useful for other modules too.

blivet/fcoe.py
```diff
@@ -110,27 +111,69 @@ def addSan(self, nic, dcb=False, auto_vlan=True):
...
-                # certain network configrations require the VLAN layer module:
+            dpath = os.readlink("/sys/class/net/%s/device/driver" % nic)
+            driver = os.path.basename(dpath)
+            if driver == "bnx2x":
+                util.run_program(["ip", "link", "set", nic, "up"])
                 util.run_program(["modprobe", "8021q"])
-                rc, out = util.run_program_and_capture_output(["fipvlan", '-c', '-s', '-f',
-                                                   "-fcoe",  nic])
+                udev.settle()
+                time.sleep(3)
```
vpodzime: Is this sleep really needed? And if so, are 3 seconds always going to be enough? I think we at least need some comment here.

vpodzime:
Looks good to me otherwise.

blivet/fcoe.py
rvykydal: Alright, I tracked it back to
https://bugzilla.redhat.com/show_bug.cgi?id=813057#c4
so what about adding
# Sleep for 3 s to allow dcb negotiation (#813057)

blivet/fcoe.py
rvykydal: Note that the timeout is common for a sequence of utilities run. Do you still think such a function could be useful? (It would have crazy amount of arguments.)

blivet/fcoe.py
vpodzime: Works for me.

blivet/fcoe.py
vpodzime: It wouldn't if you could give it something like a ``remaining`` argument that would tell it how much time is still available. But it would have to return the updated value together with other things which would probably make it too complex. So I'm okay with leaving this as it is.

blivet/fcoe.py
```diff
@@ -110,27 +111,69 @@ def addSan(self, nic, dcb=False, auto_vlan=True):
...
-            util.run_program(["dcbtool", "sc", nic, "app:fcoe",
-                                                "e:1", "a:1", "w:1"])
-            rc, out = util.run_program_and_capture_output(["fipvlan", "-c", "-s", "-f",
-                                               "-fcoe", nic])
+            while timeout > 0:
+                rc, out = util.run_program_and_capture_output(
+                    ["lldptool", "-p"])
+                if not rc:
+                    break
+                log.info("Waiting for lldpad to be ready")
```
bcl: these are going to clutter up the logs, maybe only log if the timeout is exceeded. while / else could be used for that.

blivet/fcoe.py
```diff
@@ -110,27 +111,69 @@ def addSan(self, nic, dcb=False, auto_vlan=True):
...
+                timeout -= 1
+                time.sleep(1)
+
+            while timeout > 0:
```
bcl: timeout isn't reset between these blocks.

blivet/fcoe.py
```diff
@@ -145,28 +188,46 @@ def write(self, root):
...
-            os.close(fd)
+        # Done before packages are installed so don't call
+        # write_nic_fcoe_cfg in target root but just copy the cfgs
+        shutil.copytree("/etc/fcoe", root + "/etc/fcoe")
 
-        return
+    def write_nic_fcoe_cfg(self, nic, dcb=True, auto_vlan=True, enable=True, mode=None, root=""):
+        cfg_dir = root + "/etc/fcoe"
+        example_cfg = os.path.join(cfg_dir, "cfg-ethx")
+        if os.access(example_cfg, os.R_OK):
```
bcl: Looks like blivet needs simpleconfig from anaconda.

blivet/fcoe.py
```diff
@@ -145,28 +188,46 @@ def write(self, root):
...
+            lines = open(example_cfg, "r").readlines()
+        else:
+            anaconda_cfg = """FCOE_ENABLE="yes"\n
+DCB_REQUIRED="yes"\n
+AUTO_VLAN="yes"\n
+MODE="fabric"\n
+"""
```
jkonecny12: Could you please give here better indentation. Maybe concatenation of strings will be better here for readability.

blivet/fcoe.py
rvykydal: it is intentional, we want to wait 60 secs maximum for all the tools run (as Chris Leech pointed out when reviewing my former version which had 60s timeout for each tool)

blivet/fcoe.py
rvykydal: Yes, I was missing it too.

blivet/fcoe.py
rvykydal: Yes, I am going to what you suggest, please ignore the patch added to this branch, your way is nicer.

rvykydal: Ah, sorry, actually the patch added is the way I'd go, because eg if the timeout appears in 2nd wile loop, the else branches of the loop *and* all subsequent while loops blocks will get executed.

rvykydal:
I will, thanks.

rvykydal:
Pushed so we can get as early testing as possible.

I addressed these review remarks locally:
- "if not rc" by vpodzime
- comment for time.sleep(3) by vpodzime
- cluttering log messages by bcl
- log message % by clumens
- missing timeout decrement by rvykydal
and tested the patch in beaker.

PR #201, opened 2015-08-12
Use device name from udev only if it's available (#1252052)

vpodzime:
Sometimes we may know about a device udev has no (longer) idea about. For
example if an LV is deactivated and thus it's device node gets destroyed. We
shouldn't replace a valid device spec with None in such cases.


jkonecny12:
Seems good to me

PR #203, opened 2015-08-17
Change labelFormatOK to classmethods

vojtechtrefny:
This allows fslabel validity checks without creating the format.

I think the best solution would be to just remove this tasks and move the labelFormatOK method to blivet.formats.FS class (and its subclasses) similarly to isNameValid method for lvmvg/lvmlv/btrfs names validation but it would require API change.


vpodzime:
Looks good to me and I agree we should get rid of the task for this.

vojtechtrefny:
Pushed.

PR #204, opened 2015-08-17
Add 'build' and '.directory' to gitignore

vojtechtrefny:
Would make 'git add .' a lot easier for me.


vpodzime:
Looks good to me.

vojtechtrefny:
Pushed.

bcl:
> Would make 'git add .' a lot easier for me.

Note that git add -u is better, it will only add things already in the index, not everything new.

PR #205, opened 2015-08-17
Fix copy method (#1254135)

bcl:
removed needs to be setup outside the loop.

Resolves: rhbz#1254135


vpodzime:
Looks good to me too.

bcl:
Pushed.

PR #206, opened 2015-08-17
Add OSError to list of errors in updateSysfsPath (#1252949)

bcl:
One more error that pyudev can raise.

Resolves: rhbz#1252949

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -300,7 +300,7 @@ class StorageDevice(Device):
         # ValueErrors and EnvironmentErrors that may be raised. So catch
         # those errors, but also catch DeviceNotFoundError, which is the
         # appropriate exception and which it may raise in future.
-        except (pyudev.DeviceNotFoundError, EnvironmentError, ValueError) as e:
+        except (pyudev.DeviceNotFoundError, EnvironmentError, ValueError, OSError) as e:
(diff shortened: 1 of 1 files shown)


blivet/devices/storage.py
```diff
@@ -300,7 +300,7 @@ def updateSysfsPath(self):
         # ValueErrors and EnvironmentErrors that may be raised. So catch
         # those errors, but also catch DeviceNotFoundError, which is the
         # appropriate exception and which it may raise in future.
-        except (pyudev.DeviceNotFoundError, EnvironmentError, ValueError) as e:
+        except (pyudev.DeviceNotFoundError, EnvironmentError, ValueError, OSError) as e:
```
vpodzime: I think we should either add ``OSError`` to the comment above the ``except`` line make the comment more generic instead of listing all exceptions that may be raised and duplicating the list there.

vpodzime:
Looks good to me otherwise.

blivet/devices/storage.py
bcl: > I think we should either add OSError to the comment above the except line make the comment more generic instead of listing all exceptions that may be raised and duplicating the list there.

Yeah, seems silly to list them in the comments.

bcl:
Pushed.

Issue #31, opened 2015-03-02
Cleanup/update LVM size calculations and expectations


vpodzime:
Blivet's LVM code contains many (often magic) calculations focused on free space left or required by LVM (e.g. "*How much space is this LV going to take from a disk?*" or "*How much space there will be for LVM on top of RAIDn?*"). Many of these calculations come from the lvm1 (called simply *lvm*) era and many are based on some empirical testing or just experience without any clear inspectable explanation. Also lvm2 has different size limitations than lvm1 used to have so we need to update these pieces of code/numbers too.

bcl:
Ideally this information would be provided by lvm in some way that doesn't require reading its code.

vpodzime:
I think we all dream this dream. But I think (and even more I hope) there are many needless calculations in Blivet that origin from some noise and misunderstandings.

vpodzime:
Done.

PR #207, opened 2015-08-18
Better calculate required space for LVM cache

vpodzime:
This pull request has three purposes (of descending importance):
1) it fixes a hidden issue discovered by anaconda's kickstart tests where the following kickstart snippet
```
logvol swap --name=swap --vgname=fedora --size=500 --fstype=swap
logvol / --name=root --vgname=fedora --size=4000 --grow --fstype=ext4 --cachepvs=pv.2 --cachesize=1000 --cachemode=writethrough
logvol /home --name=home --vgname=fedora --size=1000 --grow --fstype=ext4 --cachepvs=pv.2 --cachesize=1000 --cachemode=writeback
```
results in a successful installation, but the ``swap`` LV ends up being 8 MiB smaller. It's not the end of the world, but if the metadata size for cache was 256 MiB, it would be 256 MiB smaller which is quite a big difference. And even with the 8 MiB it's just wrong because the ``swap`` LV is the only one that has the exact requested size whereas the other LVs are requested to grow and fill in the remaining space.

2) It brings us closer to the target of the issue #31 where similar fixes/improvements/cleanup should happen for other types of LVs too.

3) It brings us closer to the support of LVM cache metadata size specification by users/user code.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -616,10 +616,10 @@ class LVMLogicalVolumeDevice(DMDevice):
         log.debug("trying to set lv %s size to %s", self.name, size)
         # Don't refuse to set size if we think there's not enough space in the
         # VG for an existing LV, since it's existence proves there is enough
-        # space for it.
+        # space for it. A similar reasoning applies to shrinking the LV.
         if not self.exists and \
            not isinstance(self, LVMThinLogicalVolumeDevice) and \
-           size > self.vg.freeSpace + self.vgSpaceUsed:
+           size > self.size and size > self.vg.freeSpace + self.vgSpaceUsed:
             log.error("failed to set size: %s short", size - (self.vg.freeSpace + self.vgSpaceUsed))
             raise ValueError("not enough free space in volume group")
 
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -347,6 +347,9 @@ class LVMVolumeGroupDevice(ContainerDevice):
         elif self.reserved_space > Size(0):
             reserved = self.reserved_space
 
+        # reserve space for the pmspare LV LVM creates behind our back
+        reserved += self.pmSpareSize
+
         return self.align(reserved, roundup=True)
 
     @property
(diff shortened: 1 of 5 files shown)


jkonecny12:
Looks good to me.
I think this should go to f23-branch too.

vpodzime:
Hmm, python2's ``max()`` doesn't have the same kwargs as python3's ``max()``.

vpodzime:
Should be fixed now.

vpodzime:
Glad to see the recently added tests catching changes. They needs to be updated.

vpodzime:
Should be fixed now. I also checked that the anaconda's kickstart test with a fixed-size LV and two growable cached LVs respects the requested size of the fixed-size LV and leaves no extra space in the VG.

vpodzime:
Please note that the test failure is caused by a workflow issue in Jenkins this time. All tests passed.

blivet/devices/lvm.py
```diff
@@ -614,10 +633,10 @@ def _setSize(self, size):
         log.debug("trying to set lv %s size to %s", self.name, size)
         # Don't refuse to set size if we think there's not enough space in the
         # VG for an existing LV, since it's existence proves there is enough
-        # space for it.
+        # space for it. A similar reasoning applies to shrinking the LV.
         if not self.exists and \
            not isinstance(self, LVMThinLogicalVolumeDevice) and \
-           size > self.vg.freeSpace + self.vgSpaceUsed:
+           size > self.size and size > self.vg.freeSpace + self.vgSpaceUsed:
```
dwlehman: This is for the case when you are reducing the size of a planned/non-existent lv at a point where the vg's model is messed up to the point of thinking it has too little free space even for a reduced version of an lv is already contains?

blivet/devices/lvm.py
```diff
@@ -1653,21 +1667,37 @@ def dependsOn(self, dep):
...
         :param fast_pvs: PVs to allocate the cache on/from (ignored for existing)
         :type fast_pvs: list of :class:`~.devices.storage.StorageDevice`
         :param str mode: desired mode for non-existing cache (ignored for existing)
 
+        .. note::
+            If :param:`md_size` is None for a an unexisting cache, the default
+            is used and it is subtracted from the requested :param:`size` so
+            that the whole cache (data+metadata) fits in the space of size
+            :param:`size`.
+
```
dwlehman: This seems a bit ugly, but it's similar to what happens for new md arrays.

blivet/devices/lvm.py
vpodzime: The problem is that for setting LV's size to a smaller value, it doesn't make sense to compare the requested ``size`` with ``self.vgSpaceUsed + self.vg.freeSpace`` because ``self.vgSpaceUsed`` uses the original size. And it is getting triggered when we try to set LV's size to a smaller value if we find out the original size wouldn't fit to an existing VG (when creating the LV).

blivet/devices/lvm.py
vpodzime: I agree, but actually I think we should do the same for thin pools. It brings the user code one big benefit -- they don't have to care about any meta data space if they don't want to.

PR #208, opened 2015-08-18
Do not propagate low-level blockdev.CryptoError when setting up LUKS (#1253925)

vpodzime:
Otherwise callers like anaconda would have to catch this low-level error
(requiring libblockdev as a dependency and other issues) to prevent tracebacks
when trying to unlock LUKS (possibly with a wrong passphrase).

### DIFF
--- a/blivet/formats/luks.py
+++ b/blivet/formats/luks.py
@@ -187,9 +187,12 @@ class LUKS(DeviceFormat):
     def _setup(self, **kwargs):
         log_method_call(self, device=self.device, mapName=self.mapName,
                         type=self.type, status=self.status)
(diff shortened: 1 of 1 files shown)


jkonecny12:
I think this should go to the f23-branch too.
Otherwise looks good to me.

vpodzime:
> I think this should go to the f23-branch too.

good point

bcl:
Are there other blockdev.crypto.* calls that should be wrapped as well?

vpodzime:
> Are there other blockdev.crypto.* calls that should be wrapped as well?

The others look to me like they are all serious issues not just "bad passphrase". I'm not sure if we should wrap those as it would hide/discard some information.

PR #210, opened 2015-08-20
Fix currentSize for extended partitions (#1254899)

vojtechtrefny:
It isn't possible to use size from /sys for extended partitions
because it always reports it to be 2 sectors.


vpodzime:
Looks good to me.

vojtechtrefny:
Pushed.

PR #211, opened 2015-08-20
Don't store UUIDs or labels of multipath members.

dwlehman:
multipath itself has absolutely no identifying metadata, so whatever
metadata found on the member is not related to the multipath set.
Therefore, we shouldn't be storing it in the member's format.

Resolves: rhbz#1254232


vpodzime:
The change looks good to me, but I feel like the commit message says the exact opposite of what the comment says -- "*so whatever metadata found on the member is not related to the multipath set.*" vs. "*the UUID it sees on a multipath member is for the multipath (and not the member)*".

dwlehman:
That is indeed very confusing. I'll fix it up.

PR #212, opened 2015-08-20
Continue with recursive teardown beyond inactive devices.

dwlehman:
We want to continue to recursively deactivate the parents even if
the current device is inactive, but we don't want to run the
_postTeardown method of a device we're not deactivating.

On rhel7-branch, this will be preceded by a revert of a909ea38.

Related: rhbz#1182229

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -424,6 +424,8 @@ class StorageDevice(Device):
         log_method_call(self, self.name, status=self.status,
                         controllable=self.controllable)
         if not self._preTeardown(recursive=recursive):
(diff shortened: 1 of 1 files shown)


vpodzime:
One thing I don't understand about those functions is why ``_preTeardown`` and ``_teardown`` take the ``recursive`` parameter, but that's probably a target for some future cleanup (if any). This change looks good to me, thanks for looking at it.

dwlehman:
I don't remember if there is, or ever has been, a need to pass that through to `_preTeardown` and `_teardown`.

PR #213, opened 2015-08-21
Don't teardown protected devices

jkonecny12:
We don't want to unmount protected devices.

It tries to umount ``/run/install/isodir`` even when ISO was loaded.

This will go to master too.


jkonecny12:
update: rawhide crashing with inst.repo=hd with ISO without this patch.

vpodzime:
Looks good to me.

jkonecny12:
Pushed

PR #214, opened 2015-08-25
Exclude isodir from valid disks

jkonecny12:
When dracut downloads stage2 from the repository on the iso *(inst.repo=hd:...)* this iso file will mount to ``isodir`` so if ``isodir`` was mounted we can't use the source device in a partitioning.

I don't know if this is the correct way to fix it so I would be glad if @vpodzime or @dwlehman could look on it and give me some feedback.

~~*Will go to master too.*~~

### DIFF
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -105,6 +105,7 @@ class Populator(object):
         # protected device specs as provided by the user
         self.protectedDevSpecs = getattr(conf, "protectedDevSpecs", [])
         self.liveBackingDevice = None
+        self.mountedIsoBackingDevice = None
 
         # names of protected devices at the time of tree population
         self.protectedDevNames = []
@@ -743,7 +744,8 @@ class Populator(object):
             device.protected = True
             # if this is the live backing device we want to mark its parents
             # as protected also
-            if device.name == self.liveBackingDevice:
+            if (device.name == self.liveBackingDevice or
+                device.name == self.mountedIsoBackingDevice):
                 for parent in device.parents:
                     parent.protected = True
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Honestly, I'd rather see the whole block you are extending to go away from blivet. I think we could just extend it for f23-branch, but on master, we should come up with some cleaner solution (in particular not have anaconda's weird paths in blivet's sources).

jkonecny12:
I don't like it too that's why I asked if this is correct way to do this.
I think cleaner solution for this will be to give it from anaconda to blivet somehow but I'm not enough familiar with blivet now to know how to do it.

bcl:
I agree, what I'd try to do is make some way for anaconda to pass in a list of paths for blivet to ignore and make the blivet code more general, just checking a list.

jkonecny12:
Is possible solution when the protected is set for device then it automatically set it to the parent disk? 

Or better case create there counter and add +1 for every protected child device and -1 for every protected = false call. And return ``true`` ``if >=0`` for disk device.

bcl:
I don't think there's much wrong with the current code, other than hard-coding paths. It already sets the parent's protected attr.

jkonecny12:
Yeah but when you want to remove this hard-coded paths (and we don't want them, because this should be in anaconda not blivet) we need to find some other way.
I already consulted this with @vpodzime and I think we have one. 

Question is, do we want this or the cleaner version to f23-branch in other words could I push this to f23 and close it @vpodzime @bcl ?

vpodzime:
> Question is, do we want this or the cleaner version to f23-branch in other words could I push this to f23 and close it @vpodzime @bcl ?

Works for me.

bcl:
I'm fine with it for F23.

jkonecny12:
Pushed. Thank you for help.

PR #215, opened 2015-08-26
Do not reserve extra space for metadata in a VG with RAID PVs

vpodzime:
Reserving 5 extents per disk per LV for metadata if a VG is on top of RAID PVs
is a weird magic that saved us troubles with placing LVM on top of RAID, but
even on the first it's just wrong and doesn't make sense. LVM doesn't allocate
any extra meta data if an LV is added to a VG (unless it is a meta data LV of
course) and the RAID device below it (the PV) is just a block device with some
size and no matter how many disks are below, the meta data space used by LVM is
just subtracted from the RAID device's (PV's) size. Moreover the LVM extent size
has nothing to do with the RAID device so it's like mixing apples and oranges.

If things don't fit we need to tweak our calculations of free space/size in/of
an MD RAID device (depending on the level and underlying disks) since that's the
only place where these tweaks really make sense. However, after quite a lot of
testing it seems that we are reserving more than enough space for the meta data.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -1258,11 +1258,6 @@ class LVMFactory(DeviceFactory):
                 size -= blockdev.lvm.get_lv_physical_size(self.device.size, lvm.LVM_PE_SIZE)
                 log.debug("size cut to %s to omit old device space", size)
 
-        if self.container_raid_level:
-            # add five extents per disk to account for md metadata
-            # (it was originally one per disk but that wasn't enough for raid5)
-            size += lvm.LVM_PE_SIZE * len(self.disks) * 5
-
         if self.container_encrypted:
             # Add space for LUKS metadata, each parent will be encrypted
             size += lvm.LVM_PE_SIZE * len(self.disks)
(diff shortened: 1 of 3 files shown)


bcl:
I'm reluctant to remove this given how much trouble we've had with LVM and not enough PE's in the past. It is possible that improvements in other areas have removed the need to fudge these numbers, which would be great, but I'd like to be sure. Some of the problems we used to hit were related to specific sizes, and in at least one case I've seen LVM not use a PE for no obvious reason.

vpodzime:
One thing I forgot to mention is that in the meantime (between this magic was introduced and now) we have added one nice thing to our code creating LVs -- every time an LV is about to be created, the free space in a VG is checked and the size of the LV is reduces in case there's less space in the VG than what is the calculated LV's size with a warning being logged. So even in cases where this prevented some issues, people should end up with better utilization of their storage, because an LV we would e.g. try to grow too big would be reduced and successfully created.

But generally, the lines this PR removes are clearly wrong. They just don't make sense and in cases where people use e.g. 32 MiB extent size (which was the default for kickstart installations few months ago, btw), those lines result in GiBs (5 * 5disks * 5LVs = 125 * 32 MiB) of unused space. Plus I'm not proposing this for the rhel7-branch so I don't think it could cause any serious issues.

Last but not least, I'm working on kickstart tests for LVM on RAID right now.

bcl:
I'd bet that the VG space check is what has made these kludges irrelevant, and given how much testing you've done on this I'm ok with it.

PR #216, opened 2015-08-27
Catch problems with chassis vendor names (#1256072)

bcl:
If it can't be decoded it certainly isn't a Mac.


jkonecny12:
Looks good to me

vpodzime:
> If it can't be decoded it certainly isn't a Mac.

I'm not too convinced we could be that sure. ;) But yeah, this is a correct change. ACK.

bcl:
pushed.

PR #217, opened 2015-08-27
Master format free space

jkonecny12:
We are using magic numbers acquired by testing to reduce the size depending on a used format.

Also add support for Size operators (add,mul...) by float.

*Related: rhbz#1224048*
*Main patch will go to the anaconda later.*

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -72,6 +72,11 @@ class FS(DeviceFormat):
     _sizeinfoClass = fssize.UnimplementedFSSize
     _syncClass = fssync.UnimplementedFSSync
     _writelabelClass = fswritelabel.UnimplementedFSWriteLabel
+    # This constant is aquired by testing some filesystems
+    # and it's giving us percentage of space left after the format.
+    # This number is more guess then precise number because this
+    # value is already unpredictable and can change in the future...
+    _MetadataSizeFactor = 1.0
 
     def __init__(self, **kwargs):
         """
@@ -167,6 +172,18 @@ class FS(DeviceFormat):
                   "label": self.label, "targetSize": self.targetSize,
                   "mountable": self.mountable})
         return d
+
+    @classmethod
+    def freeSpaceEstimate(cls, device_size):
+        """ Get estimated free space when format will be done on device
+            with size ``device_size``.
+            This is more guess then precise number.
+
+            :param device_size: original device size
+            :type device_size: ``Size`` object
+            :return: estimated free size after format
+        """
+        return device_size * cls._MetadataSizeFactor
 
     def labeling(self):
         """Returns True if this filesystem uses labels, otherwise False.
@@ -336,7 +353,6 @@ class FS(DeviceFormat):
             filesystem to its minimum size.
         """
         return max(Size(0), self.currentSize - self.minSize)
-
 
     def _preCreate(self, **kwargs):
         super(FS, self)._preCreate(**kwargs)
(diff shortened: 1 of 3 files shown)


blivet/formats/fs.py
```diff
@@ -72,6 +72,11 @@ class FS(DeviceFormat):
     _sizeinfoClass = fssize.UnimplementedFSSize
     _syncClass = fssync.UnimplementedFSSync
     _writelabelClass = fswritelabel.UnimplementedFSWriteLabel
+    # This constant is aquired by testing some filesystems
+    # and it's giving us percentage of space left after the format.
+    # This number is more guess then precise number because this
+    # value is already unpredictable and can change in the future...
+    _MagicMetadataMultiplierClass = 1.0
```
bcl: Shouldn't be named 'Class' if it is a constant.

blivet/formats/fs.py
```diff
@@ -168,6 +173,18 @@ def dict(self):
                   "mountable": self.mountable})
         return d
 
+    @classmethod
```
bcl: Why is this a class method? I'd just make it a @property and use self.size instead of passing it in.

vpodzime: I'm gonna reply to this since it was my idea to make it a class method. The reason is that in many cases we want to know how much space a file system would provide even if it doesn't exist (as an object). Think about things like a list of file systems shown to user together with estimations of how much space they would get. And also in anaconda, we need this estimations early e.g. in disk selection where there are no devices and no file systems, but just a number giving the disk free space size.

blivet/formats/fs.py
vpodzime: Yeah and I think ``MetadataSizeFactor`` or something like that would be a better name.

jkonecny12:
Yeah sorry my bad.

blivet/formats/fs.py
jkonecny12: OK I will use ``MetadataSizeFactor``.

blivet/formats/fs.py
bcl: Thanks, that makes sense.

jkonecny12:
I have updated the PR. I hope I did it right :) .

vpodzime:
Looks good to me now.

jkonecny12:
@vpodzime ahhh thank you I'll change it locally.

jkonecny12:
Pushed

dwlehman:
@jkonecny12 Can you please do a followup that removes the float stuff from `blivet.size.Size` and instead uses `decimal.Decimal` for the factors in `blivet.formats.fs`?

jkonecny12:
Hi @dwlehman, yeah I can do that but why we need that? Isn't better to convert it on one place for every use?

jkonecny12:
This PR is pushed already but I'm reopening PR to complete discussion on possible follow-up.

vpodzime:
> @jkonecny12 Can you please do a followup that removes the float stuff from blivet.size.Size and instead uses decimal.Decimal for the factors in blivet.formats.fs?

What would be the benefit of this? I know ``Decimal`` intentionally doesn't implement those mathematical operations with ``float``s because they result in a lost of precision, but is ``Size`` really the same case? We convert everything to whole bytes anyway so what's the point in enforcing those "infinite precision" rules? Could be I'm missing something here, so let me know. It would be a great help for me for thinking about the *GLib* implementation of ``Size``.

dwlehman:
It is about loss of precision from the use of float. We stopped using float for anything related to size some time ago. I forgot that we do round to whole bytes, which means this may be more about convention/principle than practical need.

vpodzime:
> I forgot that we do round to whole bytes, which means this may be more about convention/principle than practical need.

Good to know, I wasn't sure I wasn't missing something here. Let's discuss the approach to this at the meeting.

jkonecny12:
Output from the meeting is this follow-up https://github.com/rhinstaller/blivet/pull/229 .
Closing this issue again.

PR #218, opened 2015-09-01
Improvements for LVM on RAID calculations

vpodzime:
The first commit just removes some old, naïve comments from the code. The second and third commits make our calculations related to LVM on RAID much more precise and realistic -- from ~150 MiB free space in a VG where some LV was requested to grow to a maximum size to ~4-8 MiB (1 or 2 extents). But what's more important, the calculations now reflect the real calculations that happen in the "storage land below us" and they make sense. :)

We still have the "emergency brake" making an LV smaller if it's about to be created in a VG that doesn't have enough space for it so this should be safe. Plus I'm only proposing this for *master* right now and I'll make anaconda's kickstart tests stricter so that they reveal any potential issues this could cause and any future changes breaking this.

### DIFF
--- a/blivet/devicelibs/raid.py
+++ b/blivet/devicelibs/raid.py
@@ -281,8 +281,7 @@ class RAIDn(RAIDLevel):
             raise RaidError("superblock_size_func value of None is not acceptable")
 
         min_size = min(member_sizes)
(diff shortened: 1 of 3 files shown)


vpodzime:
I believe this is the last missing piece to call the issue #31 resolved.

PR #221, opened 2015-09-09
Mount efivarfs during os installation (#1261559)

bcl:
Tools running inside the chroot need access to /sys/firmware/efi/efivars
if it exists on the host. eg. efibootmgr

Resolves: rhbz#1261559


bcl:
Same as pull 219, the changes are just in a different file.

vpodzime:
Looks good to me too.

bcl:
Pushed

PR #222, opened 2015-09-10
Fix extended resize

vojtechtrefny:
Fix resizing of extended partitions:
1) Report minSize as size of extended partition - size of free space at the end of it. (Blivet now reports 0 B.) Fixes #1254875
2) Allow resizing of non-leaf partitions.

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -640,7 +640,7 @@ class PartitionDevice(StorageDevice):
 
     def resize(self):
         log_method_call(self, self.name, status=self.status)
-        self._preDestroy()
+        self._preResize()
 
         # partedDisk has been restored to _origPartedDisk, so
         # recalculate resize geometry because we may have new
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me since the newly added method looks like it's ``StorageDevice._preDestroy()`` merged with ``PartitionDevice._preDestroy()`` without the ``isLeaf()`` check. Size calculations look good to me.

vojtechtrefny:
New version based on @dwlehman comments:
- aligned minSize -- fixed
- minSize for empty extended partition -- 1 KiB (2 sectors) is what system reports as size for extended partition; not sure about the grainSize -- is it possible to be smaller than that?
- resizing of non-leaf partitions -- fixed to work only on extended partitions

vpodzime:
> resizing of non-leaf partitions -- fixed to work only on extended partitions

Is it even possible there would be a non-leaf partition that is not extended?

This still looks good to me.

vojtechtrefny:
resizing of non-leaf partitions -- fixed to work only on extended partitions

    Is it even possible there would be a non-leaf partition that is not extended?

Yes, every partition with children -- PVs, mdmembers, encrypted partitions...

dwlehman:
This looks good to me, except that it really needs unit tests. This is a great example of the kind of code that should have unit tests covering it to ensure people don't break it in the future without realizing it.

vojtechtrefny:
Test added.

vpodzime:
Nice! The test looks good to me.

dwlehman:
Thanks for adding the test -- it looks good.

vojtechtrefny:
Pushed.

PR #223, opened 2015-09-14
Don't teardown FSs when searching for installed systems (#1252902)

jkonecny12:
When looking for existing installations the FSs are mounted to the sysroot but if the device is protected it can't be teared down.

Umount the disks instead of teardown. The `teardownAll()` method is used when all devices were tested.

*Related: rhbz#1252902*

*Will go to master too*

### DIFF
--- a/blivet/osinstall.py
+++ b/blivet/osinstall.py
@@ -239,10 +239,11 @@ def _findExistingInstallations(devicetree):
             device.format.mount(options=options, mountpoint=getSysroot())
         except Exception: # pylint: disable=broad-except
             log_exception_info(log.warning, "mount of %s as %s failed", [device.name, device.format.type])
-            device.teardown()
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me, but I'd like to know @dwlehman's opinion here. The other solution for the issue @jkonecny12 was proposing was to skip protected devices when finding existing installations, but I think this way we would find more of existing installations (if they have parts on protected devices) and umounting devices instead of complete teardown could spare us some teardown(), setup(), teardown(), setup(),... cycles. The teardownAll() run at the end should make sure all devices are torn down.

dwlehman:
I like this approach because it potentially eliminates some unnecessary teardown/setup cycles. It's also a plus that it allows us to find existing installations on protected devices.

jkonecny12:
Pushed. Thank you for comments

PR #226, opened 2015-09-16
lvactivate when lvmetad is active

dwlehman:
When lvmetad is running and the last PV in a VG is activated, an asynchronous system unit (`lvm2-pvscan@...service`) is scheduled to update LVM's view of things and to auto-activate the LVs. During the interim between that unit being scheduled and run, LVM behaves as if the PV were still inactive. So, when we're deactivating/activating the PVs we should wait for auto-activation instead of plowing ahead and trying to activate the LVs ourselves. This has caused several bugs in which we activate an MD PV, then try to activate an LV, and get an error from LVM saying the VG doesn't exist.

### DIFF
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -20,6 +20,7 @@
 # Author(s): Dave Lehman <dlehman@redhat.com>
 #
 
+import os
 import re
 
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me. A little bit unfortunate, but we can hardly do any better here.

dwlehman:
Bah, I need to update the comment to change the first reference to the timeout from two seconds to 30 seconds.

vpodzime:
I didn't even notice that so still looks good to me.

PR #227, opened 2015-09-17
warn on failure to scan partitions

larsks:
I had a Fedora install fail due to https://bugzilla.redhat.com/1263832,
and I wonder if we should be treating failures at this point in the code
as warnings rather than errors.  This would permit blivet to continue
scanning other devices.

This would seem to be in line with the comments at the beginning of this
section of code, which suggest that the desired behavior is to ignore
the device rather than throwing an exception:

    # Ignore partitions on:
    #  - devices we do not support partitioning of, like logical volumes
    #  - devices that do not have a usable disklabel
    #  - devices that contain disklabels made by isohybrid

### DIFF
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -441,12 +441,11 @@ class Populator(object):
                not self._isIgnoredDisk(disk):
                 if info.get("ID_PART_TABLE_TYPE") == "gpt":
                     msg = "corrupt gpt disklabel on disk %s" % disk.name
-                    cls = CorruptGPTError
                 else:
                     msg = "failed to scan disk %s" % disk.name
-                    cls = DiskLabelScanError
 
-                raise cls(msg)
+                log.error(msg)
+                return
 
             # there's no need to filter partitions on members of multipaths or
             # fwraid members from lvm since multipath and dmraid are already


dwlehman:
The problem with just ignoring these devices is that other devices are often built on top of them (think lvm). What, then, do you do with those lvm devices? How can you manage them when you are actively denying that they exist?

larsks:
I don't know.  I'm not suggesting that this is the best or even an appropriate solution, but I think the current behavior is problematic, especially in how it percolates up to the user experience in Anaconda.

Maybe it's really not a blivet problem, and the fix is that anaconda should respond to these errors by saying, "there was a problem scanning device <device>, do you want to ignore the device, retry the scan, or fail?" or something (although in this case blivet would probably need to provide better exceptions that make it easy to pull out the failed device name).

Maybe blivet needs a an option to switch between device-failures-are-exceptions and device-failures-are-logged, and in the latter case ignore both the device and anything built on top of it.

dwlehman:
Ignoring the devices on top of partitions has real consequences. Namely, it prevents changes to the disk that contains them. So you would have to ignore the whole disk and any devices related to it by way of things like lvm vgs or md arrays that span multiple disks.

The current approach for this class of problems is to raise an exception because these are not problems that can be simply ignored. This is the point at which the user must take action to restore sanity to their system before any software can reasonably be expected to do a good job of reconfiguring it.

larsks:
So this argues for the first option I suggested in my previous comment (a better UI in anaconda to handle this sort of error).  I'll go ahead and close this, then.

vpodzime:
> So this argues for the first option I suggested in my previous comment (a better UI in anaconda to handle this sort of error). I'll go ahead and close this, then.

Anaconda should react to exceptions inherited from the ``UnusableConfigurationError`` class (which both ``CorruptGPTError`` and ``DiskLabelScanError`` are) by showing a dialog informing user that the system has an configuration that cannot be used for installation. So this is already covered. However, from the bug reports it seems like the resulting exception in the end is different.?

larsks:
Yes, Anaconda presents the generic "an unknown error has occurred" dialog, as seen [here](https://bugzilla.redhat.com/attachment.cgi?id=1074497).  Because this doesn't communicate the error message contained in the exception to the user it doesn't really afford the chance to correct the error.

PR #229, opened 2015-09-18
Convert float to str for better precision in Size

jkonecny12:
Convert float to string before it's converted to Decimal. This prevents precision loss.

*Related: rhbz#1224048*

----------------------------------
This was discussed on storage meeting.


vojtechtrefny:
Looks good to me.

vpodzime:
Looks good to me too.

jkonecny12:
Pushed

PR #230, opened 2015-09-21
Add a 'debug' flag and use it for LVM debugging

vpodzime:
Many times we have been asked by the LVM team to use the '-vvvv' option whenever
we do any LVM action so that the logs contain more (debugging)
information. However, getting the output of every single LVM command cluttered
with a lot of debugging info would be confusing for both us reading the logs and
the code parsing the output of such commands. The trade-off solution is to pass
the 'log{level=7}' config with all our commands that causes LVM to put the extra
info from '-vvvv' to syslog where it looks like all the other crystal-clear
stuff.

### DIFF
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -90,6 +90,8 @@ def _set_global_config():
     config_string = " devices { %s } " % (devices_string) # strings can be added
     if not flags.lvm_metadata_backup:
         config_string += "backup {backup=0 archive=0} "
+    if flags.debug:
+        config_string += "log {level=7 file=/tmp/lvm.log}"
 
     blockdev.lvm.set_global_config(config_string)
 
--- a/blivet/flags.py
+++ b/blivet/flags.py
@@ -31,6 +31,7 @@ class Flags(object):
         #
         self.testing = False
         self.installer_mode = False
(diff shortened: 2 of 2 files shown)


bcl:
Sounds like a good idea, but how verbose is it? Do you have some example output?

vpodzime:
It's very verbose, the example of what ends up in journal when running ``lvs --config='log{level=7}'`` on my system can be seen here: http://fpaste.org/269992/44290504/

However, it contains all the information the LVM team always wants from us.

bcl:
yikes, that's quite a bit of extra logging. How about only logging when anaconda's loglevel is at debug? On a simple 1 disk LVM installation I count 36 lvm calls in program.log, on systems with more complex setups this is going to be way more information than we normally need.

vpodzime:
I'm not sure how much we care about the size of the log, but it is the usual logging problem -- we often need the logs from a particular installation that didn't run in any debugging mode. Maybe we could add a prefix to those messages (which is possible) so that they can be easily filtered out?

bcl:
That's true, to a point, but we also have the problem of logging with lots of drives slowing things down, don't we? I like lots of logs by default, but this seems like too much to me.

bcl:
Looks good, thanks!

PR #231, opened 2015-09-22
Handle sysfs size if it is missing (#1265090)

bcl:
sysfs paths may, for whatever reason, not exist. Return '0' for int()
instead of None if this is the case.

Resolves: rhbz#1265090


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #232, opened 2015-09-22
Use the RAID classes to calculate btrfs sizes (#1264696)

dashea:
We already have a bunch of a RAID size calculcation code, so use that
for the raid modes of btrfs volumes. Default to a raid level of "single"
instead of None for consistency.


dashea:
The test results show, besides that I didn't run the tests, that I probably messed up something with the metadata type. Pushed a new patch that leaves the default dataLevel as None and does the switcheroo to raid.Single during _getSize instead.

vpodzime:
Looks good to me.

vpodzime:
Maybe we could do better with the default ``dataLevel`` for btrfs devices, but this definitely improves things.

dashea:
Pushed.

PR #234, opened 2015-09-24
Fix calling non-existing method (#1252902)

jkonecny12:
Fix my bug from commit (#05feef37e2c5ae8fd63abe2198ccd8d55e09b08a) where
I was calling non-existing device.format.umount() method, the unmount()
(missing 'n') method exists there but it's still only calling teardown()
instead of system umount. I'm changing call to util.umount() which
calling the system umount.
Because of this regression the first call of the
_findExistingInstallations method will raise an exception so no existing
installations were found and everything were in the Unknown group.
The exception was logged to storage log as info so I missed it.

Sorry guys for this issue.

*Related: rhbz#1252902*

*Will go to master branch too*

### DIFF
--- a/blivet/osinstall.py
+++ b/blivet/osinstall.py
@@ -239,11 +239,11 @@ def _findExistingInstallations(devicetree):
             device.format.mount(options=options, mountpoint=getSysroot())
         except Exception: # pylint: disable=broad-except
             log_exception_info(log.warning, "mount of %s as %s failed", [device.name, device.format.type])
-            device.format.umount(mountpoint=getSysroot())
+            util.umount(mountpoint=getSysroot())
             continue
 
         if not os.access(getSysroot() + "/etc/fstab", os.R_OK):
-            device.format.umount(mountpoint=getSysroot())
+            util.umount(mountpoint=getSysroot())
             device.teardown(recursive=True)
             continue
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

blivet/osinstall.py
```diff
@@ -239,11 +239,11 @@ def _findExistingInstallations(devicetree):
             device.format.mount(options=options, mountpoint=getSysroot())
         except Exception: # pylint: disable=broad-except
             log_exception_info(log.warning, "mount of %s as %s failed", [device.name, device.format.type])
-            device.format.umount(mountpoint=getSysroot())
```
bcl: I think device.format.unmount is the right thing to call here, isn't it? Looking at blivet/formats/fs.py the unmount method does several things, including udev settling, calling util.umount and running lsof if it can't be unmounted.

jkonecny12: Yeah it's little complicated. When you read PR https://github.com/rhinstaller/blivet/pull/223 which I fixed badly with the commit. You can't use `teardown()` because of it won't umount protected disks here. So when you have stage2 or repo HDD connected it's mounted forever and it's blocking repository to load properly. 
So or so after `_findExistingInstallations()` method finish `teardownAll()` should be called. Some potential problem could be the `teardown_all` parameter in the `findExistingInstallations()` method could be set false but I didn't found out calling where this happening so this could be patched for master as follow-up but to f23 this is a good patch though.

bcl: Ah, ok. Thanks for the explanation.

jkonecny12:
Pushed

PR #235, opened 2015-09-24
Add umount next to unmount method to avoid miscall

jkonecny12:
System using umount so user could expect we will use the same naming conventions.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -756,7 +756,7 @@ class FS(DeviceFormat):
     def unmount(self, **kwargs):
         """ Unmount this filesystem.
 
-            :param str mountpoint: Optional mountpoint to be unmounted.
+            :keyword str mountpoint: Optional mountpoint to be unmounted.
             :raises: FSError
 
             If mountpoint isn't passed this will unmount the most recent mountpoint listed
@@ -768,7 +768,7 @@ class FS(DeviceFormat):
     def umount(self, **kwargs):
         """ Unmount this filesystem.
 
-            :param str mountpoint: Optional mountpoint to be unmounted.
+            :keyword str mountpoint: Optional mountpoint to be unmounted.
             :raises: FSError
 
             If mountpoint isn't passed this will unmount the most recent mountpoint listed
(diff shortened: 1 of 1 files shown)


blivet/formats/fs.py
```diff
@@ -765,6 +765,18 @@ def unmount(self, **kwargs):
         """
         return self.teardown(**kwargs)
 
+    def umount(self, **kwargs):
+        """ Unmount this filesystem.
+
+            :param str mountpoint: Optional mountpoint to be unmounted.
```
vpodzime: While doing this, could you please replace ``param`` with ``keyword`` here as well as in the ``unmount()`` method? That's the proper field name, I think.

vpodzime:
Looks good to me otherwise.

jkonecny12:
Docstring param changed to keyword. Thank you @vpodzime.
PR updated.

bcl:
I don't see any reason for this. We have an API, so what if users may get confused? They'll read the docs and learn the right method to call.

jkonecny12:
IMHO many users are used to call `umount()`. Methods `umount()` and `unmount()` are really easy to mixed up also when user see somewhere in the code or in the documentation `mount()` automatic reaction is look for `umount()`.
I think it's better to have alias for existing method then anger user because of unpredictable names.

But the main word here have the storage guys, they are using blivet most so @dwlehman, @vpodzime, @vojtechtrefny what do you think?

vpodzime:
> But the main word here have the storage guys, they are using blivet most so @dwlehman, @vpodzime, @vojtechtrefny what do you think?

Well, since it was my idea to add this alias, I'm still for it. :)

dwlehman:
I often wish I had never added `unmount` and instead force people to use `teardown` like all the non-fs formats (and all of the devices, too) do.

dwlehman:
If we do add `umount` we are setting precedent for adding `activate` (`setup`), `deactivate` (`teardown`), and many other aliases to avoid frustrating people who are not willing to pay attention (or test their code).

jkonecny12:
OK, no problem. 
Closing this PR.

PR #238, opened 2015-10-02
Switch to using rd.iscsi.initiator (#1268315)

bcl:
Deprecated dracut argument.

Resolves: rhbz#1268315


vpodzime:
"The longer, the better!" Looks good to me.

bcl:
Pushed.

PR #239, opened 2015-10-06
Always build python2 and python3 versions.

bcl:
Also need to include the python2-devel package for the macros. Blivet's
Makefile now defaults to python3 so we need to explicitly set the python
version when installing.


vpodzime:
Looks good to me.

PR #240, opened 2015-10-08
Master fix btrfs tempmount

vojtechtrefny:
Mountpoint detection for btrfs is broken again. The problem is with btrfs raids -- btrfs format stores information only about one parent device (the first one we populate from udev devices) and `/proc/self/mountinfo` also contains name of only one device (it should be the first one in the raid). And sometimes these devices are not the same. I don't know why, I wasn't able to reproduce it but apparently udev sometimes doesn't return disks sorted by name. Result is `systemMountpoint` is `None` because we are looking for mountpoint for `sdb` but `/proc/self/mountinfo` contains information only for `sda` (example for btrfs raid over two disks).

I've changed `_do_temp_mount` to `contextmanager` so if we actually need to mount the btrfs we always use the right mountpoint for libblockdev and also always unmount it.

We should of course change the btrfs format to contain names of all parent devices, but this will require an API change (and could break more things in btrfs) so it is not possible for f23-branch.

--
The deprecation of `createSubVolumes` is unrelated to the bug, I just noticed this method doesn't work (`create` doesn't take arguments) and we actually don't use it.

### DIFF
--- a/blivet/devices/btrfs.py
+++ b/blivet/devices/btrfs.py
@@ -35,6 +35,7 @@ from .. import errors
 from ..flags import flags
 from ..storage_log import log_method_call
 from .. import udev
+from .. import util
 from ..formats import getFormat, DeviceFormat
 from ..size import Size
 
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Updated version based on @dwlehman comment: `listSubVolumes` now mounts the device only in installer mode.

dwlehman:
Looks good to me.

vpodzime:
Looks good to me too.

vojtechtrefny:
Pushed.

PR #241, opened 2015-10-08
suppress LVM autobackup during image installs (rhbz#1269144)

wgwoods:
Somehow the patches that fixed [rhbz#1066004] did not make their way into the previous blivet master branch, which means they got lost when this branch got rebased.. or something like that.

These patches should fix [rhbz#1269144] - which is a clone of [rhbz#1066004].

A little massaging was necessary to fix conflicts in the `devicelibs.lvm` refactor patch, but it *looks* correct, at least.

[rhbz#1066004]: https://bugzilla.redhat.com/show_bug.cgi?id=1066004
[rhbz#1269144]: https://bugzilla.redhat.com/show_bug.cgi?id=1269144

### DIFF
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -35,6 +35,7 @@ from .. import util
 from .. import arch
 from ..errors import LVMError
 from ..i18n import _, N_
+from ..flags import flags
 
 MAX_LV_SLOTS = 256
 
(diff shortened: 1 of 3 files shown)


vpodzime:
These all look good to me.

bcl:
Tested with image install and it works great, nothing written to /etc/

PR #242, opened 2015-10-08
Fix an overly inclusive regex in DeviceTree.resolveDevice.

### LINKED ISSUE #152: resolveDevice fails on systems with > 15 disks due to bad regex
In my Anaconda kickstart config, I'm specifying partition information for 16 disks, sda-sdp. When kickstart calls out to blivet/devicetree.py to look up device mappings for each of the drives, it dies and creates a traceback when resolveDevice(sdp) is being processed.

There's a regex on line 2548 that appears to be the root cause, because the device name contains 'p' it matches this regex and causes blivet to do the wrong thing:

https://github.com/rhinstaller/blivet/blob/rhel7.1-branch/blivet/devicetree.py#L2548 :
```python
  elif re.match(r'(0x)?[A-Za-z0-9]{2}(p\d+)?$', devspec):
```
I'm not exactly sure what this regex is intended to match, (I'm assuming something like "disk1p1") so anchoring it at the beginning may/may not be a trivial fix.

A simple example:
```python
>>> import re
>>> if re.match(r'(0x)?[A-Za-z0-9]{2}(p\d+)?$', 'sdp1'):
...   print "matched"
...
matched
>>> if re.match(r'(0x)?[A-Za-z0-9]{2}(p\d+)?$', 'sdq1'):
...   print "matched"
...
>>>
```

A copy of the Anaconda config+traceback error, storage.log illustrating the problem, are in this gist:
https://gist.github.com/bwann/b8286b72613c03b98ee7  This is from Anaconda 19 with CentOS 7.1, and the regex is the same even in trunk.

dwlehman:
This prevented proper resolution of some partition names.

Fixes #152

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -933,12 +933,13 @@ class DeviceTree(object):
                 (label.startswith("'") and label.endswith("'"))):
                 label = label[1:-1]
             device = self.labels.get(label)
-        elif re.match(r'(0x)?[A-Za-z0-9]{2}(p\d+)?$', devspec):
+        elif re.match(r'(0x)?[A-Fa-f0-9]{2}(p\d+)?$', devspec):
             # BIOS drive number
-            spec = int(devspec, 16)
+            (drive, _p, partnum) = devspec.partition("p")
+            spec = int(drive, 16)
             for (edd_name, edd_number) in self.eddDict.items():
                 if edd_number == spec:
-                    device = self.getDeviceByName(edd_name)
+                    device = self.getDeviceByName(edd_name + partnum)
                     break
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me. I just don't quite remember what was our decision about using the #NUMBER in commit messages for GH issues.

bcl:
Needs a test though ;)

dwlehman:
Yeah, writing tests for DeviceTree.resolveDevice is something I keep putting off doing.

dwlehman:
New version has tests!

bcl:
Perfect!

PR #243, opened 2015-10-08
Allow allocation of small, minimally-aligned partitions.

dwlehman:
Without this patch, blivet cannot create any partition smaller than the optimal io size reported by the disk it's being created on (without passing a start sector to `partitioninging.addPartition` to disable alignment). This PR makes it possible.


vpodzime:
Looks good to me. I'd just prefer this to be split into multiple patches if that's possible.

vpodzime:
Looks great to me now. Thanks!

PR #248, opened 2015-10-13
Edd: fix some "make check" problems

vathpela:
Before, "make test" worked but "make check" didn't, and for some reason I didn't notice I broke jenkins earlier.  This should fix it.


jkonecny12:
Could you please change the `dammit` commit? It's too big change for this name.

vpodzime:
> Could you please rename the dammit commit? It's too big change for this name.

Yeah, it definitely deserves some more information. This way it's quite hard to tell if it's doing the right thing.

vathpela:
Sorry for all the extra messages; was working out how to make it so we can actually /debug/ any test failures from jenkins.  Now that I've sorted that out, I'll close this PR until it's really ready.

PR #250, opened 2015-10-14
Add an edd test data harvester.

vathpela:


vpodzime:
I see nothing that would be wrong there.

PR #251, opened 2015-10-16
Use libbytesize's Size class as a base for our Size class

vpodzime:
libbytesize provides an implementation of what we need for manipulations and
calculations with sizes in bytes. It's a little bit faster and more importantly
it can be shared with other projects because it's written in C. Thus we should
use it.

At some point it would be nice if we could drop our own Size class and the
blivet/size.py module altogether because they don't bring any extra value, but
for now we need this thin wrapper to keep the API (almost*) stable. The only
real changes are:

1) Strings like "100 kibibytes" are no longer accepted, I doubt anybody has ever
used this functionality. If the opposite turns out to be true such support will
be added to libbytesize and no change will be needed in Blivet.

2) humanReadable() now ignores the 'skip' and 'min_value' attributes. The former
one is always replaced by 'True' (i.e. trailing zeroes are always stripped), the
latter one is ignored because it made things really cryptic and weird.

* see the commit, mainly the part modifying/removing the tests for more details

Note: tests are expected to be failing at this phase -- libbytesize is not available as a package yet. I am of course not going to merge this before that changes.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -19,161 +19,14 @@
 #
 # Red Hat Author(s): David Cantrell <dcantrell@redhat.com>
 
-import re
-import string           # pylint: disable=deprecated-module
-import locale
-import sys
-from collections import namedtuple
+from bytesize import bytesize
 
-from decimal import Decimal
-from decimal import InvalidOperation
-from decimal import ROUND_DOWN, ROUND_UP, ROUND_HALF_UP
-import six
+# we just need to make these objects available here
+# pylint: disable=unused-import
+from bytesize.bytesize import B, KiB, MiB, GiB, TiB, PiB, EiB, ZiB, YiB, KB, MB, GB, TB, PB, EB, ZB, YB
+from bytesize.bytesize import ROUND_UP, ROUND_DOWN
 
-from .errors import SizePlacesError
-from .i18n import _, N_
-from .util import stringize, unicodeize
-
-ROUND_DEFAULT = ROUND_HALF_UP
-
-# Container for size unit prefix information
-_Prefix = namedtuple("Prefix", ["factor", "prefix", "abbr"])
-
-_DECIMAL_FACTOR = 10 ** 3
-_BINARY_FACTOR = 2 ** 10
-
-# TRANSLATORS: 'B' for "Bytes"
-_BYTES_SYMBOL = N_("B")
-_BYTES_WORDS = (N_("bytes"), N_("byte"))
-
-# Symbolic constants for units
-B = _Prefix(1, "", "")
-
-KB = _Prefix(_DECIMAL_FACTOR ** 1,
-             N_("kilo"),
-             # TRANSLATORS: "k" for "kilo-" (x 10^3)
-             N_("k"))
-
-MB = _Prefix(_DECIMAL_FACTOR ** 2,
-             N_("mega"),
-             # TRANSLATORS: "M" for "mega-" (x 10^6)
-             N_("M"))
-
-GB = _Prefix(_DECIMAL_FACTOR ** 3,
-             N_("giga"),
-             # TRANSLATORS: "G" for "giga-" (x 10^9)
-             N_("G"))
-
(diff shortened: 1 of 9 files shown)


blivet/size.py
```diff
@@ -19,229 +19,20 @@
...
+from gi.repository import ByteSize
+from gi.repository.ByteSize import ROUND_DOWN, ROUND_UP
 
-        :param spec: the specification of a size with, optionally, units
-        :type spec: any type of string like object
-        :returns: numeric value of the specification in bytes
-        :rtype: Decimal
+# we just need to make these objects available here
+# pylint: disable=unused-import
+from gi.repository.ByteSize import B, KiB, MiB, GiB, TiB, PiB, EiB, ZiB, YiB, KB, MB, GB, TB, PB, EB, ZB, YB
```
jkonecny12: Can this be changed to import `gi.repository.ByteSize`, this seems to me odd and error-prone.

jkonecny12:
Except the one it looks good to me but it would be nice if more people will look on this one.

Good job with the libbytesize @vpodzime .

blivet/size.py
vpodzime: Not sure what you are asking about, but the point here is to get those "objects" (integer constants actually) to the namespace of this module because they were/are part of the API and other modules import and use them.

dwlehman:
A note in the commit message about humanReadable -v- human_readable would be nice.

Otherwise, it looks good.

blivet/size.py
jkonecny12: I meant to import it without naming all the constants. To import whole module or move the constants to special module if there are too much stuff which will be loaded too.

vpodzime: That would only be possible with a ``*`` import which I really want to avoid doing.

vpodzime:
> A note in the commit message about humanReadable -v- human_readable would be nice.

Good point, will do.

clumens:
Updated breaker with libbytesize stuff.

vpodzime:
> Updated breaker with libbytesize stuff.

Thanks!

vpodzime:
ok, this should be done. With *libbytesize-0.2-1* tests work on my system, but right now only with python 2 because of a GI bug I'm about to file. So we need to hold this off anyway because of that bug.

vpodzime:
Just for the record, this is the bug in GI:
https://bugzilla.redhat.com/show_bug.cgi?id=1275259

jkonecny12:
Looks good to me. 
You are again using something which wasn't used before you? :D

vpodzime:
> You are again using something which wasn't used before you? :D

To be honest, I haven't expected it to work at all. But since it works with Python 2 just fine I think it should also work with Python 3.

dwlehman:
@vpodzime you don't want to do this for blivet-2.0?

vpodzime:
> @vpodzime you don't want to do this for blivet-2.0?

I do, but due to the [bug in GObject introspection](https://bugzilla.redhat.com/show_bug.cgi?id=1275259), libbytesize doesn't work with Python3 properly. I realized I don't need the GObject stuff at all if I write the bindings myself which should be fairly easy. This should happen this week.

vpodzime:
The next build of libbytesize (0.4) won't require/use GLib/GObject (see https://github.com/rhinstaller/libbytesize/pull/5 for details). These are the adapted patches for Blivet to make use of the new version of libbytesize.

dwlehman:
Check the console output from jenkins.

dwlehman:
Will this introduce any changes in terms of handling for translated size strings? (input or output)

vpodzime:
> Check the console output from jenkins.

Thanks for the heads-up. Somehow I ignored everything below the fact that it couldn't import *bytesize*. Should be fixed now.

vpodzime:
> Will this introduce any changes in terms of handling for translated size strings? (input or output)

There should be no difference as longs as everything works as expected (there will likely be some bugs, hopefully minor, in *libbytesize* in the beginning).

dwlehman:
Looks good to me.

PR #252, opened 2015-10-19
Master pv free

vojtechtrefny:
There are two changes in this PR:
- I've added `pvFree` to lvmpv format. I need this for LVM cache support in blivet-gui.
- Thanks to pvFree we know if the PV is empty, so we can the solve problem with removing empty PVs from VG. We run `pvmove` automatically before `vgreduce` and `pvmove` returns 5 for empty PVs so we think it failed (and 5 is not unique return code for this situation).


vpodzime:
Looks good to me.

jkonecny12:
+1 for me

PR #253, opened 2015-10-19
Require the python libblockdev packages

dashea:
See also https://github.com/rhinstaller/libblockdev/pull/49


vpodzime:
As mentioned in the rhinstaller/libblockdev#49, I'd prefer the packages to be ``python[23]-blockdev`` without the ``lib`` part.

vpodzime:
New libblockdev providing separate ``python[23]-`` packages is released, built and available in Rawhide, so this can be merged.

PR #255, opened 2015-10-23
Make MetadataSizeFactor public variable

jkonecny12:
Required for calculation how much space we need to create device with
usable size X.


vpodzime:
As explained in the https://github.com/rhinstaller/anaconda/pull/424 I don't think we need to do this and thus we should avoid doing it.

jkonecny12:
After discussion with @vpodzime we found better way to do this. Because there is almost nothing similar I'm closing this PR and will create new one later.

PR #256, opened 2015-10-23
edd: Fix one regexp so it matches against a real system /or/ test data.

vathpela:
Woops.


vpodzime:
Looks good to me too.

jkonecny12:
Seems good to me too.

PR #257, opened 2015-10-26
Make the function adding the deprecation doc text more generic

vpodzime:
This way we can use it not only for the deprecation decorator, but for similar
things as well.

(this is a usable thing I implemented on a dead-end way to implement a ``@singleton`` decorator)


jkonecny12:
Looks good to me.

PR #262, opened 2015-10-28
Edd more testcases

vathpela:
This adds (on top of PR #261) a bunch more EDD test cases from various physical and virtual machines I've found.


bcl:
Other than those few comments I think I'm ok with this. Other eyes would be good, I'm afraid I didn't actually read every single line :)

PR #263, opened 2015-10-28
PEP8 compliance

dwlehman:
This brings blivet into compliance with PEP8, except for maximum line length. At some point I'd like to shorten many of the long lines, but I've had enough of this type of work for the time being.

I am open to squashing some of these together.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -31,9 +31,9 @@ __version__ = '1.16'
 ##
 iutil = None
 ROOT_PATH = '/'
-_storageRoot = ROOT_PATH
+_storage_root = ROOT_PATH
 _sysroot = ROOT_PATH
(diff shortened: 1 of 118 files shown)


jkonecny12:
Good luck with 2.x version of blivet.
Looks good to me but this needs more eyes here.
Did you use some script for the renaming?

dwlehman:
I wrote and used a very dumb/simple script to do a (very) rough first pass at converting the names. You can find it here: 

   https://gist.github.com/dwlehman/ab8e159f98d7d9708e4a

I went through and manually cleaned up a lot after running it. The biggest problem with the script is that it does not know where a name has come from, so it even converts names from other modules like parted, logging, &c.

vpodzime:
Looks good to me otherwise.

PR #265, opened 2015-10-29
Add more class methods for better use of metadata

jkonecny12:
Add 3 new class methods:
1) addMetadataToUsableSpace(cls, usable_size) - takes how much space we
need and returns this space + metadata of this FS.
2) hungriestFS(cls, fs_list) - Return FS from the list which takes most
of the space by metadata. Input list is useful to filter only the
required FSs.
3) hungriestFSFromAll(cls) - Same as hungriestFS but with all
filesystems in the blivet.


jkonecny12:
If you have better names I'm all ears. I don't like them but can't figure anything else.

blivet/formats/fs.py
```diff
@@ -179,12 +179,64 @@ def freeSpaceEstimate(cls, device_size):
...
+            :type usable_size: :class:`~.size.Size` object
+            :return: estimated size of the device which will have given amount of
+                ``usable_size``
+            :rtype: :class:`~.size.Size`
+        """
+        # to get usable size without metadata we will use
+        # usable_size = device_size * _MetadataSizeFactor
+        # we can change this to get device size with required usable_size
+        # device_size = usable_size / _MetadataSizeFactor
+        return Size(Decimal(usable_size) / Decimal(cls._MetadataSizeFactor))
```
dwlehman: I'm fairly certain this should be more like (easily verifiable with usable_size of 10 GiB and factor 0.80):

```python
return usable_size + Size(Decimal(usable_size) * Decimal(1 - cls._MetadataSizeFactor))
```
or, slightly easier to read:

```python
metadata_size = Size(Decimal(usable_size) * Decimal(1 - cls._MetadataSizeFactor))
return usable_size + metadata_size
```
This makes me wonder if the method should just return the metadata size and allow the caller to do with that what it wants.

dwlehman:
I would suggest that you change the first method to `getMetadataSize(cls, total_size)` or similar and remove the other two methods unless you have plans for them.

blivet/formats/fs.py
jkonecny12: Yeah you doing it the same way how I did it in start and why I can't finish my computation right. For true this solution was find by vtrefny not me :) .

The problem is this:
``10 GiB * (1 - 0.8) = 2 GiB``
when you try to get usable size:
``(10 GiB + 2 GiB) * 0.8 = 9.6 GiB``

So you don't get same value as:
``10 GiB / 0.8 = 12.5 GiB``
which is equal to usable size we want
``12.5 GiB * 0.8 = 10 GiB``

And for the returning only metadata, I see no problem with it but it seems to me the same. You only do ``result - usable_size`` and this can be done by a caller of this method if he/she wants to. This seems more logical to me but as I said no problem to change it.

jkonecny12:
The two method there I'm using them right now. When you looked to the anaconda PR you will see I need to get info which of the FSs is with greatest metadata size [1] .
When we show user the resize dialog we don't have any information which format will be used for the root so I need to get a biggest possible value there.
I can remove the ``hungriestFS`` with the list argument but I think it could be used in the future.

1 https://github.com/rhinstaller/anaconda/pull/424/files#diff-67c821c2c42c4c96a7ad07c5a772061eR88

blivet/formats/fs.py
```diff
@@ -179,12 +179,64 @@ def freeSpaceEstimate(cls, device_size):
             with size ``device_size``.
             This is more guess then precise number.
 
+            .. note::
+
+                This is more guess then precise number. To get precise
```
vpodzime: *than*

blivet/formats/fs.py
vpodzime: I see everybody has the same first wrong idea about the computation. :) I must say I don't see too much sense in returning the metadata size as it's not really useful for anything else than adding it to the requested size and getting the required size.

vpodzime:
What about something like ``getRequiredSize(free_space)``? And what about ``biggestOverheadFS(fs_list=None)`` where all file systems are taken into account if ``fs_list`` is ``None``?

blivet/formats/fs.py
jkonecny12: Thank you for grammar correction. I will change it locally.

jkonecny12:
I like the name change proposal. I will change it locally and remove the ``hungriestFSFromAll()`` method.

jkonecny12:
Updated PR. I hope I didn't missed anything.

jkonecny12:
Hmm I updated this PR which was for old master (1.x-branch now) but now master is v2.

I'm closing this one and I will open new against 1.x-branch

PR #266, opened 2015-11-03
Perform fast format when creating NTFS

vojtechtrefny:
Without the `-f` option mkfs.ntfs takes ages.

### DIFF
--- a/blivet/tasks/fsmkfs.py
+++ b/blivet/tasks/fsmkfs.py
@@ -195,7 +195,7 @@ class NTFSMkfs(FSMkfs):
 
     @property
     def args(self):
-        return []
+        return ["-f"]
 
 class ReiserFSMkfs(FSMkfs):
     ext = availability.MKREISERFS_APP


M4rtinK:
Looks good & thanks for looking into this! :)

jkonecny12:
Looks good to me too :+1:

vpodzime:
Looks good to me too.

dwlehman:
I have been trying to avoid passing non-default arguments to mkfs. We used to pass a variety of custom arguments, but it doesn't make sense for blivet to default to behavior the tools consider non-default. The proper way to fix this would be to add proper support for specifying format options for all format classes.

vojtechtrefny:
I didn't notice we have an option for passing other arguments to mkfs -- https://github.com/rhinstaller/blivet/commit/68019e972e9577157270446d94d3a2dd2184bd20

I will add the `-f` option in blivet-gui.

PR #267, opened 2015-11-03
Update dmdev size when setting up disk images

atodorov:
The size needs to be updated after the sysfs path has been setup.
Related: rhbz#1252703

This is fixed on rhel7-branch but not on master:
https://github.com/rhinstaller/blivet/commit/828ddb09535357a35bde6b3b18ae871bfe2c52ad

The result is --image installs don't work b/c the device has size 0 and anaconda ignores it. I've also added a test case for this. It looks like this test should probably go into anaconda but I'm not sure if the infrastructure there allows for that that's why I've added it here.

### DIFF
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -1534,6 +1534,7 @@ class Populator(object):
                                        exists=True)
                 dmdev.setup()
                 dmdev.updateSysfsPath()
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me. Thanks for tracking this down and adding the test.

PR #268, opened 2015-11-03
Add a class constructor for namedtuples with default values

vpodzime:
namedtuple classes are great as pure data objects one can store things in and
then access them really easily. The problem is that if a new field is added to a
namedtuple, all places creating instances of it has to be modified to provide a
value for that new field. The default_namedtuple class constructor implemented
in this commit constructs namedtuples with default values for (some) fields and
fields without any value passed defaulting to None.

Such default namedtuple classes are still great for storing data and
e.g. passing data to functions/methods grouped into a few objects (like
DeviceInfo, FormatInfo,...) passed as arguments, but allow us to add extra
fields in the future (e.g. a new device attribute to DeviceInfo) without
breaking the API.

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -15,6 +15,7 @@ import warnings
 from decimal import Decimal
 from contextlib import contextmanager
 from functools import wraps
+from collections import namedtuple
 
 import gi
 gi.require_version("BlockDev", "1.0")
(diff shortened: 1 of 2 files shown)


blivet/util.py
```diff
@@ -805,3 +806,43 @@ def the_func(*args, **kwargs):
...
+    The difference between a namedtuple class and this class is that default
+    values may be specified for fields and fields with missing values on
+    initialization being initialized to None.
+
+    :param str name: name of the new class
+    :param fields: field descriptions - an iterable of either "name" or ("name", default_value)
+    :type fields: list of str or (str, object) objects
+    :param str doc: the docstring for the new class (should at least describe the meanings and
+                    types of fields)
+
```
jkonecny12: Please add return to documentation string ideally with type of return.

jkonecny12:
Seems good to me, somehow crazy but usable :D

dwlehman:
Can you whip up a proof of concept illustrating this class as a way to simplify Device constructor argument passing? It doesn't necessarily have to be complete or even work -- just a snippet showing how it would look in practice.

vpodzime:
Sure, this is a short and incomplete example of what I have in mind -- https://gist.github.com/vpodzime/b195b4e24f139cc6b048

dwlehman:
That example looks like a definite improvement to me.

PR #270, opened 2015-11-04
LVM factory LVM metadata

vpodzime:
The key commit here is the third one (https://github.com/rhinstaller/blivet/commit/9c30318f34093253e0c875ab8b25522069f0e654). It fixes an issue our tests caught together with a recent change in libblockdev. These together revealed the fact that the LVMFactory doesn't account for LVM metadata taking space from the PVs.

The other patches are just to make this change nicer.


dwlehman:
These all look good to me.

PR #271, opened 2015-11-05
Improve clarity of error when setting new default fstype fails.

dwlehman:
It's good to log the full format we instantiated, but we should be
inlcuding the type -- not a DeviceFormat instance -- in the error
message.


vpodzime:
Looks good to me.

PR #272, opened 2015-11-06
Add PEP8 compliance checking to 'check' target.

dwlehman:
Moves 'pylint' to its own target, adds a 'pep8' target, and makes
the 'check' target empty with 'check' and 'pep8' as prerequisites.
This allows the user to run either pylint or pep8 separately or
together.

It also makes it possible to plumb a user-specified PYTHONPATH
into the 'pylint' target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ VERSION_TAG=$(PKGNAME)-$(VERSION)
 
 PYTHON=python3
 COVERAGE=coverage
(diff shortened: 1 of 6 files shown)


jkonecny12:
Looks good to me. Nice to have pep8 testing.

vpodzime:
Looks good to me otherwise.

PR #273, opened 2015-11-10
Fix pylint problems with the new default namedtuple stuff.

clumens:
(1) Something's going on with pylinting util_test.py leading to
tracebacks that end like this:

  File "/usr/lib/python3.4/site-packages/astroid/builder.py", line 39, in parse
    return compile(string, "<string>", 'exec', PyCF_ONLY_AST)
  File "<string>", line 2
    class 1(tuple):

No one wants to wade into that, so just disable pylint on that one file
for the moment.

(2) Don't worry about TheDefaultNamedTuple not having an __init__.
That's just one pylint can't figure out.

(3) Get rid of an unnecessary lambda.

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -1040,12 +1040,13 @@ def default_namedtuple(name, fields, doc=""):
             field_names.append(field)
     nt = namedtuple(name, field_names)
 
+    # pylint: disable=no-init
     class TheDefaultNamedTuple(nt):
         if doc:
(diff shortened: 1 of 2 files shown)


jkonecny12:
Looks good to me.

vpodzime:
Looks good to me too, thanks for tracking this down.

dwlehman:
This got pushed as 61c75c89.

PR #274, opened 2015-11-10
Do not try to get LVM cache's size from stats for inactive LV

vpodzime:
If the cached LV is not active, we cannot determine the size of its cache from
the stats because these are only available for activated LVs. However, if the LV
and cache exist, we examined them on devicetree population and got the
information about the size at that point. Thus we can used the stored value if
the LV is no longer active.

Also, keep the logic of when we should be able to get stats in one place.

note: this fixes a bug I hit when trying to install F23 on my workstation -- blivet.reset() fails because when we try to dump the state, the cached LV is inactive.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -1786,8 +1786,10 @@ class LVMCache(Cache):
 
     @property
     def size(self):
-        if self.exists:
-            return self.stats.size
+        # self.stats is always dynamically fetched so store and reuse the value here
+        stats = self.stats
+        if stats:
+            return stats.size
         else:
             return self._size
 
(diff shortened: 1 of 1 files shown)


blivet/devices/lvm.py
```diff
@@ -1786,8 +1786,9 @@ def __init__(self, cached_lv, size=None, md_size=None, exists=False, fast_pvs=No
 
     @property
     def size(self):
-        if self.exists:
-            return self.stats.size
+        stats = self.stats
```
bcl: Why use an intermediate variable here? It isn't unreadable to just say if self.stats and return self.stats.size

blivet/devices/lvm.py
```diff
@@ -1808,9 +1809,11 @@ def exists(self):
 
     @property
     def stats(self):
-        if not self._exists:
+        # to get the stats we need the cached LV to exist and be activated
```
bcl: This would be easier to read if it was flipped, if self._exists and self._cached_lv.status: return LVMCacheStats...  else: return None

blivet/devices/lvm.py
vpodzime: Because the ``stats`` property is not a static value, we always fetch the fresh data from LVM. It is a cheap operation since it is implemented in libblockdev by looking directly at device-mapper's data via libdevmapper, but there's no reason to do it twice in a row here.

blivet/devices/lvm.py
vpodzime: Good point, I'll change this before pushing.

vpodzime:
Note: the test failures are not caused by this commit

PR #275, opened 2015-11-11
Bump version numbers to 2.0 before we forget.

bcl:
This branch is now 2.x development. The first build will need to skip
bumpver to keep the 2.0 release number, for now change it manually so
that we don't accidentally build with a 1.x version.


vpodzime:
Good point, thanks!

bcl:
pushed

PR #276, opened 2015-11-11
edd_test: don't run on non-x86

vathpela:
Since this doesn't make any sense on non-x86 machines anyway, and
apparently I've got some endian bugs in it (which don't really
matter...), just don't run the edd tests on non-x86.

Signed-off-by: Peter Jones pjones@redhat.com

### DIFF
--- a/tests/devicelibs_test/edd_test.py
+++ b/tests/devicelibs_test/edd_test.py
@@ -28,6 +28,8 @@ class FakeEddEntry(edd.EddEntry):
     def __repr__(self):
         return "<FakeEddEntry%s>" % (self._fmt(' ', ''),)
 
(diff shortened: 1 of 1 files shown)


tests/devicelibs_test/edd_test.py
```diff
@@ -28,6 +28,8 @@ def load(self):
     def __repr__(self):
         return "<FakeEddEntry%s>" % (self._fmt(' ', ''),)
 
+@unittest.skipIf(not os.uname().machine in ['i386','i686','x86_64'],
+                 reason='incompatible arch')
```
vpodzime: ``@unittest.skipUnless()`` would fit better in here, I think.

vojtechtrefny:
#632 replaces this

PR #277, opened 2015-11-12
Update the getattr calls with PEP8 property/attribute names

vpodzime:
In order to have a nicer code we have recently moved from camelCase names to
split_words names. While the change covered all the direct accesses (pylint
would catch any problematic ones), it didn't cover the cases where we access the
items indirectly via the getattr() call. This fixes all the getattr() calls that
tried to access any camelCase items.


dwlehman:
Thanks for fixing this.

PR #278, opened 2015-11-13
Do not catch CryptoError when decrypting LUKS format (#1280239)

vojtechtrefny:
LUKS format.setup now raises LUKSError not blockdev.CryptoError


vpodzime:
Looks good to me.

vojtechtrefny:
Pushed.

PR #279, opened 2015-11-13
Only prepend the root to get the ondisk path if there is some

vpodzime:
Otherwise we get things like this:
'/home/user/sources/blivet/None/sys/devices/.../block/sda/size'

which is not really what we want.

@vathpela, could you please add unit tests for the ``Path`` class?


vojtechtrefny:
Looks good to me.

PR #280, opened 2015-11-13
Teardown all descendants devices before tearing disk images down

vpodzime:
DM maps cannot be removed while they are in use so in order to make them
removable, we need to teardown all devices "built" on top of them first. This
makes our examples (and possibly some tests) better clean after themselves.

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -1048,6 +1048,9 @@ class DeviceTree(object):
 
     def teardown_disk_images(self):
         """ Tear down any disk image stacks. """
+        # teardown all devices first so that there's nothing active running on
+        # top of disk images when we try to tear those down
+        self.teardown_all()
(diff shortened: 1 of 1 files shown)


dwlehman:
When disk images are in used all other disks should be hidden, which means it is safe to do this at the beginning of teardown_disk_images instead: `self.teardown_all()`

vpodzime:
That's what I had in there first. But then I was thinking about some potential cases when somebody would like to add some more devices to the devicetree (e.g. as part of the test) and ended up with this "finer-granular" solution.

dwlehman:
I don't want to do anything to suggest, imply, &c support for using a mixture of disk images and normal disks.

vpodzime:
> I don't want to do anything to suggest, imply, &c support for using a mixture of disk images and normal disks.

Works for me, ``teardown_all()`` it is then. :)

PR #285, opened 2015-11-20
Fix few more camelCase leftovers

vpodzime:
These are not so easy to find by pep8 or pylint because they are just strings.


vojtechtrefny:
Looks good to me.

PR #286, opened 2015-11-23
LVM RAID

vpodzime:
This adds an elementary support of non-linear LVs to Blivet. The first patch is unrelated, but also useful. Tests will follow later.

### DIFF
--- a/blivet/devicelibs/raid.py
+++ b/blivet/devicelibs/raid.py
@@ -422,8 +422,16 @@ class RAID0(RAIDn):
     def _get_recommended_stride(self, member_count):
         return member_count * 16
 
+
+class Striped(RAID0):
+    """ subclass with canonical lvm name """
+    name = 'striped'
+    names = [name]
+
 RAID0 = RAID0()
 ALL_LEVELS.add_raid_level(RAID0)
+Striped = Striped()
+ALL_LEVELS.add_raid_level(Striped)
 
 
 class RAID1(RAIDn):
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -714,6 +714,18 @@ def compare(first, second):
     else:
         return (first > second) - (first < second)
 
+
+def dedup_list(alist):
+    """Deduplicates the given list by removing duplicates while preserving the order"""
+    seen = set()
+    ret = []
+    for item in alist:
+        if item not in seen:
+            ret.append(item)
+        seen.add(item)
+    return ret
+
+
 ##
 # Convenience functions for examples and tests
 ##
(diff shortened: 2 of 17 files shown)


blivet/devicelibs/raid.py
```diff
@@ -668,7 +676,7 @@ class Single(ErsatzRAID):
 
 class Dup(RAIDLevel):
 
-    """ A RAID level which expresses one way btrfs metadata may be distributed.
+    """ A RAID level which expresses one way btrfs metadata may be distriuted.
```
jkonecny12: Please fix this small typing error.

vpodzime: Will do, good catch.

blivet/util.py
```diff
@@ -714,6 +714,17 @@ def compare(first, second):
     else:
         return (first > second) - (first < second)
 
+
+def dedup_list(alist):
```
jkonecny12: Please change the name to something like ``deduplicate_list(alist)`` or something but ``dedup_list`` isn't really self explanation.
Also please add doc string here.

jkonecny12:
Otherwise seems good to me as far as I can tell.

blivet/util.py
M4rtinK: I think just "dedup" itself is fine but I think there should b e a docstring noting how the list deduplication works as I can imagine multiple ways how it can be done (keeping order, not keeping order, deduplicating only element sequences, etc.).

dwlehman:
Should we support the "striped" and "mirror" segment types? I think at least one of those is the old lvm implementation and not using the md driver. My suspicion is the lvm team would steer you and anyone else away from using that.

dwlehman:
This looks good overall. It would be nice if md, btrfs, and lvm all accepted the same raid_level ctor argument (it could be an alias/alternative to seg_type for lvm).

vpodzime:
> Should we support the "striped" and "mirror" segment types? I think at least one of those is the old lvm implementation and not using the md driver. My suspicion is the lvm team would steer you and anyone else away from using that.

Since there's no such thing as LVM RAID0, ``striped`` should definitely be supported. And while ``mirror`` is superseded by ``raid1``, it's still quite commonly used, AFAICT, especially because it is the default on RHEL 5 (if you just specify ``--mirrors=1`` and leave the decision about the type to LVM). So I think this decision should probably be left to anaconda and blivet-gui, but not blivet which also needs to deal with existing configurations.

vpodzime:
> It would be nice if md, btrfs, and lvm all accepted the same raid_level ctor argument (it could be an alias/alternative to seg_type for lvm).

Yeah, that would be nice. But I'd like to fully support the ``seg_type`` argument -- e.g. once all the LV classes are unified by the single class and mixins, the ``thin-pool`` segment type will be used to create a thin pool LV.

vpodzime:
Added unit tests and fixed 3 minor things they discovered. Unless somebody complains somehow in the next two days, I'm gonna merge this PR.

jkonecny12:
My quick look found nothing bad so I think it should be good for pushing :) .

PR #287, opened 2015-11-25
Add more class methods for better use of metadata

jkonecny12:
Add new class methods:
1) getRequiredSize(cls, free_space) - takes how much space we need and returns this space + metadata of this FS.
2) biggestOverheadFS(cls, fs_list=None) - Return FS from the list which takes most of the space by metadata. Input list is useful to filter only the required FSs if None is set then it takes biggest from all supported FSs.

This patch is needed for this https://github.com/rhinstaller/anaconda/pull/424 which will go to anaconda.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -74,7 +74,7 @@ class FS(DeviceFormat):
     _writelabelClass = fswritelabel.UnimplementedFSWriteLabel
     # This constant is aquired by testing some filesystems
     # and it's giving us percentage of space left after the format.
-    # This number is more guess then precise number because this
+    # This number is more guess than precise number because this
     # value is already unpredictable and can change in the future...
     _MetadataSizeFactor = 1.0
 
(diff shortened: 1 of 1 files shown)


blivet/formats/fs.py
```diff
@@ -177,14 +177,63 @@ def dict(self):
...
+        if fs_list is None:
+            from . import device_formats
+            fs_list = []
+            for fs_class in device_formats.values():
+                if issubclass(fs_class, cls):
+                    fs_list.append(fs_class)
+        else:
+            for fs_class in fs_list:
+                if not issubclass(fs_class, cls):
+                    raise ValueError("You can provide only Filesystem classes!")
```
vpodzime: *Only file system classes may be provided!* sounds better to me

vpodzime:
Loooks good to me otherwise, but for the 1.x-branch. For master the code needs to be made pep8-compliant.

blivet/formats/fs.py
jkonecny12: Good idea, I'll change it.

jkonecny12:
@dwlehman could you please look on this too?

blivet/formats/fs.py
dwlehman: The `for` and `if` statements above could combined into this:

  `if any(not issubclass(fs_class, cls) for fs_class in fs_list)`

 or this:

  `if not all(issubclass(fs_class, cls) for fs_class in fs_list)`

The former protects against an empty fs_list, while the latter does not.

This code should probably protect against:
1. empty fs_list
2. fs_class w/ _MetadataSizeFactor of 0

jkonecny12:
```
        if fs_list is None:
            from . import device_formats
            fs_list = []
            for fs_class in device_formats.values():
                if issubclass(fs_class, cls):
                    fs_list.append(fs_class)
        elif not fs_list:
            raise ValueError("Empty list is not allowed here!")
        # all items in the list must be subclass of FS class
        elif not all(issubclass(fs_class, cls) for fs_class in fs_list):
            raise ValueError("Only filesystem classes may be provided!")
```
This solution seems best to me but you don't like it I don't have anything against the change.
This will fix 1).

Ad 2)
I don't think this is really needed. This is hard coded constant which is default **1** so there should be nothing with ``_MetadataSizeFactor`` of **0**.

dwlehman: This version looks good to me.Thanks.

jkonecny12:
I updated the PR is it good now?

jkonecny12:
I'm thinking that his could be tested by unit test so I will create one later.
Also I will create the master (2.x) version of this patch.

vpodzime:
Looks good to me.

jkonecny12:
Pushed. The tests and master version will be added as new PR. Closing this one.

PR #289, opened 2015-12-01
Update partedPartition before (un)setting partition flags

vojtechtrefny:
Fixes (un)setting of partition flags when changing format of a partition.

### DIFF
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -552,6 +552,8 @@ class ActionCreateFormat(DeviceAction):
             callbacks.create_format_pre(CreateFormatPreData(msg))
 
         if isinstance(self.device, PartitionDevice):
+            # update partedPartition to point to the partition on the disklabel
+            self.device.partedPartition = self.device.disk.format.partedDisk.getPartitionByPath(self.device.path)
             for flag in partitionFlag.keys():
                 # Keep the LBA flag on pre-existing partitions
                 if flag in [ PARTITION_LBA, self.format.partedFlag ]:


vpodzime:
Looks good to me.

bcl:
Could you add a description of what was going wrong in the commit message? I'm not sure why this is needed, shouldn't that have already been setup?

vojtechtrefny:
I've added more information to the commit message, hopefully it's better now. The problem is we don't change the instance of partedPartition in disklabel.partedDisk.partitions so commitToDisk  actually doesn't change the flags.

dwlehman:
See `PartitionDevice.pre_commit_fixup`. It looks up the `parted.Partition` from the disk right after we call `DiskLabel.reset_parted_disk`. Are you sure `self.parted_partition` is supposed to be a partition on the model format and not the original format? If so, a change should be made in `PartitionDevice.pre_commit_fixup` -- not in `ActionCreateFormat.execute`.

vojtechtrefny:
> Are you sure self.parted_partition is supposed to be a partition on the model format and not the original format?

We call `self.device.disk.format.commit_to_disk()` in `ActionCreateFormat.execute()` so it has to be on the model format.
And we can't add simple "if new partition, use model format and if existing partition use original format" because even "new" partition already exists during format creation.

> If so, a change should be made in PartitionDevice.pre_commit_fixup

Currently `pre_commit_fixup` updates the `parted_partition` using original format. Changing this to the model format works for the flags but `pre_commit_fixup` is used also when destroying partitions. And destroying needs "correct" `parted_partition` on the original format not on the model format.

Maybe there is some easy solution for this but I just don't see it.

dwlehman:
@vojtechtrefny given all of that I think your original patch is the best short-term fix (assuming it doesn't break anything), #297 being a worthy longer-term approach.

dwlehman:
Here are my notes from trying to sort this out today:

Action | parted.Partition and parted.Disk usage
---------|---------------------------------------------------
device create | add partition to disk.format (partition is from now-obsolete non-original disk.format)
device destroy | remove partition from disk.original_format and, if necessary, disk.format
format create | flags/system written to device.parted_partition, which is assumed to be on disk.format. _If the partition was just created, its partition is on disk.format. Otherwise, it's on disk.original_format (result of pre_commit_fixup)._
format destroy | flags/system not modified. _The partition should be on disk.original_format. If it is not going to be destroyed, it should also be on disk.format._

dwlehman:
Base on the table I just posted, I think this should work:

```patch
diff --git a/blivet/actionlist.py b/blivet/actionlist.py
index a9c0a6a..cc93abd 100644
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -227,6 +227,10 @@ class ActionList(object, metaclass=SynchronizedMeta):
         fixup_devices = devices + [a.device for a in self._actions
                                    if a.is_destroy and a.is_device]
         for device in fixup_devices:
+            # Do not run pre_commit_fixup if no device actions are scheduled for this device.
+            if not self.find(device=device, object_type="device"):
+                continue
+
             device.pre_commit_fixup()
 
         # setup actions to create any extended partitions we added
```

vpodzime:
> Base on the table I just posted, I think this should work:
> 
> diff --git a/blivet/actionlist.py b/blivet/actionlist.py
> index a9c0a6a..cc93abd 100644
> --- a/blivet/actionlist.py
> +++ b/blivet/actionlist.py
> @@ -227,6 +227,10 @@ class ActionList(object, metaclass=SynchronizedMeta):
>          fixup_devices = devices + [a.device for a in self._actions
>                                     if a.is_destroy and a.is_device]
>          for device in fixup_devices:
> +            # Do not run pre_commit_fixup if no device actions are scheduled for this device.
> +            if not self.find(device=device, object_type="device"):
> +                continue
> +
>              device.pre_commit_fixup()
> 
>          # setup actions to create any extended partitions we added

Looks promising.

vojtechtrefny:
#490 fixed this

PR #291, opened 2015-12-04
Translation tests

dashea:
This is more or less a template for applying translation-canary to non-autotools projects. This changes the source tarball created in "make release" from everything in git archive (+ ChangeLog + po/) to only the files selected by setup.py sdist. The things that are excluded in the sdist are not used to build the rpm, using sdist makes "make archive" consistent with "make local", and it makes it easier to modify what goes into the source distribution (with translation-canary).

These translation-canary subtree commits are starting to get kind of unwieldy. Maybe it would be better if I used --squash instead of merging the entire history?

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -42,29 +42,92 @@ _Prefix = namedtuple("Prefix", ["factor", "prefix", "abbr"])
 _DECIMAL_FACTOR = 10 ** 3
 _BINARY_FACTOR = 2 ** 10
 
+# TRANSLATORS: 'B' for "Bytes"
 _BYTES_SYMBOL = N_("B")
 _BYTES_WORDS = (N_("bytes"), N_("byte"))
 
 # Symbolic constants for units
 B = _Prefix(1, "", "")
 
-KB = _Prefix(_DECIMAL_FACTOR ** 1, N_("kilo"), N_("k"))
-MB = _Prefix(_DECIMAL_FACTOR ** 2, N_("mega"), N_("M"))
-GB = _Prefix(_DECIMAL_FACTOR ** 3, N_("giga"), N_("G"))
-TB = _Prefix(_DECIMAL_FACTOR ** 4, N_("tera"), N_("T"))
-PB = _Prefix(_DECIMAL_FACTOR ** 5, N_("peta"), N_("P"))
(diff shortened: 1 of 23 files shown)


blivet/size.py
```diff
@@ -42,29 +42,92 @@
...
-MiB = _Prefix(_BINARY_FACTOR ** 2, N_("mebi"), N_("Mi"))
-GiB = _Prefix(_BINARY_FACTOR ** 3, N_("gibi"), N_("Gi"))
-TiB = _Prefix(_BINARY_FACTOR ** 4, N_("tebi"), N_("Ti"))
-PiB = _Prefix(_BINARY_FACTOR ** 5, N_("pebi"), N_("Pi"))
-EiB = _Prefix(_BINARY_FACTOR ** 6, N_("exbi"), N_("Ei"))
-ZiB = _Prefix(_BINARY_FACTOR ** 7, N_("zebi"), N_("Zi"))
-YiB = _Prefix(_BINARY_FACTOR ** 8, N_("yobi"), N_("Yi"))
+KB = _Prefix(_DECIMAL_FACTOR ** 1, 
+             N_("kilo"),
+             # TRANSLATORS: "k" for "kilo-" (x 1,000)
```
dwlehman: I think it would be better to use the scientific notation like you do for the binary units, eg: `10^3` since some places use the comma where we use the decimal point.

dwlehman:
It would probably be nice to squash all of this together. Also, please add a little note to the commit message saying how to handle translation errors turned up by the canary. Other than that, I guess this is a pretty good solution to an unfortunate problem.

dashea:
Updated and rearranged. I redid the subtree commit with --squash, so that's now just two commits: one for the squashed translation-canary history and one for the special merge commit that puts it in its own directory.

I squashed all the pieces that integrate translation-canary into blivet into one commit, 996d3e9653b41e3ec0f65ccbc7ca551b1ed68e3c, and changed the commit message to something that I hope explains things better.

There are two commits that change the blivet build process. One switches to using setup.py sdist instead of git archive to create release tarballs, which allows for changes not kept in git to be distributed in the tarball and makes `make release` consistent with `make local`. The other adds an update to blivet.pot in scratch-bumpver so that changes to blivet.pot are included in testing.

The last commit is the comments on the byte prefixes, and I changed the decimal comments to use 10^x as suggested.

dashea:
Rebased and pushed.

PR #292, opened 2015-12-09
RFC: Populator refactor

dwlehman:
There are a lot of minor commits sprinkled throughout, which I know makes it a bit harder to follow the plot. The important ones are:

67ff0db (Make blivet.populator into a subpackage.)
140cced (Add base classes for type-specific Populator helpers.)
b7a5575 (Move device-type-specific populator bits to populator helpers.)
f0576da (Move format-type-specific bits into populator helpers.)
685d7ca (Add functions to return a helper class based on device data.)
ff93e3e (Use helper getter functions to find populator helpers.)
0efaeeb (Convert Populator to a mixin class.)

Browsing the tip of the branch would probably help more than reading the larger patches. I especially recommend looking through `blivet/populator/` and comparing it to `blivet/populator.py` on master.

I would be willing to consider dropping the patch to convert Populator to a mixin, whether from this PR or completely.


vpodzime:
> Given the from .helper import PopulatorHelper imports, shouldn't the module with the helper be named just helper.py?

I see it's fixed later.

vpodzime:
> I would be willing to consider dropping the patch to convert Populator to a mixin, whether from this PR or completely.

I don't see a reason to do so. I personally like such approach and I'd like to do a similar thing with our LV classes and a goal to have a single class for all LVs.

vpodzime:
Generally, I cannot say how great improvement these changes are from my point of view. So big ACK from me.

dwlehman:
New version should address everything except for preserving the single return statement in that one short method.

dwlehman:
The new version also simplifies mocking of blockdev methods in the unit tests. It isn't possible to use unittest.mock.patch on blockdev stuff directly, eg:

```python
patch("blockdev.loop.get_backing_file")
```

but it is possible to patch them within blivet's namespace, eg: 

```python
patch("blivet.populator.helpers.disklabel.blockdev.mpath_is_mpath_member")
```

vpodzime:
I'm not sure I want to go through all those patches again. So I'm gonna trust you that everything mentioned above is fixed. The ``patch()`` call doesn't seem too bad for tests so it's a nice improvement.

PR #293, opened 2015-12-10
Let Device know about its children

dwlehman:
Instead of a simple child device counter, these patches add a list of children to each `Device` instance. This makes `Device` more complete and makes `DeviceTree.get_children` unnecessary.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -61,10 +61,10 @@ import logging
 log = logging.getLogger("blivet")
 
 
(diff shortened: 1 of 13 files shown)


jkonecny12:
Seems good to me but someone from storage should look on this too.

bcl:
Looks good to me as well.

vpodzime:
Looks good to me too. I again expected this would be worse.

PR #294, opened 2015-12-10
Make blivet thread-safe

dwlehman:
The idea here is to make blivet thread-safe without making a ton of changes. This adds a global reentrant lock and a metaclass. The metaclass wraps every method call with a decorator that acquires the global lock prior to calling the method.


jkonecny12:
Looks good to me. Nice implementation.

bcl:
Very nice technique.

vpodzime:
Oh, this is even nicer than I expected it to be. Really great!

PR #296, opened 2016-01-05
Add _netdev when mounting network devices during installation (#1290046)

bcl:
systemd needs to know that a filesystem depends on the network,
otherwise it will pull the plug before properly unmounting the device.

Resolves: rhbz#1290046


jkonecny12:
Looks good to me.

dwlehman:
You could just cherry-pick 39bbea13 from master.

bcl:
> You could just cherry-pick 39bbea1 from master.

Ha! Yes, that looks like a better choice, thanks.

PR #298, opened 2016-01-06
RFC: Add a file with release notes

vpodzime:
These are more high-level notes about changes done in the releases (at least
major). Useful for people outside of the team, but also for us to easily find
out when and in to what extent something was added.

I believe everybody know what the difference between changelog and release notes is.

Also, obviously, this should be merged after the LVM RAID PR #286 gets merged and then amended with more content related to other big changes.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,3 @@
-include COPYING ChangeLog Makefile python-blivet.spec
+include COPYING ChangeLog Makefile python-blivet.spec release_notes.rst
 recursive-include po *.mo *.po *.pot Makefile
 recursive-include examples *.py
--- a/release_notes.rst
+++ b/release_notes.rst
@@ -1,0 +1,34 @@
+2.0
+====
+
+LVM RAID
+---------
+
+Blivet now recognizes and supports creation of new non-linear LVs. The segment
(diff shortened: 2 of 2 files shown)


dwlehman:
This is a good idea. I wonder if we can reference files in the tree from the wiki so people can find the release notes in either place without us having to maintain two copies.

vpodzime:
> This is a good idea. I wonder if we can reference files in the tree from the wiki so people can find the release notes in either place without us having to maintain two copies.

You mean the wiki here on GitHub?

dwlehman:
Yes.

vpodzime:
We can definitely add a link to the file in the tree. Something like:

The release notes can be found [here] (https://github.com/vpodzime/blivet/blob/master-release_notes/release_notes.rst)

GitHub takes care of the nice rendering.

vpodzime:
Or it should be possible to have the wiki as a git submodule in the sources (http://brendancleary.com/2013/03/08/including-a-github-wiki-in-a-repository-as-a-submodule/), but I think it's not worth it.

vojtechtrefny:
It should be possible to link it (or even directly display/embed) the release notes from our website (Drupal has some support for RST rendering).

PR #299, opened 2016-01-11
Two docs related fixes

vojtechtrefny:
1. Use "sphinx-build-3" and "sphinx-apidoc-3", it fixes the problem with "empty" documentation (not sure if it help with readthedocs, but it fixes the problem when building locally).

2. Fix "class:`~..." links in intro. Blivet 2.0 API change, devices split up and other changes broke them.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -3,8 +3,8 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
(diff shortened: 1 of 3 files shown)


vpodzime:
These both look good to me. Thanks for improving our docs!

vpodzime:
Merging this as I need it for my work on API/non-API documentation tricks.

PR #301, opened 2016-01-13
Fix AttributeError in populator LUKS helper

vojtechtrefny:


vpodzime:
Looks exactly like my suggestion and thus good to me. :D

dwlehman:
Looks good to me.

vojtechtrefny:
Pushed.

PR #302, opened 2016-01-14
followups for various recent commits

dwlehman:


vpodzime:
The first two patches look good to me.

PR #303, opened 2016-01-15
Blacklisted attrs should be unwrapped, not omitted.

dwlehman:
Thanks for pointing out the problem, @vojtechtrefny.


vojtechtrefny:
Looks good to me.

vpodzime:
Same here.

jkonecny12:
Could you please rename ``wrapped`` list to something else because now it has wrapped and blacklisted items too.

dwlehman:
It has always contained both wrapped and unwrapped attributes. I will try to think of a better name for the dict.

jkonecny12:
Looks much better to me now. Thank you.

PR #305, opened 2016-01-15
storaged iscsi

vpodzime:
This is a totally untested PR just ready for comments. It replaces all the calls to *libiscsi* with DBus calls to *storaged* and its iSCSI functionality. Is it worth testing and getting into a shape that could be merged?

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -67,6 +67,13 @@ def has_iscsi():
 
 
 NodeInfo = namedtuple("NodeInfo", ["name", "tpgt", "address", "port", "iface"])
+TargetInfo = namedtuple("TargetInfo", ["ipaddr", "port"])
+
+
+class LoginInfo(object):
+    def __init__(self, node, logged_in):
+        self.node = node
+        self.logged_in = logged_in
 
 
 def _to_node_infos(variant):
@@ -97,8 +104,8 @@ class _iSCSI(object):
     """
 
     def __init__(self):
-        # Dictionary of discovered targets containing list of (node,
-        # logged_in) tuples.
+        # Dictionary mapping discovered TargetInfo data to lists of LoginInfo
+        # data.
         self.discovered_targets = {}
         # This list contains nodes discovered through iBFT (or other firmware)
         self.ibft_nodes = []
@@ -171,9 +178,8 @@ class _iSCSI(object):
     def active_nodes(self, target=None):
         """Nodes logged in to"""
         if target:
-            return [node for (node, logged_in) in
-                    self.discovered_targets.get(target, [])
-                    if logged_in]
+            return [info.node for info in self.discovered_targets.get(target, [])
+                    if info.logged_in]
         else:
             return [info.node for info in itertools.chain(*list(self.discovered_targets.values()))
                     if info.logged_in] + self.ibft_nodes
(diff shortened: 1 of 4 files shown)


blivet/iscsi.py
```diff
@@ -330,95 +324,64 @@ def discover(self, ipaddr, port="3260", username=None, password=None,
...
-                self.discovered_targets[(ipaddr, port)]
-                if not logged_in]
+        return [info.node for info in self.discovered_targets[TargetInfo(ipaddr, port)]
+                if not info.logged_in]
 
     def log_into_node(self, node, username=None, password=None,
                       r_username=None, r_password=None):
-        """
-        Raises IOError.
-        """
```
jkonecny12: It would be nice if you could instead of only removing this docstring also add new one.

blivet/iscsi.py
```diff
@@ -330,95 +324,64 @@ def discover(self, ipaddr, port="3260", username=None, password=None,
...
-            # Close the writing end of the pipe in the parent
-            con_write.close()
-
-            with con_recv:
-                try:
-                    (ok, data) = con_recv.recv()
-                    if not ok:
-                        log.debug("iSCSI: exception raised when "
-                                  "discover_sendtargets process called: %s",
-                                  str(data))
```
jkonecny12: I want to ask. You removed this ugly error detection which is great but how this will behave when iscsi is for some reason throw an exception?

jkonecny12:
I really like this idea and implementation. It should be heavily tested before this goes out but it's much cleaner and better than before. You have my ACK for this one.

blivet/iscsi.py
vpodzime: That's what I hope to figure out during testing these changes. :) No idea right now, but I hope we will get some nice and descriptive GDBus error.

blivet/iscsi.py
vpodzime: Sure thing, I'm gonna add a whole bunch of docstrings here and there before I merge this.

blivet/iscsi.py
jkonecny12: I'm looking forward for the results.

dwlehman:
@rvykydal Can you please read the commit message for 9601350 and see what you think about the second paragraph?

rvykydal:
I agree with Vratislav that iscsi-initiator-utils should handle the case with a node.startup set to autostart which had been already started in initramfs gracefully. Also I checked that setting autostart with current implementation works - actually the node.startup is set to onboot after system start in both cases (node.startup set to manual or autostart), so I think iscsi utils are handling starting of node in initramfs fine.

vpodzime:
This is a tested and "seemed-to-be-working" version of the patches. Don't know why GH puts the first commit in here (the release notes file), that one is already on master. I tested this code (ported to branch 1.x) in anaconda and it seems to be working great.

vpodzime:
Thanks @rvykydal and @jkonecny12 for their help with testing!

vpodzime:
Just wanted to mention that I'd like to port this to *1.x-branch* so that we start using the new API in Fedora 24.

jkonecny12:
Have this in F24 would be super nice.
Thank you for the patch. It could prevent many iSCSI issues in future.

dwlehman:
Looks good to me for both 1.x and master.

dwlehman:
Well, except that you added release notes about lvm-dbus.

vpodzime:
> Well, except that you added release notes about lvm-dbus.

I didn't, that was GitHub. No idea why.

vpodzime:
Added three more commits implementing a check for storaged-iscsi availability and making iscsi operations fail-safe.

dwlehman:
Looks good. Thanks.

PR #306, opened 2016-01-19
pylint fixes for rawhide

dwlehman:


blivet/devices/lvm.py
```diff
@@ -1523,14 +1523,14 @@ def _update_format_from_origin(self):
             fmt._chrooted_mountpoint = None
             fmt.device = self.path  # pylint: disable=no-member
 
-        super(LVMSnapShotBase, self)._set_format(fmt)
+        super(LVMSnapShotBase, self)._set_format(fmt)  # pylint: disable=no-member
```
jkonecny12: IMHO to be Python3 compatible it would be nice to change this ``super(LVMSnapShotBase, self)`` to just ``super()``.

blivet/devices/lvm.py
```diff
@@ -1523,14 +1523,14 @@ def _update_format_from_origin(self):
...
 
     def _set_format(self, fmt):
         # If a snapshot exists it can have a format that is distinct from its
         # origin's. If it does not exist its format must be a copy of its
         # origin's.
         if self.exists:  # pylint: disable=no-member
-            super(LVMSnapShotBase, self)._set_format(fmt)
+            super(LVMSnapShotBase, self)._set_format(fmt)  # pylint: disable=no-member
```
jkonecny12: Same here.

jkonecny12:
Seems good to me otherwise.

vpodzime:
Looks good to me too otherwise. And you can leave the ``super()`` calls as they are, I'm just removing them in my new code anyway. :)

PR #308, opened 2016-01-22
LUKS resize

vojtechtrefny:
LUKS encrypted partitions resize support.

Known issues:
 - `LUKSFormat` has "wrong" `min_size` -- because minimal size depends on minimal size of LUKS device's format, it's not possible to set it for LUKS format. Slave device, LUKS device and filesystem have correct min_size.
 - Both `LUKSFormat` and `LUKSDevice` don't have `max_size` -- it depends on the slave device. We don't set max_size for formats (just limits), so it probably makes sense to set the max size correctly just for the slave device. But it would be easy to set the max_size correctly for the LUKS device.
 - Tasks -- I just took code from mulhern's original pull request and made changes to make pep8 and pylint happy. It works, but I don't have idea how :)
 - `FSError` and `FSResizeError` -- original `do_resize` and `_set_target_size` raised only these exceptions but it doesn't make sense for "non-filesystems". I've tried to change it to raise `DeviceFormatError` and `FormatResizeError` where it makes sense. Not sure if this is the best way.
 - There are no tests. I will add some next week.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -38,6 +38,7 @@ from .deviceaction import ActionCreateDevice, ActionCreateFormat, ActionDestroyD
 from .deviceaction import ActionDestroyFormat, ActionResizeDevice, ActionResizeFormat
 from .devicelibs.edd import get_edd_dict
 from .devicelibs.btrfs import MAIN_VOLUME_ID
+from .devicelibs.crypto import LUKS_METADATA_SIZE
 from .errors import StorageError
 from .size import Size
 from .devicetree import DeviceTree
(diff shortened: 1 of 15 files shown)


vpodzime:
Looks good to me otherwise. I'll give my ACK once the ``lsblk`` stuff is gone.

vojtechtrefny:
Updated version based on @vpodzime comments.

Major change is the `LUKSSize` task now reads the size from `/sys/block...` instead of using `lsblk`.

vojtechtrefny:
New version. Fixed rebase mistake (change in a wrong commit) and added fix for canceling resize action on luks devices.

vojtechtrefny:
Update: Added test, few fixes based on @jkonecny12 comments.

jkonecny12:
It looks good to me now. If it's OK with everyone I'm giving you ACK.

dwlehman:
It seems like the resize machinery all belongs in the format. I don't care if LUKSDevice is resizable but, if it is, the resize method should probably just be `pass`.

vojtechtrefny:
> It seems like the resize machinery all belongs in the format. I don't care if LUKSDevice is resizable but, if it is, the resize method should probably just be pass.

The idea here is that we don't create a resize action for the LUKS format but only for the LUKS device. It's mostly because of the `target_size` -- we need to set it for both format and the device and only the device "knows" about the LUKS format -- so with just a one action, we can set target size for both LUKS format and device. And without a resize action for the format, LUKS device `resize` method has to call the `do_resize` for the LUKS format.

dwlehman:
We could also make `LUKSDevice.size` grab its value from the parent/slave so it doesn't need to store its own target size.

I would rather make the convention to schedule actions on both the luks format and the luks device. The action on the luks device would be a no-op, but it makes it much cleaner than having to remember which of the two you're supposed to schedule.

vojtechtrefny:
New version based on @dwlehman comments.

vojtechtrefny:
Fixed `max_size` and `target_size` setter so it's not possible to make luks device larger than the slave device.

vpodzime:
Other than the nitpicks above this looks good to me.

dwlehman:
Looks good to me, but please make sure you reset the date on one or all (`git rebase --ignore-date origin/master`) of the commits when rebasing to trigger jenkins so we can be sure these are clean from pylint's perspective.

PR #310, opened 2016-01-25
LVM classes unification

vpodzime:
These commits change how we represent LVs. Instead of having special classes for various types of LVs, we now have a single class with the type of LV being more a dynamic property than a once-given thing. This will allow us to implement various LVM-related features in the future that have so far been limited by our multi-class representation.

I'm not adding any new functionality or features so no tests are added. Just the existing ones are modified to continue working.

I'll add a section to the *release_notes.rst* file tomorrow as an extra commit to this PR.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -32,7 +32,7 @@ from pykickstart.constants import AUTOPART_TYPE_LVM, CLEARPART_TYPE_ALL, CLEARPA
 
 from .storage_log import log_method_call, log_exception_info
 from .devices import BTRFSDevice, BTRFSSubVolumeDevice, BTRFSVolumeDevice
-from .devices import LVMLogicalVolumeDevice, LVMThinLogicalVolumeDevice, LVMThinPoolDevice, LVMVolumeGroupDevice
+from .devices import LVMLogicalVolumeDevice, LVMVolumeGroupDevice
 from .devices import MDRaidArrayDevice, PartitionDevice, TmpFSDevice, device_path_to_name
 from .deviceaction import ActionCreateDevice, ActionCreateFormat, ActionDestroyDevice
 from .deviceaction import ActionDestroyFormat, ActionResizeDevice, ActionResizeFormat
@@ -916,6 +916,11 @@ class Blivet(object, metaclass=SynchronizedMeta):
         else:
             vg = parent
 
+        if thin_volume:
+            kwargs["seg_type"] = "thin"
+        if thin_pool:
+            kwargs["seg_type"] = "thin-pool"
+
         mountpoint = kwargs.pop("mountpoint", None)
         if 'fmt_type' in kwargs:
             kwargs["fmt"] = get_format(kwargs.pop("fmt_type"),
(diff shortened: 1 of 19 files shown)


vpodzime:
Feel free to ignore the first patches that are not part of this PR. No idea why GH thinks they are.

vpodzime:
Instead of reviewing the changes in the commit bdacb4c (Rewrite the code...), have a look at the (file with the changes)[https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py] instead.

jkonecny12:
I can't add comment on that file :( so https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L545 change ``too`` to ``to`` in comment.

jkonecny12:
If it's not required somewhere could you please change ``0``,``1``,``-1`` to Enum or constants or anything similar. I think the time when you get ``-1`` from program and you need to know what that means are somehow old now.

https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L792

jkonecny12:
I think instead of disabling pylint error it would be better to only change name of ``value`` to ``_value``.

https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1027

vpodzime:
> If it's not required somewhere could you please change 0,1,-1 to Enum or constants or anything similar. I think the time when you get -1 from program and you need to know what that means are somehow old now.
> 
> https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L792

It is required by the place where this method is called (determining the particular error/problem).

vpodzime:
> I think instead of disabling pylint error it would be better to only change name of value to _value.
> 
> https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1027

I'm all for it as long as it helps.

jkonecny12:
Only small idea. What about changing these ``def type`` return value to Enum some enum. If you will use this for comparison then you need to find exact string. The Enum could be helpful here thought. 

https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1031

jkonecny12:
>> I think instead of disabling pylint error it would be better to only change name of value to _value.
>> 
>>     https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1027
> 
> I'm all for it as long as it helps.

It will be when you use ``name`` instead of ``_``. Sorry for the first bad variant. Also it's on more places if you want to change it.

jkonecny12:
Try to use ``@staticmethod`` to remove this pylint error.

https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1190

vpodzime:
> I think instead of disabling pylint error it would be better to only change name of value to _value.
> 
> https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1027

Thanks for the tip, it seems to work. Fixing this plus a bunch of other places.

vpodzime:
> Only small idea. What about changing these def type return value to Enum some enum. If you will use this for comparison then you need to find exact string. The Enum could be helpful here thought.
> 
> https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1031

I absolutely agree, but that's something we should do across all blivet codebase at once in some followup patch (set).

vpodzime:
> Try to use @staticmethod to remove this pylint error.
> 
> https://github.com/vpodzime/blivet/blob/bdacb4c311b02431090ae2cb999fe7d366239afe/blivet/devices/lvm.py#L1190

No. I'll rather keep telling pylint to shut up than lying to myself and other blivet developers. It's a decorator, not a static method of the class.

jkonecny12:
Ok then, I think it should be fine now.

dwlehman:
Looks good to me.

PR #312, opened 2016-02-03
1.x-branch pylint cleanup

bcl:
Fixed everything except a couple LVMSnapShotBase errors and the intermittent element error.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -325,7 +325,7 @@ class Size(Decimal):
         return Size(Decimal.__mul__(self, other))
     __rmul__ = __mul__
 
-    def __div__(self, other, context=None):
+    def __div__(self, other, context=None):             # pylint: disable=unused-argument
         if six.PY2:
             # This still needs to be ignored by pylint, because it will get
             # through the above guard.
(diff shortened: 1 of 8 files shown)


blivet/size.py
```diff
@@ -325,7 +325,7 @@ def __mul__(self, other, context=None):
         return Size(Decimal.__mul__(self, other))
     __rmul__ = __mul__
 
-    def __div__(self, other, context=None):
+    def __div__(self, other, context=None):             # pylint: disable=unused-argument
```
vpodzime: why so many spaces?

vpodzime:
Looks good to me.

blivet/size.py
bcl: > why so many spaces? "Two ought to be enough." :)

makes it easier to read.

dwlehman:
I think we could use 25afe59 ("Cleanup wildcard...") and fb45fe1 ("Include python3-bugzilla...") on master, too.

blivet/size.py
vpodzime: Fair enough :)

bcl:
Pushed to 1.x-branch and cherry-picked those 2 to master.

PR #316, opened 2016-02-09
some pre-2.0 housekeeping

dwlehman:


vpodzime:
Looks great to me. I always like to see code being replaced with docs say it's gone. :)

snbueno:
Agree, looks good; especially since this set removes more code than it introduces. ;)

PR #319, opened 2016-02-15
Do not call pvmove when removing an empty PV from VG

vojtechtrefny:
We now know both PV free and PV size so we can finally fix the bug with removing empty PV from a VG.


vpodzime:
Looks good to me otherwise.

dwlehman:
Looks okay to me.

vojtechtrefny:
Pushed.

PR #320, opened 2016-02-15
Fix TypeError/AttributeError when adding thinpools and snapshots

vojtechtrefny:


vpodzime:
Just adding a note that I told @vojtechtrefny to create properties for those two attributes -- they are not currently ment to be set from the outside post creation of the objects.

Looks good to me.

vojtechtrefny:
Pushed.

PR #322, opened 2016-02-16
Fix current_size name in StorageDevice.update_size

vojtechtrefny:


dwlehman:
Yikes -- how did that go so long without being noticed?

PR #323, opened 2016-02-16
More 2.0 stuff

dwlehman:

PR #324, opened 2016-02-19
Fix non-ASCII comment

vojtechtrefny:
Gettext doesn't like non-ascii comments.


jkonecny12:
Looks good to me.

dwlehman:
Fixed without altering your name in the last patch of #323.

vojtechtrefny:
Oh, I didn't notice. I just found my daily builds are failing because of this.

PR #325, opened 2016-02-24
Use _netdev mount option as needed. (#1290046)

bcl:
The intention is to pass the option as needed when mounting filesystems
in addition to putting it in /etc/fstab.

(cherry picked from commit 39bbea13f22c6137625d55cd4902f09be6dc032c)

Resolves: rhbz#1290046


vpodzime:
Looks good to me.

bcl:
Pushed.

PR #326, opened 2016-03-02
1.x libbytesize

vpodzime:
This is the same thing as PR #251 just for the *1.x-branch*. I think it's worth it to switch the 1.x blivet to libbytesize. If nothing else we would get a lot more testing and attention for libbytesize.


vpodzime:
We are not going to do this.

PR #330, opened 2016-03-03
Fixes for cached LV VG space used

vpodzime:
These fixes some inconsistencies and problems @vojtechtrefny found when working on LVM cache and LVM RAID support in *blivet-gui*.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -495,7 +495,14 @@ class LVMVolumeGroupDevice(ContainerDevice):
 
         """
         # TODO: report correctly/better for existing VGs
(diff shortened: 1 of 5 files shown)


vpodzime:
This needs some more refinements as the tests fail right now.

dwlehman:
That's okay with me, but it will have to be addressed before we pull this into a release.

vpodzime:
> That's okay with me, but it will have to be addressed before we pull this into a release.

Fair enough. However, I'm not sure how much time I want to spend on it if we may end up removing/avoiding the padding entirely as suggested by the LVM team recently.

vpodzime:
I need to take a second look on this. In any case, these changes are going to -devel branches now so I'm closing this PR.

PR #331, opened 2016-03-08
Fix the parsing of translated sizes.

dashea:
Use only one translation for "B" and "b" to mean bytes. This way partial
translations will not cause problems when attempting to parse a size
string after it has changed case.

Needs acks still.

Of the languages that have different or partial translations for "B" vs "b", it looks like all but one (sr) has the correct translations for "B", which is the one that this commit keeps.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -59,14 +59,20 @@ _binaryPrefixes = [_Prefix(1024, N_(b"kibi"), N_(b"Ki")),
                    _Prefix(1024**7, N_(b"zebi"), N_(b"Zi")),
                    _Prefix(1024**8, N_(b"yobi"), N_(b"Yi"))]
 
-_bytes = [N_(b'B'), N_(b'b'), N_(b'byte'), N_(b'bytes')]
+# Handle 'B' separately so that it can be localized without translating
+# both 'B' and 'b'
+_bytes_letter = N_(b'B')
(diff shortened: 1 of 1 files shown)


blivet/size.py
```diff
@@ -59,14 +59,21 @@
...
 
-_bytes = [N_(b'B'), N_(b'b'), N_(b'byte'), N_(b'bytes')]
+# Handle 'B' separately so that it can be localized without translating
+# both 'B' and 'b'
+_bytes_letter = N_(b'B')
+_bytes_words = [N_(b'byte'), N_(b'bytes')]
 _prefixes = _binaryPrefixes + _decimalPrefixes
 
+def _bytes():
+    return [_bytes_letter + _lowerASCII(_bytes_letter)] + _bytes_words
```
vpodzime: Why does this needs to be a function?

vpodzime:
Looks good to me otherwise.

blivet/size.py
dashea: It doesn't, I suppose. I just did it that way for consistency with _xlated_bytes()

vpodzime: Makes sense, thanks for the explanation.

dashea:
Updated to change _bytes back to a list

vpodzime:
Looks even better to me.

dashea:
Moved _lowerASCII to before _bytes so that it actually works, and pushed.

PR #333, opened 2016-03-09
Add zanata-python-client to TEST_DEPENDENCIES

bcl:


jkonecny12:
Looks good to me. Curiosity question, this is required because of Translation Canary or for some other reason?

jkonecny12:
Thank you. It would be nice to have this explanation in the commit message. But even without it, still ACK :).

bcl:
Pushed. Wasn't needed on 1.x-branch

PR #336, opened 2016-03-11
Add xfs to default filesystems (#1256249)

kellinm:
Running blivet tests as a non-root user leads to failed instantiation of
the core Blivet class if the ext4 kernel module is not loaded on the
test system. Added xfs to the list of acceptable default file systems.

Related: rhbz#1256249


kellinm:
This needed a re-write - the patch, title, and message are all incorrect.  Closing and resubmitting as a fresh PR.

PR #339, opened 2016-03-11
Kickstart missing bootloader partitions (#1256249)

kellinm:

Resolves: rhbz#1256249
Resolves: rhbz#1242666

### DIFF
--- a/tests/blivet_test.py
+++ b/tests/blivet_test.py
@@ -1,0 +1,42 @@
+import unittest
+from unittest.mock import Mock
+from pykickstart.version import returnClassForVersion
+from blivet.osinstall import InstallerStorage
+from blivet.devices import PartitionDevice
+from blivet import formats
+import re
+
+class BlivetTestCase(unittest.TestCase):
+    '''
+    Define tests for the Blivet class
+    '''
+    def setUp(self):
+        pass
+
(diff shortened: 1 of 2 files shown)


tests/blivet_test.py
```diff
@@ -0,0 +1,53 @@
...
+        my_blivet.ksdata = my_ksdata
+
+        # add device and update ksdata
+        my_blivet.devicetree._add_device(my_bootloader_device)
+        my_blivet.devicetree._add_device(my_root_device)
+
+        my_blivet.update_ksdata()
+
+        my_result = re.search('part prepboot --fstype="prepboot" --size=5',
+                              str(my_blivet.ksdata))
```
dwlehman: Putting the entire line in there makes it susceptible to kickstart outputting the arguments in a different order, omitting the quotes, &c, none of which is relevant to what you are testing here. You can do it a bit more directly like this:
```python

self.assertTrue("part prepboot" in str(my_blivet.ksdata))

```

dwlehman:
Please open a pull request based on the `2.0-devel` branch. We've changed our branching scheme as described [here](https://gist.github.com/dwlehman/ceaf09d7ae1ac1c9e13da79285f5f738).

PR #340, opened 2016-03-11
Fix blivet constructor fs support check (#1242666)

kellinm:

Related: rhbz#1242666

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -46,13 +46,13 @@ def register_device_format(fmt_class):
     log.debug("registered device format class %s as %s", fmt_class.__name__,
                                                          fmt_class._type)
 
-default_fstypes = ("ext4", "ext3", "ext2")
+default_fstypes = ("xfs", "ext4", "ext3", "ext2")
 def get_default_filesystem_type():
     for fstype in default_fstypes:
         try:
-            supported = get_device_format_class(fstype).supported
+            supported = getFormat(fstype).supported
         except AttributeError:
-            supported = None
+            supported = False
 
         if supported:
             return fstype


blivet/formats/__init__.py
```diff
@@ -46,13 +46,13 @@ def register_device_format(fmt_class):
     log.debug("registered device format class %s as %s", fmt_class.__name__,
                                                          fmt_class._type)
 
-default_fstypes = ("ext4", "ext3", "ext2")
+default_fstypes = ("ext4", "xfs", "ext3", "ext2")
```
vpodzime: Since *xfs* is the default fstype in RHEL7 now, I think it should go to the first place here.

vpodzime:
> Fixed the guard to check the boolean value of the property rather than
> checking for the existence of the pointer.

Existence of which pointer? I don't really understand this.

vpodzime:
Anyway, this change introduces a potential bug where we would tell people some format is supported in their case while it wouldn't be. The correct place where to deal with the issue in tests are the tests. They should mock the ``FS.load_module()`` method.

vpodzime:
The same applies to #341

dwlehman:
In this line:

`get_device_format_class(fstype).supported`

it is effectively evaluating `Ext4.supported`, which has a value that is an instance of "property". The value of `supported` is only meaningful in the context of the calling code once that class has been instantiated.

Here's an example illustrating the bug:

```python
>>> class X:
...   @property
...   def x(self):
...     return 26
... 
>>> X.x
<property object at 0x7f05e836fea8>
>>> X().x
26
>>> bool(X.x)
True

```

dwlehman:
It does makes sense to mock things such that the tests that are not testing anything related to the default filesystem can pass regardless of whether the default filesystem's support is present on the testing system.

kellinm:
Moved XFS to head of the line.

PR #341, opened 2016-03-11
Fix blivet constructor fs support check (#1256249)

kellinm:

Related: rhbz#1256249


dwlehman:
Please open a new pull request against the `2.0-devel` branch.

PR #342, opened 2016-03-14
Mock all blockdev's listing functions for populator tests

vpodzime:
We don't want to use the information from the system the tests are running on
and actually we cannot even grab that information because the tests may not be
run as root. To prevent issues and exceptions, we can easily just mock all three
LVM listing functions - lvs(), vgs() and pvs().


vpodzime:
See also https://github.com/rhinstaller/libblockdev/pull/68

jkonecny12:
Looks good to me.

PR #343, opened 2016-03-15
CONTRIBUTING update

vpodzime:
Two minor updates to our CONTRIBUTING file.


vojtechtrefny:
Looks good to me.

dwlehman:
Hmm, I guess this has become inaccurate since we've changed the branching policy. Maybe just push the one that removes the python3-check bit?

vpodzime:
> Hmm, I guess this has become inaccurate since we've changed the branching policy. Maybe just push the one that removes the python3-check bit?

Yeah, that's the only bit that makes sense. I think this can go just to *master*. Or do you think it should go to the *-devel* branches too?

vpodzime:
Obsoleted by PR #362

PR #344, opened 2016-03-18
Allow custom chunk size specification for MDRaidArrayDevice

vojtechtrefny:


blivet/devices/md.py
```diff
@@ -188,6 +193,35 @@ def level(self, value):
...
+            raise ValueError("cannot set chunk size for an existing device")
+
+        self._chunk_size = newsize
+
+    def read_chunk_size(self):
+        log_method_call(self, exists=self.exists, path=self.path,
+                        sysfs_path=self.sysfs_path)
+        chunk_size = Size(0)
+        if self.exists and os.path.exists(self.path) and \
+           os.path.isdir(self.sysfs_path):
```
dwlehman: I wonder if it would be better to use `self.status` here.

blivet/devices/md.py
```diff
@@ -188,6 +193,35 @@ def level(self, value):
         self._level = level
 
     @property
+    def chunk_size(self):
+        if self.exists and self._chunk_size == Size(0):
+            self._chunk_size = self.read_chunk_size()
+        return self._chunk_size
```
dwlehman: If someone accesses `chunk_size` while the array is inactive it will overwrite the potentially correct chunk size with `Size(0)`. The way we handle many things of this sort is to update the value whenever we know the array is active, eg: in `__init__` if the device is active and then again in `_post_setup`.

dwlehman: Oops -- clearly it will not overwrite a non-zero value here. My mistake.

blivet/devices/md.py
vojtechtrefny: Yes, it makes more sense here, `self.status` already checks if the device exist and if the `path`/`sysfs_path` exist.

tests/devices_test/md_test.py
```diff
@@ -0,0 +1,49 @@
...
+from blivet.devicelibs import mdraid
+
+DEVICE_CLASSES = [
+    MDRaidArrayDevice,
+    StorageDevice
+]
+
+
+@unittest.skipUnless(not any(x.unavailable_type_dependencies() for x in DEVICE_CLASSES), "some unsupported device classes required for this test")
+class LVMDeviceTest(unittest.TestCase):
```
vpodzime: nice name for an MDArray test ;)

vpodzime:
Other than the above nitpick this looks good to me.

tests/devices_test/md_test.py
vojtechtrefny: Fixed, thanks.

PR #346, opened 2016-03-22
Fix the _bytes string list

dashea:
The start of the list of is supposed to be the upper and lowercase
versions of the abbreviation for 'byte', not both cases at once.

Related: rhbz#1314301


vpodzime:
Looks good to me.

dashea:
Pushed.

PR #347, opened 2016-03-23
Anaconda to 2.0

dwlehman:
This is the stuff I had to fix to make anaconda able to run using blivet-2.0.

### DIFF
--- a/blivet/autopart.py
+++ b/blivet/autopart.py
@@ -206,7 +206,7 @@ def _schedule_partitions(storage, disks, implicit_devices, min_luks_entropy=0, r
 
     # basis for requests with required_space is the sum of the sizes of the
     # two largest free regions
-    all_free = (Size(reg.get_length(unit="B")) for reg in get_free_regions(disks))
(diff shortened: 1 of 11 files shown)


blivet/autopart.py
```diff
@@ -206,7 +206,7 @@ def _schedule_partitions(storage, disks, implicit_devices, min_luks_entropy=0, r
 
     # basis for requests with required_space is the sum of the sizes of the
     # two largest free regions
-    all_free = (Size(reg.get_length(unit="B")) for reg in get_free_regions(disks))
+    all_free = (Size(reg.getLength(unit="B")) for reg in get_free_regions(disks))
```
vpodzime: The regions are parted partitions or something? IOW, are they from outside of blivet? If not, the ``getLength()`` method should be renamed instead.

vpodzime:
Looks good to me otherwise.

blivet/autopart.py
dwlehman: They are `parted.Geometry` instances.

PR #348, opened 2016-03-28
Installer isolation: part one

snbueno:
I don't know what to call this thing, but I'm doing the GitHub thing now.

This is just the beginning pieces of isolating installer-specific code that lives in blivet. This time around, the storage config is moved to blivet.osinstall out of blivet.blivet, and blivet.blivet.Blivet is significantly cleaned up as well. Everything's going to blivet.osinstall for the interim; eventually it's all going to be moved to anaconda when the timing is right, etc.

The commits tell a more robust story.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -27,8 +27,6 @@ import contextlib
 import time
 import functools
 
-from pykickstart.constants import AUTOPART_TYPE_LVM
-
 from .storage_log import log_method_call, log_exception_info
 from .devices import BTRFSSubVolumeDevice, BTRFSVolumeDevice
 from .devices import LVMLogicalVolumeDevice, LVMVolumeGroupDevice
@@ -45,11 +43,7 @@ from .formats import get_default_filesystem_type
 from .flags import flags
 from .platform import platform as _platform
 from .formats import get_format
-from .osinstall import StorageDiscoveryConfig find_existing_installations
 from . import arch
-from .iscsi import iscsi
-from .fcoe import fcoe
-from .zfcp import zfcp
 from . import devicefactory
 from . import get_bootloader, get_sysroot, short_product_name, __version__
 from .threads import SynchronizedMeta
@@ -70,13 +64,10 @@ class Blivet(object, metaclass=SynchronizedMeta):
         self.ksdata = ksdata
         self._bootloader = None
 
-        self.config = StorageDiscoveryConfig()
-
         # storage configuration variables
         self.do_autopart = False
         self.clear_part_choice = None
         self.encrypted_autopart = False
-        self.autopart_type = AUTOPART_TYPE_LVM
         self.encryption_passphrase = None
         self.encryption_cipher = None
         self.escrow_certificates = {}
@@ -105,7 +96,6 @@ class Blivet(object, metaclass=SynchronizedMeta):
                                      disk_images=self.disk_images)
         self.roots = []
         self.services = set()
-        self._free_space_snapshot = None
 
     def do_it(self, callbacks=None):
         """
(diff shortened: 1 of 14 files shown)


blivet/osinstall.py
```diff
@@ -1094,6 +1098,284 @@ def get(self, key, default=None):
         return self.mappings.get(key, default)
 
 
+class InstallerStorageConfig(Blivet):
+    def __init__(self, ksdata=None):
+        Blivet.__init__(self, ksdata=ksdata)
```
dwlehman: You can just remove the `__init__` method since it doesn't do anything.

dwlehman:
The commit message for b7a78f1 needs updating (that commit isn't moving `StorageDiscoveryConfig` like it says).

Please move the docstring updates from 153bbdb to the commits that make the changes you're documenting.

snbueno:
Implemented all of dlehman's comments. Hopefully didn't accidentally implement anything else when I rebased against master just now.

blivet/osinstall.py
```diff
@@ -1094,6 +1099,584 @@ def get(self, key, default=None):
...
 
 
+class InstallerStorageConfig(Blivet):
+    """ Top-level class for managing installer-related storage configuration. """
+    def __init__(self, ksdata=None):
+        """
+            :keyword ksdata: kickstart data store
+            :type ksdata: :class:`pykickstart.Handler`
+        """
+        super(InstallerStorageConfig, self).__init__(ksdata=ksdata)
```
dwlehman: Just do `super().__init__(ksdata=ksdata)` here.

blivet/osinstall.py
```diff
@@ -1094,6 +1099,653 @@ def get(self, key, default=None):
         return self.mappings.get(key, default)
 
 
+class InstallerStorageConfig(Blivet):
```
vpodzime: The ``Config`` part of this class' name makes me think it is just some static configuration not an active object like ``Blivet`` is. What about just ``InstallerStorage``? Or ``InstallerBlivet``?

blivet/blivet.py
```diff
@@ -460,31 +425,17 @@ def get_free_space(self, disks=None, clear_part_type=None):
...
-                    # only check actual filesystems since lvm &c require a bunch of
-                    # operations to translate free filesystem space into free disk
-                    # space
-                    should_clear = self.should_clear(partition,
-                                                     clear_part_type=clear_part_type,
-                                                     clear_part_disks=[disk.name])
-                    if should_clear:
-                        disk_free += partition.size
-                    elif hasattr(partition.format, "free"):
+                for partition in [p for p in partitions if p.disk == disk]:
```
vpodzime: This should use parentheses not square brackets (i.e. a generator instead of list).

blivet/osinstall.py
```diff
@@ -1465,176 +1470,22 @@ def _update_custom_storage_ksdata(self):
...
 
+        # these will both be empty until our reset method gets called
+        # instantiate our own devicetree here to override the default created
+        # in Blivet so that protected_dev_specs gets handled
+        self.devicetree = DeviceTree(passphrase=self.encryption_passphrase,
+                                     luks_dict=self.__luks_devs,
+                                     ignored_disks=self.ignored_disks,
+                                     exclusive_disks=self.exclusive_disks,
+                                     disk_images=self.disk_images,
+                                     protected_dev_specs=self.config.protected_dev_specs)
```
vpodzime: This is not a cheap thing to do. I think ``Blivet`` should use the ``self.config.protected_dev_specs`` value too.

vpodzime: Related to this - I think the ``Blivet`` class should have ``self.conf`` too. More generic than the installer-specific one, but some config could be useful in general.

blivet/osinstall.py
snbueno: I'm not married to the name I originally gave it, so I can change it, sure. I'll just call it InstallerStorage.

dwlehman: I agree. I think I may have suggested the current name, but when I read the patches I found it confusing.

blivet/osinstall.py
dwlehman: We talked about moving the protected device handling out of `DeviceTree` completely, since all we do there is set a simple attribute (`device.protected`) anyway.

We also talked about refactoring such that `InstallerSorage.reset` include a call to `super().reset(cleanup_only=cleanup_only)` instead of duplicating that code.

As far as `Blivet` having its own configuration class/instance, I do like the idea of reducing clutter somewhat, but I'm not sure it's warranted just for ignored_disks, exclusive_disks, and disk_images.

dwlehman:
Please open a new pull request against `2.0-branch` so we can merge this instead of having to cherry-pick. You can close this one.

snbueno:
Will open a new one against 2.0-branch as requested.

blivet/osinstall.py
snbueno: Right; now that I will be able to test again, I'll do the protected device handling changes and push those with my next PR. Same with the super().reset call.

PR #350, opened 2016-03-31
Set both req_size and size of thin pool when growing LVM

vpodzime:
This fixes an issue with growing thin pool requests where the thin pools don't
have their size set but their req_size is set to the sum of their thin LVs.


dwlehman:
Please close this pull request and open one against 2.0-devel instead. Patch looks fine.

vpodzime:
> Please close this pull request and open one against 2.0-devel instead. Patch looks fine.

Sure thing, I wasn't sure if we are already using the new approach or not. Some announcement or note would be nice ;)

dwlehman:
Yeah, I am going to write up the branching and merging policy tomorrow morning. I know it's sort of willy-nilly right now.

PR #353, opened 2016-03-31
Set both req_size and size of thin pool when growing LVM

vpodzime:
This fixes an issue with growing thin pool requests where the thin pools don't
have their size set but their req_size is set to the sum of their thin LVs.


dwlehman:
Looks good to me. Is this needed for F24 (blivet-1.x) as well?

vpodzime:
> Looks good to me. Is this needed for F24 (blivet-1.x) as well?

No, the issue was introduced with the changes for the LVM RAID support which we don't have on the blivet-1.x branch.

PR #356, opened 2016-04-05
Fix traceback when writing dasd.conf (#1301589)

snbueno:
If using only FBA DASDs during installation, anaconda crashes when writing
dasd.conf (more specifically, when running the check to see if there are any
hyperPAV aliases). So, just be a bit safer and wrap the affected code in a
try/except block. No special handling needs to be done if OSError is thrown --
which happens if a user has only FBA DASDs -- so we can safely ignore it.

Resolves: rhbz#1031589

### DIFF
--- a/blivet/devicelibs/dasd.py
+++ b/blivet/devicelibs/dasd.py
@@ -295,6 +295,13 @@ def write_dasd_conf(disks, root):
 
     # check for hyper PAV aliases; they need to get added to dasd.conf as well
     sysfs = "/sys/bus/ccw/drivers/dasd-eckd"
+
+    # in the case that someone is installing with *only* FBA DASDs,the above
+    # sysfs path will not exist; so check for it and just bail out of here if
+    # that's the case
+    if not os.path.exists(sysfs):
+        return
+
     # this does catch every DASD, even non-aliases, but we're only going to be
     # checking for a very specific flag, so there won't be any duplicate entries
     # in dasd.conf


blivet/devicelibs/dasd.py
```diff
@@ -298,15 +298,18 @@ def write_dasd_conf(disks, root):
...
+                with open(aliasfile, "r") as falias:
+                    alias = falias.read().strip()
+
+                # if alias == 1, then the device is an alias; otherwise it is a
+                # normal dasd (alias == 0) and we can skip it, since it will have
+                # been added to dasd.conf in the above block of code
+                if alias == "1":
+                    f.write("%s\n" % d)
+    except OSError:
+        pass
```
vpodzime: I don't think this is the right thing to do here as it may hide real issues. Is the real problem here that the config file doesn't exist or what? If so, test the file existence before trying to append to it and add a comment why it is okay if it doesn't exist. Or is it the problem with some of the ``/alias`` files? If that's the case, there should again be a check before trying to open it with an explanation in a comment.

snbueno: The config file (/etc/dasd.conf)? No. It's the sysfs directory (/sys/bus/ccw/drivers/dasd-eckd) that doesn't exist and causes a traceback. This happens if you install only with FBA DASDs -- they do not use the ECKD driver, hence why the sysfs dir being checked doesn't exist (I believe it would be /sys/bus/ccw/drivers/dasd-fba for FBA DASDs).

FBA DASDs are pretty uncommon nowadays; most DASDs now are of the ECKD variety. If you install with a mixed group of DASDs (ECKD and FBA) this problem doesn't occur (since the existence of ECKD DASDs means /sys/bus/ccw/drivers/dasd-eckd will exist). And AFAIK if you have one of these HyperPAV aliases, it's not going to ever show up as an FBA disk.

I'll add a comment to explain this further (I didn't realize it was unclear what the issue is; oops :-/), and I can log the error for informational purposes, but I don't think any real issue is being avoided.

dwlehman: So the only line that could possibly generate a traceback is this one?

```python
devs = [d for d in os.listdir(sysfs) if d.startswith("0.0")]
```
If that's so, why is anything else wrapped in try/except? Why not just return early if `sysfs` does not exist instead of catching exceptions?

snbueno: That's what I decided, to just do os.path.exists() and return early if it doesn't. Sorry, did not get to push the updated PR until....well, now.

vpodzime:
Much better. :) Thanks for the change.

dwlehman:
You should probably update the commit message to take into account the new approach.

snbueno:
Updated commit msg.

dwlehman:
Looks good now.

PR #357, opened 2016-04-05
Make sure to add hyperPAV aliases to dasd.conf.

snbueno:
HyperPAV aliases enable improved I/O efficiency by allowing parallel
access to a real DASD through a base subchannel (device) and multiple
aliases.

Unfortunately, while the kernel recognizes such devices, they don't
seem to generate any actual udev events so that blivet picks them up;
nor do these aliases have a device node. Thus, if they are used at all
during installation, they are not brought online after rebooting, since
they are on the device ignore list. They are not removed from the device
ignore list since there is no entry in dasd.conf, so this patch loops
through all DASD devices in sysfs and adds all detected aliases to dasd.conf.

(HyperPAV support was added in rhel7-branch quite a while ago, but
I forgot to push it to any other branch as well.)

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -1406,6 +1406,31 @@ class Blivet(object, metaclass=SynchronizedMeta):
                 fields = [dasd.busid] + dasd.get_opts()
                 f.write("%s\n" % " ".join(fields),)
 
+        # check for hyper PAV aliases; they need to get added to dasd.conf as well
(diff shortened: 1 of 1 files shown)


vpodzime:
Commented in #356

vpodzime:
Looks good to me.

dwlehman:
Looks good to me now as well.

PR #358, opened 2016-04-06
Port LVM on RAID space calculations fixes (#1284660)

vpodzime:
These are all ported patches from the master branch fixing the bug rhbz#1284660 where our magical calculations of space in case of LVM on RAID lead to a failure of installation with a valid storage configuration in kickstart.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -1219,11 +1219,6 @@ class LVMFactory(DeviceFactory):
                    len(self.container.parents))
                 log.debug("size cut to %s to omit old device space", size)
 
-        if self.container_raid_level:
-            # add five extents per disk to account for md metadata
-            # (it was originally one per disk but that wasn't enough for raid5)
-            size += lvm.LVM_PE_SIZE * len(self.disks) * 5
-
         if self.container_encrypted:
             # Add space for LUKS metadata, each parent will be encrypted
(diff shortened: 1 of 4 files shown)


dwlehman:
I think an important test for this is to do an installation using anaconda that includes a new lvm-on-raid5 setup, since that was where we were seeing such excessive metadata size long ago.

vpodzime:
> I think an important test for this is to do an installation using anaconda that includes a new lvm-on-raid5 setup, since that was where we were seeing such excessive metadata size long ago.

Will do.

vpodzime:
> I think an important test for this is to do an installation using anaconda that includes a new lvm-on-raid5 setup, since that was where we were seeing such excessive metadata size long ago.
> 
> Will do.

Seems to be working great. I extended the kickstart file from the bug report to use RAID 5 instead of RAID 1 for the LVM part and adapted the size of the ``root`` LV and it works just fine. I also tried to add ``--grow`` to the ``root`` LV and the result is a successful installation with 0 free space in the PV. So it works as expected and very precisely (don't forget we still have the "emergency brake" in place which shrinks an LV if it doesn't fit in the available space; but it wasn't triggered).

vpodzime:
Custom partition with LVM on top of RAID5 works just fine too.

PR #359, opened 2016-04-06
Preserve traceback when re-raising exceptions

vpodzime:
When raising an exception like with 'raise SomeException(str(e))', the traceback
from the original exception is lost (just 'raise' to re-raise an exception is
okay). Python 3 has an extended raise statement 'raise SomeException from e'
which preserves the traceback. That's what we should use to make our lives
easier when debugging issues.


dwlehman:
Looks good to me.

PR #362, opened 2016-04-07
Update CONTRIBUTING file to match current practices.

dwlehman:
You can see a preview of what it will look like [here](https://gist.github.com/dwlehman/ceaf09d7ae1ac1c9e13da79285f5f738).

### DIFF
--- a/CONTRIBUTING
+++ b/CONTRIBUTING
@@ -1,282 +1,124 @@
-0. Thanks to the kernel patch submission guidelines, which were used as a
-   framework for creating this document.
+Contributor's Guide
+====================
 
-   https://www.kernel.org/doc/Documentation/SubmittingPatches
-
-I. Source
-=========
-
-A) Obtain a current source tree.
-If you do not have a repository with the current source handy, use git to obtain
-one.  All installer-related projects can be found on the RH Installer project
-page:
-
-  http://github.com/rhinstaller/
-
-In particular, the python-blivet repository can be found here:
-
-  https://github.com/rhinstaller/blivet
-
-Additional help for working within the codebase and testing changes can be found
-on the Fedoraproject wiki:
-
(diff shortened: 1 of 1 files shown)


CONTRIBUTING
```diff
@@ -1,282 +1,113 @@
...
+- if your work is on `x.y-devel`, merge it onto `x.{y+1}-devel`
+      ```
+      git checkout 2.1-devel
+      ```
+      ```
+      git merge --no-ff 2.0-devel
+      ```
+      Take a look to make sure things are as expected, then...
+      ```
+      git push
```
vpodzime: Does this actually work? Maybe I have some different config, but just ``git push`` pushes only the current branch to its upstream. What I have to do is something like this ``git push origin master 2.0-devel 2.1-devel 3.0-devel``.

vpodzime:
Apart from the above question I'd like to ask why this document uses the markdown syntax while we use the ReST syntax everywhere else (AFAICT).

CONTRIBUTING
dwlehman: That is an example of merging 2.0-devel to 2.1-devel. Maybe I should make that more clear.

dwlehman:
I used markdown because I forgot that github will render the ReST nicely as well when people click the link when creating a pull request. I'll convert it.

dwlehman:
I converted it to ReST and made it clear that that example was merging `2.0-devel` into `2.1-devel`.

I also added a note about release branches since it seems we're bringing them back.

dwlehman:
Also updated demo version [here](https://gist.github.com/dwlehman/ceaf09d7ae1ac1c9e13da79285f5f738) (same link as before).

vpodzime:
Looks good to me. And since it is now ReST, I think it would be nice to add a table of contents to the beginning of the file, perhaps with depth=1?

PR #363, opened 2016-04-07
Protect the live device's parent (#1172342)

bcl:
When an iso9660 is written to a USB the child partitions are excluded
from the devicetree by addUdevPartitionDevice. This means it never gets
to the point where it can protect the parent devices.

This adds a second check to see if the liveBackingDevice name starts
with any of the other device names. If so then the device is set as
protected and added to the protectedDevNames list.

### DIFF
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -747,6 +747,14 @@ class Populator(object):
                 for parent in device.parents:
                     parent.protected = True
 
+        # When an iso9660 is written to a USB the child partitions are not added to the
+        # devicetree (addUdevPartitionDevice skips them) so the block above will not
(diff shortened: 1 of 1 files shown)


vpodzime:
I'm okay with this for the *1.x* branch. However, for *2.0-devel* and above, I feel like it would be better to add the ``liveBackingDevice`` to the device tree and thus make the existing block of code effective.

bcl:
Yeah, this is only for 1.x, I haven't been able to get a working rawhide image yet to look at it there. If we're lucky it's already fixed.

bcl:
Moved to PR #364

PR #364, opened 2016-04-08
Protect the live device's parent (#1172342)

bcl:
When an iso9660 is written to a USB the child partitions are excluded
from the devicetree by addUdevPartitionDevice. This means it never gets
to the point where it can protect the parent devices.

This adds a second check to see if the liveBackingDevice name starts
with any of the other device names. If so then the device is set as
protected and added to the protectedDevNames list.

### DIFF
--- a/blivet/populator.py
+++ b/blivet/populator.py
@@ -429,6 +429,13 @@ class Populator(object):
                 lvm.lvm_cc_addFilterRejectRegexp(name)
                 return
 
+        # If this is the live device protect the whole disk
+        if self.liveBackingDevice == name and not disk.protected \
+           and disk.name not in self.protectedDevNames:
+            log.info("Protecting %s, parent of live device %s", disk.name, name)
+            disk.protected = True
+            self.protectedDevNames.append(disk.name)
+
         if not disk.partitioned:
             # Ignore partitions on:
             #  - devices we do not support partitioning of, like logical volumes


dwlehman:
What about moving the block that marks the devices protected into a method and calling that method from `addUdevPartitionDevice` (right before we check if the parent disk is partitioned) as well as from where the block is currently located?

bcl:
I don't see any point in doing it both places, I checked addUdevPartitionDevice and the parents haven't been setup yet so it doesn't really help.

dwlehman:
Oh, you'd have to pretty much duplicate the current code with minor modifications to additionally protect `disk` instead of the current device's parents.  Still, it wouldn't be vulnerable to goofy errors like the current version will see with lots of disks (eg: `sda`, `sdaa`, `sdab`, &c).

bcl:
Found a simpler solution, I think.

blivet/populator.py
```diff
@@ -429,6 +429,13 @@ def addUdevPartitionDevice(self, info, disk=None):
                 lvm.lvm_cc_addFilterRejectRegexp(name)
                 return
 
+        # If this is the live device protect the whole disk
+        if self.liveBackingDevice == name and not disk.protected \
+           and not disk.name in self.protectedDevNames:
```
dwlehman: I would prefer if you would switch from `not disk.name in` to `disk.name not in`, but otherwise this looks like a nice solution.

dwlehman:
It's probably worhwhile to port this version to `2.0-devel` so we don't lose it going forward. For `3.0-devel`, I think this code has moved into `blivet.osinstall.InstallerStorage.reset`.

PR #366, opened 2016-04-11
Correctly recognize internal RAID LVs

vpodzime:
If an internal LV is a RAID LV, it's lv_attr is really interesting. Usually
(bug in LVM?) it has 'r' as the first letter so we need to allow that and then
do some magic to correctly determine the role of the LV.

Using 'lvs -o+lv_role' would be much better for this, but I only learnt about it
today and using it would require more changes (incl. some in libblockdev).

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -958,6 +958,18 @@ class LVMInternalLVtype(Enum):
             else:
                 return cls.data
 
+        if lv_attr[0] == "r":
+            # internal LV which is at the same time a RAID LV
+            if lv_attr[6] == "C":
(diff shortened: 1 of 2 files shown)


vpodzime:
Same here with the CI as in https://github.com/rhinstaller/blivet/pull/365#issuecomment-208345471

dwlehman:
This looks okay for the short term, but if `lv_role` could clean some of that attribute parsing up it would be nice in the future.

vpodzime:
> This looks okay for the short term, but if lv_role could clean some of that attribute parsing up it would be nice in the future.

Sure, that's the ultimate goal.

PR #368, opened 2016-04-12
A simple script for multi-merges we now do

vpodzime:
Merging PRs into multiple branches is a tedious and error prone work. In order
to make our lives easier, here comes a simple script doing the multi-merge in a
way we want it to happen.

### DIFF
--- a/scripts/git-multi-merge
+++ b/scripts/git-multi-merge
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+if [ $# -lt 3 ]; then
+    echo "Usage: $0 ORIGIN BRANCH_TO_MERGE DEST_BRANCH1 [DEST_BRANCH2 [DEST_BRANCH3 [...]]]"
+    exit 0
+fi
+
+# terminate on any error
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Looks good to me.

dwlehman:
This makes me a bit nervous about people merging commits from their local branches unintentionally.

vpodzime:
> This makes me a bit nervous about people merging commits from their local branches unintentionally.

Do you think it should merge ``$origin/$src``?

dwlehman:
I guess that would alleviate my concern.

vpodzime:
Done.

vpodzime:
That didn't really work and the merge commits looked weird. So I added this check instead:
```
# checkout the $origin/$src branch to check if it exists and thus prevent
# merging local branches
git checkout $origin/$src
```
In combination with ``set -e`` this makes sure that it's not just some weird local branch. And ``git rebase $origin/$src`` below that makes sure the to-be-merged branch is up to date.

PR #369, opened 2016-04-12
Installer Isolation: part two (bootloader)

snbueno:
This moves a bunch of bootloader-related logic out of Blivet class and out of blivet.partitioning and into osinstall.InstallerStorage instead.

### DIFF
--- a/blivet/osinstall.py
+++ b/blivet/osinstall.py
@@ -34,7 +34,7 @@ from gi.repository import BlockDev as blockdev
 from pykickstart.constants import AUTOPART_TYPE_LVM, CLEARPART_TYPE_NONE, CLEARPART_TYPE_LINUX, CLEARPART_TYPE_ALL, CLEARPART_TYPE_LIST
 
 from . import util
-from . import get_sysroot, get_target_physical_root, error_handler, ERROR_RAISE
+from . import get_sysroot, get_target_physical_root, get_bootloader, error_handler, ERROR_RAISE
 
 from .blivet import Blivet
 from .storage_log import log_exception_info
@@ -1108,6 +1108,7 @@ class InstallerStorage(Blivet):
         """
         super().__init__(ksdata=ksdata)
 
+        self._bootloader = None
         self.config = StorageDiscoveryConfig()
         self.autopart_type = AUTOPART_TYPE_LVM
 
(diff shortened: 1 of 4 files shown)


blivet/osinstall.py
```diff
@@ -1297,6 +1297,37 @@ def get_fstype(self, mountpoint=None):
...
+        :param req_disks: a list of required disks for partitioning
+        :type req_disks: list
+        :returns: The list of sorted disks to be used for partitioning
+        :rtype: list
+        """
+        for disk in req_disks:
+            if self.boot_disk and disk == self.boot_disk:
+                boot_index = req_disks.index(disk)
+                req_disks.insert(0, req_disks.pop(boot_index))
+
```
vpodzime: Do not modify the list you are iterating over. I know this probably works in Python 3 somehow, but it's at least suspicious.

blivet/partitioning.py
```diff
@@ -564,17 +564,10 @@ def do_partitioning(storage):
...
-            part.req_size = part.req_base_size
 
-    try:
-        storage.boot_device.req_bootable = True
-    except AttributeError:
-        # there's no stage2 device. hopefully it's temporary.
-        pass
+    if hasattr(storage, "set_boot"):
+        # set the boot flag on boot device
+        storage.set_boot()
```
vpodzime: I'm sorry to say this, but I don't like these ``hasattr()`` calls. I have no great idea how to replace them, but maybe by defining ``set_bool()`` doing nothing in the ``Blivet`` class? Or maybe add ``cares_about_bootloader (prop)`` to the ``Blivet`` class, override it in the ``InstallerStorage`` and check that?

snbueno: I guess the property idea is better to me than anything else, so I can do that.

blivet/osinstall.py
```diff
@@ -1194,6 +1195,139 @@ def write(self):
...
+    def get_fstype(self, mountpoint=None):
+        """ Return the default filesystem type based on mountpoint. """
+        fstype = super().get_fstype(mountpoint=mountpoint)
+
+        if not mountpoint:
+            # just return the default
+            pass
+        elif mountpoint == "/boot":
+            fstype = self.default_boot_fstype
+
```
dwlehman: You could just skip the first if and do `if mountpoint == "/boot": ...`

snbueno: I wasn't actually sure if that check was over-kill or necessary, so I'll get rid of it.

blivet/partitioning.py
```diff
@@ -564,22 +566,25 @@ def do_partitioning(storage):
...
+            storage.boot_device.req_bootable = True
+        except AttributeError:
+            # there's no stage2 device. hopefully it's temporary.
+            pass
 
     remove_new_partitions(disks, partitions, partitions)
     free = get_free_regions(disks)
     try:
-        allocate_partitions(storage, disks, partitions, free)
+        allocate_partitions(storage, disks, partitions, free, boot_disk)
```
dwlehman: Please use the kwarg syntax here, eg: `boot_disk=boot_disk`.

blivet/partitioning.py
```diff
@@ -681,12 +688,14 @@ def allocate_partitions(storage, disks, partitions, freespace):
...
         req_disks.sort(key=storage.compare_disks_key)
-        for disk in req_disks:
-            if storage.boot_disk and disk == storage.boot_disk:
-                boot_index = req_disks.index(disk)
-                req_disks.insert(0, req_disks.pop(boot_index))
+
+        if boot_disk:
+            # make sure the boot disk is at the beginning of the disk list
+            for disk in req_disks[:]:
+                if storage.boot_disk and disk == storage.boot_disk:
```
dwlehman: You should probably catch `AttributeError` here like we do up in `do_partitioning` since `boot_disk` being not-`None` doesn't  guarantee that storage has a `boot_disk` attribute.

dwlehman:
Looks good to me except for a couple of minor complaints expressed against the diff.

blivet/partitioning.py
dwlehman: Oh -- you should be using `boot_disk` instead of `storage.boot_disk` here, right? If so there is no need to catch `AttributeError`.

snbueno: Yes, I should be using boot_disk there. I'll update the PR again once 'make check test' finishes running. Thank.

dwlehman:
Looks good to me.

PR #370, opened 2016-04-12
Some small, stupid bugfixes to first installer isolation patch set.

snbueno:
Well, this is what I get for not testing thoroughly. And what I at
first attributed to rawhide problems actually turned out to be....well,
my problems. Fortunately, not many though.

* osinstall.InstallerStorage.mountpoints needed the @property heading.
Oops.

* Apparently I forgot, in one of the manual rebases I had to do, to actually
move get_free_space (with all the clearpart type stuff, which anaconda needs) into osinstall.InstallerStorage. Mega-oops.

* And some smaller issue with protected devices, where I was just
passing the wrong variable into a function.

### DIFF
--- a/blivet/osinstall.py
+++ b/blivet/osinstall.py
@@ -1194,6 +1194,7 @@ class InstallerStorage(Blivet):
         fcoe.write(get_sysroot())
         zfcp.write(get_sysroot())
 
+    @property
     def mountpoints(self):
         return self.fsset.mountpoints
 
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

blivet/osinstall.py
```diff
@@ -1406,10 +1469,10 @@ def reset(self, cleanup_only=False):
     def _resolve_protected_device_specs(self):
         """ Resolve the protected device specs to device names. """
         for spec in self.config.protected_dev_specs:
-            name = self.devicetree.resolve_device(spec)
-            log.debug("protected device spec %s resolved to %s", spec, name)
-            if name:
-                self.protected_dev_names.append(name)
+            dev = self.devicetree.resolve_device(spec)
+            if hasattr(dev, "name"):
```
dwlehman: I think `if dev is not None:` would be better here.

dwlehman:
Looks good other than that one minor gripe.

blivet/osinstall.py
snbueno: Definitely can't disagree with that.

PR #372, opened 2016-04-14

clumens:

Resolves: rhbz#1270883

### DIFF
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -163,7 +163,7 @@ class Platform(object):
 
     def setPlatformBootPartition(self):
         """Return the default /boot partition for this platform."""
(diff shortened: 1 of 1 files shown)


dwlehman:
You'll want to open another pull request against 2.0-devel if this is also destined for rawhide/f25.

clumens:
Yeah, definitely.  I think we should talk about it for Fedora first, though.  They may have different plans there.

dwlehman:
I want to note that doing this on rhel7-branch without having decided on a course of action for newer branches introduces an opportunity for a regression if there is no follow-up.

clumens:
Yeah, good point.  I don't really want us to have to maintain this off on the side in rhel-whatever branches for all time, but I also don't really want to get into this in fedora-devel.  I could at least start something on anaconda-devel-list, though.

PR #375, opened 2016-04-18
Kickstart missing bootloader partitions (#1256249)

kellinm:

Resolves: rhbz#1256249
Resolves: rhbz#1242666


kellinm:
Merged two bugs, altering PR branch.

PR #376, opened 2016-04-19

clumens:

Resolves: rhbz#1270883


clumens:
This is the same as rhel7-branch, but for Fedora so we don't introduce regressions when it comes time for RHEL8.

dwlehman:
This pull request is against rhinstaller:master even though your branch appears to be based on 2.0-devel. You probably have to pay attention to the base/fork combo boxes when making the pull request.

PR #377, opened 2016-04-19

clumens:

Resolves: rhbz#1270883

### DIFF
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -165,7 +165,7 @@ class Platform(object):
 
     def set_platform_boot_partition(self):
         """Return the default /boot partition for this platform."""
-        return [PartSpec(mountpoint="/boot", size=Size("500MiB"),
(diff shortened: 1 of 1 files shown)


clumens:
Let's try that again.

dwlehman:
Looks okay to me.

@snbueno do you know why we have a block in platform.S390 that basically duplicates what the base class is doing? Is s390 going to continue with a partition /boot even when other platforms start using lvm for it?

snbueno:
@dwlehman, yeah, s390 has to keep a separate /boot partition. It can't exist on LVM unless it's on a single PV, because zipl (bootloader) can't read the initrd+kernel across multiple disks. It can theoretically exist on LVM if /boot is restricted to a singlePV, but when we tried that, some pretty big issues cropped up that resulted in us just ditching singlePV altogether (7df46eb8 in anaconda; 7de937b6, bfded4f7, and bae141e7 in blivet)

PR #378, opened 2016-04-19
Fix blivet constructor fs support check (#1242666)

kellinm:

Related: rhbz#1242666

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -62,15 +62,15 @@ def register_device_format(fmt_class):
     log.debug("registered device format class %s as %s", fmt_class.__name__,
               fmt_class._type)
 
-default_fstypes = ("ext4", "ext3", "ext2")
+default_fstypes = ("xfs", "ext4", "ext3", "ext2")
 
 
 def get_default_filesystem_type():
     for fstype in default_fstypes:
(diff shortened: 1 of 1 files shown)


vpodzime:
> Fixed the guard to check the boolean value of the property rather than
> checking for the existence of the pointer.

Please change the wording of this. There are no pointers of python and thus it's confusing.

vpodzime:
I don't understand what this has to do with the referenced bug.

vpodzime:
I also don't understand how this fixes the issue described above. Well, other than relying on the fact that *xfs* will probably be loaded. The tests should mock the functions that are needed to load the kernel modules so that they behave like everything was okay and the module was used.

dwlehman:
Leave the bug reference out of the first line since you have it as Related below.

Also, I'd leave out the long confusing explanation and just say something like this: the test for supported format types was broken because it was evaluating an unbound property instance as a boolean rather than the value the property would have if it were bound to an instance of `DeviceFormat`. In effect, it was doing `bool(id(get_device_format_class(fstype).supported))` where `get_device_format_class(fstype).supported` is a `property` instance.

dwlehman:
I think the reordering of the fstypes merits a separate commit.

PR #379, opened 2016-04-20
Add name of problematic device to UnusableConfigurationError

vojtechtrefny:
Having name of the problematic/unusable disk in this exception would help when trying to ignore the disk during (next) reset.


vpodzime:
Looks good to me.

PR #380, opened 2016-04-20
Ensure install-requires captures all packages

kellinm:
The install-requires and check-requires Makefile targets depend on
having the rpm-build package installed. Added warning that advises
user to install rpm-build. The script exits and does not attempt
installation using the incomplete requirements to avoid giving the
appearance that the Makefile target worked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ ZANATA_PUSH_ARGS = --srcdir ./po/ --push-type source --force
 
 MOCKCHROOT ?= fedora-rawhide-$(shell uname -m)
 
-TEST_DEPENDENCIES = $(shell rpm --specfile python-blivet.spec --requires | cut -d' ' -f1 | grep -v ^blivet)
+TEST_DEPENDENCIES = $(shell rpm --specfile python-blivet.spec --requires 2> /dev/null | cut -d' ' -f1 | grep -v ^blivet)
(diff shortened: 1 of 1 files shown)


dashea:
What do they require rpmdevtools for? The targets are just using rpm and dnf, so I would think they would work without rpmdevtools.

kellinm:
If you run make install-requires target; it calls rpm --specfile and grabs a list of required packages using grep and CLI tools. If rpmdevtools is not required, then this does not work and the user would miss installing those build dependencies.

kellinm:
The CI failure appears to be the patch carrying out the intended design. It's calling check-requires and sees that rpmdevtools is missing thus exits with error 1.

kellinm:
Changed rpmdevtools to rpm-build based on conversation in IRC.

Makefile
```diff
@@ -47,7 +47,13 @@ po-empty:
 po-fallback:
 	$(MAKE) po-pull || $(MAKE) po-empty
 
-check-requires:
+check-rpm-build:
+	@if rpm -q rpm-build | grep -q "not installed" ; then \
```
vpodzime: The exit code of ``rpm -q`` is enough here, the ``grep`` is not needed.

vpodzime:
Looks good to me otherwise.

vojtechtrefny:
#628 removes dependecy on rpm-build

PR #382, opened 2016-04-20
Kickstart missing bootloader partitions (#1242666)

kellinm:

Resolves: rhbz#1242666


dwlehman:
As @vpodzime said in #378 , you should probably mock `blivet.formats.fs.FS.load_module` in your test so you're not trying to load any modules. For this test you could even mock all of the default_fstype stuff since none of it is relevant.

kellinm:
@dwlehman - this patch already does it differently than the first iteration in that it mocks mountpoints and explictly uses prepboot format for the test.

The other patch is still needed because it bubbled up the known error we discussed, but this iteration of the code doesn't default_filesystem as it can run w/o the XFS patch.

PR #383, opened 2016-04-20
Obsolete python-blivet and blivet-data packages.

dwlehman:
This should make it possible for users to update rawhide
smoothly, unless they have something installed that still
requires the older python-blivet that isn't getting updated.


vpodzime:
Looks good to me. I think we don't have to care too much about people requiring the older python-blivet on rawhide.

PR #371, opened 2016-04-13
dbus proof of concept / initial interface

dwlehman:
Future pull requests will add `Device` and `Format` interfaces with corresponding sub-trees.

After that will come removal of devices, initialization of disks, and a `Factory` interface to provide some high-level configuration of new device stacks.

PR #384, opened 2016-04-20
dbus proof of concept / initial interface (2.2-devel)

dwlehman:
This is #371 rebased onto `2.2-devel`. While it may be useful to change the `DeviceFactory` API at some point, I think it is possible to have a fully-functioning dbus interface with the current API.


vpodzime:
Looks good to me.

PR #385, opened 2016-04-21
Compare sizes instead of numbers of sectors

vpodzime:
Instead of dividing two potentially big sizes by sector size and then comparing
the numbers of sectors we can compare the sizes and divide just the smaller
(more limitting) one. This is a minor optimization, but more importantly a nice
way to prevent issues with really big numbers that may not fit into 64bits (Size
divided by Size is a number).

Related: rhbz#1326897


dwlehman:
Looks okay to me.

PR #388, opened 2016-04-25
Add xfs to default filesystem types

kellinm:
XFS is the default filesystem type for several of the current linux
distributions. It should be included as a potential default filesystem
type in the constructor list.

Related: rhbz#1242666

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -62,7 +62,7 @@ def register_device_format(fmt_class):
     log.debug("registered device format class %s as %s", fmt_class.__name__,
               fmt_class._type)
 
-default_fstypes = ("ext4", "ext3", "ext2")
+default_fstypes = ("ext4", "xfs", "ext3", "ext2")
 
 
 def get_default_filesystem_type():


dwlehman:
I think `xfs` should only go to the front of the list on branches related to products whose default fs is `xfs`. On `2.0-devel` I think it should go right after `ext4`.

kellinm:
@dwlehman which product has ext4 as its default?

dwlehman:
@kellinm some of the fedora variants did last time I checked.

kellinm:
So I don't know that it matters then - if nothing else we have xfs as default some places, ext4 others; so having xfs and then all the ext together makes as much sense as anything else would at this point.

dwlehman:
You moved it to the front of the list on the basis that it's the default for rhel7-branch. I don't want it at the front for other branches. Outside the context of OS installation that would effectively make it the default for blivet, which I do not want to change now.

kellinm:
Lunch discussion with @dwlehman - pushed xfs to second spot for both RHEL and Fedora versions of this patch for two reasons:

- despite xfs being the default filesystem in RHEL, this would constitute a behavior change for blivet which is not desired
- if xfs is first in one branch, and second in the other, we'd have an internal inconsistency between the two branches and that is not desired either

PR #389, opened 2016-04-25
Add xfs to default filesystem types

kellinm:
XFS is the default filesystem type for several of the current linux
distributions. It should be included as a potential default filesystem
type in the constructor list.

Related: rhbz#1242666


dwlehman:
Looks good.

PR #392, opened 2016-04-26
Add a Device interface to the dbus service

dwlehman:
This adds a subtree `/com/redhat/Blivet1/Devices/` with individual devices' object paths being based on their `id`, eg: `/com/redhat/Blivet1/Devices/10`. As of now I don't see any need for various interfaces for device and format types, so all devices only implement `com.redhat.Blivet1.Device`.

I also added an `ObjectManager` interface at `/com/redhat/Blivet1` and moved the `Blivet` object to `/com/redhat/Blivet1/Blivet`. Other services use `Manager` for this, so I'm open to doing that instead if people think it would be better.

Things I already know I need to do:
- more unit tests
- class and module docstrings
- be a bit more stringent about checking return values, types, &c


dwlehman:
- fixed unit test for ListDevices
- slightly expanded commit messages for c8a8eb8 and 0caf942

examples/dbus_client.py
```diff
@@ -0,0 +1,13 @@
+
+import dbus
+
+bus = dbus.SystemBus()
+bus.add_match_string("type='signal',sender='com.redhat.Blivet1',path_namespace='/com/redhat/Blivet1'")
```
vpodzime: I don't quite understand what this is supposed to do. Maybe a comment would be nice here?

vpodzime:
Other than the nitpick above this looks good to me. Very nice :)

dwlehman:
Changed since last version:
- added some basic tests for `DBusObject` and `DBusDevice`
- added a few comments and docstrings (including the set-match in the client example)

dwlehman:
Jenkins, please test this.

dwlehman:
Jenkins, old chap, test this please.

blivet/dbus/device.py
```diff
@@ -0,0 +1,82 @@
...
+        props = {"Name": self._device.name,
+                 "Path": self._device.path,
+                 "Type": self._device.type,
+                 "Size": dbus.UInt64(self._device.size),
+                 "ID": self._device.id,
+                 "UUID": self._device.uuid or "",
+                 "Status": self._device.status or False,
+                 "RaidLevel": self._get_raid_level(),
+                 "Parents": dbus.Array((self.get_object_path_by_id(d.id) for d in self._device.parents), signature='s'),
+                 "Children": dbus.Array((self.get_object_path_by_id(d.id) for d in self._device.children), signature='s'),
```
vpodzime: Shouldn't this have a signature ``'ao'`` and thus the call ``signature='o'``? I have always found the ``'o'`` and ``'s'`` signatures confusing even though I have known the difference.

vpodzime:
Looks good to me otherwise.

blivet/dbus/device.py
dwlehman: You're correct. Nice catch. I forgot to update the signature when I changed the value.

PR #393, opened 2016-04-27
Fix root detection on btrfs in rescue mode

vojtechtrefny:
Mount btrfs volumes during rescue mode (to get list of its
subvolumes) and include non-leaf btrfs subvolumes when looking for
root devices and make sure it's mounted with subvol mount option.

Resolves: rhbz#1250011

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -3025,7 +3025,10 @@ def _findExistingInstallations(devicetree):
         util.makedirs(getTargetPhysicalRoot())
 
     roots = []
-    for device in devicetree.leaves:
+
+    direct_devices = (dev for dev in devicetree.devices if dev.direct)
+
+    for device in direct_devices:
         if not device.format.linuxNative or not device.format.mountable or \
            not device.controllable:
             continue
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -2106,7 +2106,7 @@ def mountExistingSystem(fsset, rootDevice,
         rootDevice.setup()
         rootDevice.format.mount(chroot=rootPath,
                                 mountpoint="/",
-                                options=readOnly)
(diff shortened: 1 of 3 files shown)


blivet/__init__.py
```diff
@@ -2106,7 +2106,7 @@ def mountExistingSystem(fsset, rootDevice,
         rootDevice.setup()
         rootDevice.format.mount(chroot=rootPath,
                                 mountpoint="/",
-                                options=readOnly)
+                                options="%s,%s" % (rootDevice.format.options, readOnly))
```
vpodzime: This should go into a separate commit.

vojtechtrefny: Makes sense, fixed.

dwlehman:
Should we be enabling `installer_mode` for rescue mode? Also, I think you could use the `direct` attribute instead of looking for leaves, &c.

vojtechtrefny:
> Should we be enabling `installer_mode` for rescue mode?

I'm not sure. `installer_mode` enables a lot things including running fsck on filesystems and enabling iscsi and fcoe and I don't know if we want this in rescue mode. I think this solution is safer for rhel7-branch.

> Also, I think you could use the `direct` attribute instead of looking for leaves, &c.

I didn't know we have such attribute :) I changed the code to use `direct` instead of `devicetree.leaves`.

blivet/__init__.py
```diff
@@ -3025,7 +3025,10 @@ def _findExistingInstallations(devicetree):
         util.makedirs(getTargetPhysicalRoot())
 
     roots = []
-    for device in devicetree.leaves:
+
+    direct_devices = [dev for dev in devicetree.devices if dev.direct]
```
vpodzime: This could be a generator.

vpodzime:
Looks good to me other than the nitpick above.

PR #394, opened 2016-04-28
Add a RemoveDevice method to DBusBlivet

dwlehman:
This is implemented via `DeviceTree.recursive_remove`, so I'm open to different names, including `RecursiveRemove`. Since #392 isn't merged yet its commits are appearing here, too. This one starts with 7b2ebfc and currently has six commits.


vpodzime:
Looks good to me.

PR #396, opened 2016-04-28
Add an InitializeDisk method to DBusBlivet

dwlehman:
This is just one commit on top of #392 and #394.

I closed #395 because I accidentally created it against `master` instead of `2.2-devel`.

### DIFF
--- a/blivet/dbus/blivet.py
+++ b/blivet/dbus/blivet.py
@@ -109,3 +109,10 @@ class DBusBlivet(DBusObject):
         """ Remove a device and all devices built on it. """
         device = self._dbus_devices[object_path]
         self._blivet.devicetree.recursive_remove(device)
+
+    @dbus.service.method(dbus_interface=BLIVET_INTERFACE, in_signature='o')
+    def InitializeDisk(self, object_path):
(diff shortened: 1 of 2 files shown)


blivet/dbus/blivet.py
```diff
@@ -100,9 +100,13 @@ def ResolveDevice(self, spec):
...
-        if device is not None:
-            dbus_device = next(d for d in self._dbus_devices if d._device == device)
-            object_path = dbus_device.object_path
+        if device is None:
+            raise dbus.exceptions.DBusException('%s.DeviceLookupFailed' % BUS_NAME,
+                                                'No device was found that matches the device '
+                                                'descriptor "%s".' % spec)
+
+        dbus_device = next(d for d in self._dbus_devices if d._device == device)
+        object_path = dbus_device.object_path
```
vpodzime: Either an exception or ``'/'`` as that's what other DBus APIs do in cases like this. On the other hand I think returning ``'/'`` would make more sense if the method was called like ``FindDevice()``. ``ResolveDevice()`` can consider the device not existing an exception.

vpodzime:
Also looks good to me.

PR #397, opened 2016-04-29
Break the cycle in LVMPhysicalVolume.destroy (#1331630)

vpodzime:
If pvremove() fails, we need to jump to DeviceFormat._destroy() not destroy()
because destroy calls self._destroy() which jumps to LVMPhysicalVolume.destroy()
again because 'self' is a LVMPhysicalVolume instance.

### DIFF
--- a/blivet/formats/lvmpv.py
+++ b/blivet/formats/lvmpv.py
@@ -128,7 +128,7 @@ class LVMPhysicalVolume(DeviceFormat):
         try:
             blockdev.lvm.pvremove(self.device)
         except blockdev.LVMError:
-            DeviceFormat.destroy(self, **kwargs)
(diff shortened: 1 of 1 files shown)


vpodzime:
We have the same issue also on all of the 2.0-devel...master branches.

dwlehman:
While you're changing stuff you could also use `super` instead of `DeviceFormat`, although it'll be the super-explicit wordy version for `1.20-devel`.

vpodzime:
> While you're changing stuff you could also use super instead of DeviceFormat, although it'll be the super-explicit wordy version for 1.20-devel.

I planned to do that on 2.0-devel+ branches. I don't really like the python2 version of the ``super()`` call.

dwlehman:
Works for me.

PR #398, opened 2016-04-29
Remove installer mode

snbueno:
This gets rid of flags.installer_mode from blivet. A few flags are created in its stead which are hopefully more generic, although I'm really bad at thinking of names. In other places local/surrounding logic could be used instead of just checking for flags.installer_mode.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -22,41 +22,11 @@
 
 __version__ = '2.0.2'
 
-##
-# Default stub values for installer-specific stuff that gets set up in
-# enable_installer_mode.  These constants are only for use inside this file.
-# For use in other blivet files, they must either be passed to the function
-# in question or care must be taken so they are imported only after
-# enable_installer_mode is called.
-##
-iutil = None
-ROOT_PATH = '/'
-_storage_root = ROOT_PATH
-_sysroot = ROOT_PATH
-short_product_name = 'blivet'
-ERROR_RAISE = 0
-
-
-class ErrorHandler(object):
-
-    def cb(self, exn):
-        # pylint: disable=unused-argument
-        return ERROR_RAISE
-
-error_handler = ErrorHandler()
-
-get_bootloader = lambda: None
-
-##
-# end installer stubs
-##
-
 import sys
 import importlib
 import warnings
 
-from . import util, arch, udev
-from .flags import flags
+from . import util, arch
 
 import logging
 log = logging.getLogger("blivet")
@@ -96,69 +66,6 @@ else:
 missing_plugs = _REQUESTED_PLUGIN_NAMES - avail_plugs
 for p in missing_plugs:
     log.info("Failed to load plugin %s", p)
-
-
-def enable_installer_mode():
-    """ Configure the module for use by anaconda (OS installer). """
-    global iutil
-    global ROOT_PATH
-    global _storage_root
-    global _sysroot
-    global short_product_name
-    global get_bootloader
-    global error_handler
-    global ERROR_RAISE
-
-    from pyanaconda import iutil  # pylint: disable=redefined-outer-name
-    from pyanaconda.constants import shortProductName as short_product_name  # pylint: disable=redefined-outer-name
-    from pyanaconda.bootloader import get_bootloader  # pylint: disable=redefined-outer-name
-    from pyanaconda.errors import errorHandler as error_handler  # pylint: disable=redefined-outer-name
-    from pyanaconda.errors import ERROR_RAISE  # pylint: disable=redefined-outer-name
-
-    if hasattr(iutil, 'getTargetPhysicalRoot'):
-        # For anaconda versions > 21.43
-        _storage_root = iutil.getTargetPhysicalRoot()  # pylint: disable=no-name-in-module
-        _sysroot = iutil.getSysroot()
-    else:
-        # For prior anaconda versions
-        from pyanaconda.constants import ROOT_PATH  # pylint: disable=redefined-outer-name,no-name-in-module
-        _storage_root = _sysroot = ROOT_PATH
-
-    from pyanaconda.anaconda_log import program_log_lock
-    util.program_log_lock = program_log_lock
-
-    udev.device_name_blacklist = [r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram']
-
-
-def get_sysroot():
-    """Returns the path to the target OS installation.
-
-    For traditional installations, this is the same as the physical
-    storage root.
-    """
-    return _sysroot
-
-
-def get_target_physical_root():
-    """Returns the path to the "physical" storage root.
-
-    This may be distinct from the sysroot, which could be a
-    chroot-type subdirectory of the physical root.  This is used for
-    example by all OSTree-based installations.
-    """
-    return _storage_root
-
-
-def set_sysroot(storage_root, sysroot=None):
-    """Change the OS root path.
-       :param storage_root: The root of physical storage
-       :param sysroot: An optional chroot subdirectory of storage_root
-    """
-    global _storage_root
-    global _sysroot
-    _storage_root = _sysroot = storage_root
-    if sysroot is not None:
-        _sysroot = sysroot
 
 
 class _LazyImportObject(object):
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -55,7 +55,7 @@ import sys
 import importlib
 import warnings
 
-from . import util, arch
+from . import util, arch, udev
 from .flags import flags
 
 import logging
@@ -127,7 +127,7 @@ def enable_installer_mode():
     from pyanaconda.anaconda_log import program_log_lock
     util.program_log_lock = program_log_lock
 
-    flags.installer_mode = True
+    udev.device_name_blacklist = [r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram']
 
 
 def get_sysroot():
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -199,7 +199,7 @@ class ActionList(object, metaclass=SynchronizedMeta):
 
         problematic = self._find_active_devices_on_action_disks(devices=devices)
         if problematic:
-            if flags.installer_mode:
+            if flags.auto_dev_updates:
                 for device in devices:
                     if device.protected:
                         continue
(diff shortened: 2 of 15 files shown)


blivet/flags.py
```diff
@@ -64,6 +64,10 @@ def __init__(self):
 
         self.multipath_friendly_names = True
 
+        # set to False since blivet doesn't need to account for any
+        # blacklisted devices by default
+        self.blacklisted_devs = False
```
vpodzime: I think this shouldn't be a boolean flag, but a configuration item holding the list of regexes like ``INSTALLER_BLACKLIST`` does now.

blivet/actionlist.py
```diff
@@ -199,7 +199,7 @@ def _pre_process(self, devices=None):
 
         problematic = self._find_active_devices_on_action_disks(devices=devices)
         if problematic:
-            if flags.installer_mode:
+            if flags.force_dev_update:
```
vpodzime: ``force_dev_updates`` would make more sense to me.

blivet/__init__.py
```diff
@@ -127,8 +127,6 @@ def enable_installer_mode():
     from pyanaconda.anaconda_log import program_log_lock
     util.program_log_lock = program_log_lock
 
-    flags.installer_mode = True
-
```
vpodzime: It would be nice if this whole function could be replaced by some ``Config`` object the attributes of which would be the things imported above set from the Anaconda and passed to ``Blivet()`` (or ``InstallerStorage()`` of course).

blivet/devices/btrfs.py
```diff
@@ -334,9 +333,9 @@ def _remove_subvolume(self, name):
 
     def list_subvolumes(self, snapshots_only=False):
         subvols = []
-        if flags.installer_mode:
+        if self.original_format.status:
```
vpodzime: I think this should use the new ``force_dev_update(s)`` flag.

blivet/flags.py
```diff
@@ -109,7 +108,6 @@ def update_from_boot_cmdline(self):
             self.noiswmd = True
 
     def update_from_anaconda_flags(self, anaconda_flags):
-        self.installer_mode = True
```
vpodzime: I think this whole function should go to Anaconda.

blivet/__init__.py
vpodzime: Oh, I now see the code is moved to ``osinstall`` in the end which is good enough for now, I think.

blivet/osinstall.py
```diff
@@ -1188,13 +1276,13 @@ def do_it(self, callbacks=None):
...
     def write(self):
-        Blivet.write(self)
+        super().write()
 
         self.make_mtab()
         self.fsset.write()
-        iscsi.write(get_sysroot(), self)
-        fcoe.write(get_sysroot())
-        zfcp.write(get_sysroot())
+        iscsi.write(self.sysroot, self)
```
vpodzime: Shouldn't this and all below be ``self._sysroot``?

vpodzime:
Looks good to me otherwise. I think we will need one more step of "good Anaconda riddance" in the future, though.

blivet/osinstall.py
snbueno: Eh, you technically can but in blivet.Blivet I defined the property "sysroot" (which is r/w). I'd say for the sake of being proper it should be self.sysroot.

blivet/flags.py
snbueno: I agree that yes, eventually it should. For now I will move it to blivet.osinstall with all the other installer-specific items temporarily living there.

blivet/devices/btrfs.py
snbueno: This was one place where I was uncertain whether the surrounding logic was really adequate to check, so yeah, I'm fine with using the flag instead.

snbueno:
Yep, that last step will take place when we're closer to releasing 3.0 + migrating Anaconda to use it.

blivet/flags.py
snbueno: Good idea; I've implemented that in my latest revision.

blivet/osinstall.py
vpodzime: oh, I didn't notice the new property. This is alright then.

blivet/flags.py
```diff
@@ -64,6 +64,10 @@ def __init__(self):
 
         self.multipath_friendly_names = True
 
+        # set to None since blivet doesn't need to account for any
+        # additional blacklisted devices by default
+        self.blacklisted_devs = None
```
vpodzime: An empty list would be a better default value here. And it would also allow you to remove the pylint-disable thing below.

blivet/populator/populator.py
```diff
@@ -453,7 +453,7 @@ def _populate(self):
 
         self.drop_lvm_cache()
 
-        if flags.installer_mode and not flags.image_install:
+        if flags.multipath_friendly_names and not flags.image_install:
```
vpodzime: Are you sure we never want to disable friendly names? The old behavior was to disable it in installer mode if requested now it would be left enabled.

blivet/udev.py
```diff
@@ -143,8 +140,8 @@ def __is_blacklisted_blockdev(dev_name):
     if dev_name.startswith("ram") or dev_name.startswith("fd"):
         return True
 
-    if flags.installer_mode:
-        if any(re.search(expr, dev_name) for expr in INSTALLER_BLACKLIST):
+    if flags.blacklisted_devs:
+        if any(re.search(expr, dev_name) for expr in flags.blacklisted_devs):  # pylint: disable=not-an-iterable
```
vpodzime: I still think this should be something like ``conf.blacklisted_devs``, having a list stored in something called ``flags`` is really counter-intuitive.

snbueno: Yeah; so dlehman suggested I just make udev.device_name_blacklist and set it to be an empty list by default; enable_installer_mode() can set it to the list of regexes. I think I prefer that , so that's what I've implemented.

blivet/populator/populator.py
snbueno: Yes, looking over it again, I can see how this might not be the best way to handle this. I don't really want to make a new flag just for that, so I suppose I can check flags.force_dev_updates

blivet/formats/luks.py
```diff
@@ -250,7 +250,7 @@ def _create(self, **kwargs):
     def _post_create(self, **kwargs):
         super(LUKS, self)._post_create(**kwargs)
         self.uuid = blockdev.crypto.luks_uuid(self.device)
-        if flags.installer_mode or not self.map_name:
+        if not self.map_name:
```
dwlehman: You're probably going to have to unset `self.map_name` in `LUKS._pre_create` as I think it might default to `luks-$(basename $device)`.

blivet/partitioning.py
```diff
@@ -350,7 +350,7 @@ def remove_new_partitions(disks, remove, all_partitions):
         # remove empty extended so it doesn't interfere
         extended = disk.format.extended_partition
         if extended and not disk.format.logical_partitions and \
-           (flags.installer_mode or
+           (flags.force_dev_updates or
```
dwlehman: Please make a separate flag for this -- maybe `keep_empty_extended_partition`. Note the sense would be the reverse of `installer_mode` above since we keep empty pre-existing extended partitions except when anaconda is driving.

blivet/blivet.py
```diff
@@ -147,7 +137,7 @@ def reset(self, cleanup_only=False):
         self.edd_dict = get_edd_dict(self.partitioned)
         self.devicetree.edd_dict = self.edd_dict
 
-        if not flags.installer_mode:
+        if not flags.include_nodev:
```
dwlehman: Should be `if flags.include_nodev:`.

blivet/osinstall.py
```diff
@@ -71,6 +151,50 @@ def copy_to_system(source):
...
+
+    blivet_flags.ibft = anaconda_flags.ibft
+    blivet_flags.dmraid = anaconda_flags.dmraid
+
+    # We don't want image installs writing backups of the *image* metadata
+    # into the *host's* /etc/lvm. This can get real messy on build systems.
+    if blivet_flags.image_install:
+        blivet_flags.lvm_metadata_backup = False
+
+    blivet_flags.blacklisted_devs = (r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram')
```
dwlehman: Is this staying or going? If it's staying it should be a `list` instead of a `tuple`.

dwlehman:
Other than those few minor comments this looks like a real improvement -- thanks.

blivet/partitioning.py
snbueno: Fixed, along with all the other things you mentioned.

blivet/osinstall.py
```diff
@@ -71,6 +151,50 @@ def copy_to_system(source):
...
+    blivet_flags.dmraid = anaconda_flags.dmraid
+
+    # We don't want image installs writing backups of the *image* metadata
+    # into the *host's* /etc/lvm. This can get real messy on build systems.
+    if blivet_flags.image_install:
+        blivet_flags.lvm_metadata_backup = False
+
+    blivet_flags.auto_dev_updates = True
+    blivet_flags.selinux_reset_fcon = True
+    blivet_flags.keep_empty_ext_partitions = False
```
dwlehman: Please move these flag settings that aren't conditional up into `enable_installer_mode`.

blivet/__init__.py
```diff
@@ -128,6 +128,7 @@ def enable_installer_mode():
     util.program_log_lock = program_log_lock
 
     flags.installer_mode = True
```
vpodzime: It's quite confusing to see the flag still being set here with what the commit message says. :) Would it be possible to remove the above line as part of this commit?

vpodzime:
Other than the nitpick above and the unconditional flags not being moved this looks good to me.

vpodzime:
And thanks for all the work on this, this is a great step forward.

blivet/__init__.py
snbueno: I totally missed that, so yes, it is possible. :)

blivet/osinstall.py
```diff
@@ -56,6 +77,125 @@
...
+
+    # We don't want image installs writing backups of the *image* metadata
+    # into the *host's* /etc/lvm. This can get real messy on build systems.
+    if flags.image_install:
+        flags.lvm_metadata_backup = False
+
+    flags.auto_dev_updates = True
+    flags.selinux_reset_fcon = True
+    flags.keep_empty_ext_partitions = False
+    flags.force_dev_update = True
```
dwlehman: What's this?

dwlehman:
Looks good to me.

PR #399, opened 2016-04-29
Size tests fix

vpodzime:
This should fix the recent failures of tests because of libbytesize's translations being finally in place.

### DIFF
--- a/blivet/size.py
+++ b/blivet/size.py
@@ -24,9 +24,9 @@ from bytesize import bytesize
 # we just need to make these objects available here
 # pylint: disable=unused-import
 from bytesize.bytesize import B, KiB, MiB, GiB, TiB, PiB, EiB, ZiB, YiB, KB, MB, GB, TB, PB, EB, ZB, YB
-from bytesize.bytesize import ROUND_UP, ROUND_DOWN
+from bytesize.bytesize import ROUND_UP, ROUND_DOWN, ROUND_HALF_UP
 
-ROUND_DEFAULT = ROUND_UP
+ROUND_DEFAULT = ROUND_HALF_UP
 
 
 def unit_str(unit, xlate=False):
@@ -146,6 +146,8 @@ class Size(bytesize.Size):
             :type rounding: one of ROUND_UP, ROUND_DOWN, or ROUND_DEFAULT
             :returns: Size rounded to nearest whole specified unit
             :rtype: :class:`Size`
+
+            .. warning:: Always think about the rounding mode and specify it!
 
             If size is Size(0), returns Size(0).
         """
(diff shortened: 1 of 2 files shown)


dwlehman:
Wow, how did I not notice that libbytesize doesn't use `ROUND_HALF_UP` (or even support it)?

My brain struggles mightily with the idea that `round_to_nearest('10.3')` would return `11` as opposed to `10` given that `10.3` is more than _twice_ as close to `10` as it is to `11`.

Since we are dealing with size quantities here I can see an argument for rounding up in certain situations, but it is still a mental hurdle for me. If the size represents a quantity used, rounding up seems to make sense (so that we do not under-estimate how much we have taken). But what if the size represents a quantity of free space? Rounding up seems to be a bad idea for this (because it leads us to over-estimate how much we have available). And what if we're relating two complementary sizes? `ROUND_HALF_UP` makes those kinds of relations sensible in general even when rounding.

Consider a `100 GiB` drive with a `44.2 GiB` partition and `55.8 GiB` of free space. If someone wanted to get a general feel for the situation, for example to display in a UI, here's what we'd see:

With `ROUND_UP`:
```
  Disk Size: 100 GiB
  Used: 45 GiB
  Free: 56 GiB
```

With `ROUND_HALF_UP`:
```
  Disk Size: 100 GiB
  Used: 44 GiB
  Free: 56 GiB
```

Which one do you think does a better job of representing the truth?

vpodzime:
Originally I didn't want to set any default because with storage, it's always a bug to rely on some default rounding. That's also why we always specify the rounding mode in *Blivet*.

As for the ``ROUND_HALF_UP`` I wasn't able to come up with an example of where it could be useful as using such value would always lead to issues. Even when showing to the user in the example above -- the user could tell themselves "Okay, I have 56 GiB of free space so I'm gonna create a 56GiB partition." which would fail. Thus the sizes should have been rounded up (used) and down (free) to prevent such issues.

So what I'm gonna do is:
* add ROUND_HALF_UP to rounding modes in *libbytesize*
* remove the default rounding mode in ``round_to_nearest()`` in *libbytesize*
* set the default rounding mode to ``ROUND_HALF_UP`` in ``round_to_nearest()`` in *Blivet* and add a warning to the docstring that relying on the default rounding mode is always a bug

vpodzime:
The tests are gonna fail here because they require the new libbytesize. I'll build and install it on our jenkins slaves once https://github.com/rhinstaller/libbytesize/pull/15 gets merged.

dwlehman:
I would be okay with making rounding mode a positional/mandatory argument for `blivet-3.0`.

vpodzime:
> I would be okay with making rounding mode a positional/mandatory argument for blivet-3.0.

I had the very same idea.

vpodzime:
Jenkins, please test this.

vpodzime:
Jenkins, please test this again.

vpodzime:
Ran the check manually on the Jenkins slave and all tests passed.

PR #402, opened 2016-05-02
Some low-hanging fruit to speed reset() up

vpodzime:
Three small and simple changes that make ``reset()`` quite a bit faster (~3s on my testing VM). A lot more can/needs to be done in the future, but the rest is quite complicated.

Just a note for the second patch:
``resolve_devspec()`` is not needed there. Something like ``resolve_dev_symlink()`` would be sufficient and much more efficient, but since it is a common operation, I'd like to add such function to *libblockdev* and then do the change in *Blivet* to start using it.

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -50,7 +50,6 @@ from ..tasks import fsminsize
 import logging
 log = logging.getLogger("blivet")
 
-
 device_formats = {}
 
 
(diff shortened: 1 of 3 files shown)


vpodzime:
The tests failure is the same old unrelated issue with the ``Size`` class. Everything else passed just fine.

vpodzime:
Jenkins, please test this.

snbueno:
These all look ok to me.

vpodzime:
Jenkins, please test this.

PR #404, opened 2016-05-02
LUKS data moved to singleton

japokorn:
- added "static_data" directory
- the following was moved to luks_data singleton:
    - luks_devs
    - encryption_passphrase
    - luks_min_entropy
    - save_passphrase
    - min_luks_entropy attribute

### DIFF
--- a/blivet/autopart.py
+++ b/blivet/autopart.py
@@ -31,6 +31,7 @@ from .errors import NoDisksError, NotEnoughFreeSpaceError
 from .formats import get_format
 from .partitioning import do_partitioning, get_free_regions, grow_lvm
 from .i18n import _
+from .static_data import luks_data
 
 from pykickstart.constants import AUTOPART_TYPE_BTRFS, AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP, AUTOPART_TYPE_PLAIN
 
@@ -136,7 +137,7 @@ def _get_candidate_disks(storage):
     return disks
 
 
-def _schedule_implicit_partitions(storage, disks, min_luks_entropy=0):
+def _schedule_implicit_partitions(storage, disks):
     """ Schedule creation of a lvm/btrfs member partitions for autopart.
 
         We create one such partition on each disk. They are not allocated until
@@ -146,9 +147,6 @@ def _schedule_implicit_partitions(storage, disks, min_luks_entropy=0):
         :type storage: :class:`~.Blivet`
         :param disks: list of partitioned disks with free space
         :type disks: list of :class:`~.devices.StorageDevice`
-        :param min_luks_entropy: minimum entropy in bits required for
-                                 luks format creation
-        :type min_luks_entropy: int
         :returns: list of newly created (unallocated) partitions
         :rtype: list of :class:`~.devices.PartitionDevice`
     """
@@ -162,11 +160,11 @@ def _schedule_implicit_partitions(storage, disks, min_luks_entropy=0):
     for disk in disks:
         if storage.encrypted_autopart:
             fmt_type = "luks"
-            fmt_args = {"passphrase": storage.encryption_passphrase,
+            fmt_args = {"passphrase": luks_data.encryption_passphrase,
                         "cipher": storage.encryption_cipher,
                         "escrow_cert": storage.autopart_escrow_cert,
                         "add_backup_passphrase": storage.autopart_add_backup_passphrase,
-                        "min_luks_entropy": min_luks_entropy}
+                        "min_luks_entropy": luks_data.min_entropy}
         else:
             if storage.autopart_type in (AUTOPART_TYPE_LVM, AUTOPART_TYPE_LVM_THINP):
                 fmt_type = "lvmpv"
@@ -183,7 +181,7 @@ def _schedule_implicit_partitions(storage, disks, min_luks_entropy=0):
     return devs
 
 
-def _schedule_partitions(storage, disks, implicit_devices, min_luks_entropy=0, requests=None):
+def _schedule_partitions(storage, disks, implicit_devices, requests=None):
     """ Schedule creation of autopart/reqpart partitions.
 
         This only schedules the requests for actual partitions.
@@ -192,9 +190,6 @@ def _schedule_partitions(storage, disks, implicit_devices, min_luks_entropy=0, r
         :type storage: :class:`~.Blivet`
         :param disks: list of partitioned disks with free space
         :type disks: list of :class:`~.devices.StorageDevice`
-        :param min_luks_entropy: minimum entropy in bits required for
-                                 luks format creation
-        :type min_luks_entropy: int
         :param requests: list of partitioning requests to operate on,
                          or `~.storage.autopart_requests` by default
         :type requests: list of :class:`~.partspec.PartSpec` instances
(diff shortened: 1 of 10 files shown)


blivet/autopart.py
```diff
@@ -494,15 +486,17 @@ def do_autopart(storage, data, min_luks_entropy=0):
...
+    devs = _schedule_implicit_partitions(storage, disks)
     log.debug("candidate disks: %s", disks)
     log.debug("devs: %s", devs)
 
     if disks == []:
         raise NotEnoughFreeSpaceError(_("Not enough free space on disks for "
                                         "automatic partitioning"))
 
-    devs = _schedule_partitions(storage, disks, devs, min_luks_entropy=min_luks_entropy)
+    luks_data.min_entropy = min_luks_entropy
```
dwlehman: I don't think `min_luks_entropy` is defined here.

blivet/static_data/luks_data.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class LUKS_Data(object):
+    """ Class to be used as a singleton.
+        Maintains the LUKS data.
+    """
+
+    def __init__(self):
+        """
+            TODO - fix doc text
```
vpodzime: Would be nice to DO the TODO :)

blivet/static_data/luks_data.py
```diff
@@ -0,0 +1,91 @@
...
+    def save_passphrase(self, device):
+        """ Save a device's LUKS passphrase in case of reset. """
+        passphrase = device.format._LUKS__passphrase
+        if passphrase:
+            luks_data.devs[device.format.uuid] = passphrase
+            self.add_passphrase(passphrase)
+
+    @property
+    def passphrases(self):
+        return self. __passphrases
```
dwlehman: Is there a space between `self.` and `__passphrases` here?

blivet/blivet.py
```diff
@@ -145,8 +144,8 @@ def __init__(self, ksdata=None):
 
         # these will both be empty until our reset method gets called
         self.devicetree = DeviceTree(conf=self.config,
-                                     passphrase=self.encryption_passphrase,
-                                     luks_dict=self.__luks_devs)
+                                     passphrase=luks_data.encryption_passphrase,
+                                     luks_dict={})
```
vpodzime: I think the ``luks_dict`` parameter should default to ``None`` with ``DeviceTree`` creating the empty dict in such cases.

blivet/devicefactory.py
```diff
@@ -336,7 +337,7 @@ def __init__(self, storage, size, disks, fstype=None, mountpoint=None,
 
         self.child_factory = None
         self.parent_factory = None
-        self.min_luks_entropy = min_luks_entropy
+        luks_data.min_entropy = min_luks_entropy
```
vpodzime: I'm not sure this should be set globally. It makes some sense to specify minimal entropy required per device. It makes sense in ``_schedule_partitions()`` and autopart in general, but for "explicitly created" devices it should be possible to specify minimal entropy separately.

vpodzime: And maybe store the value for future devices if it is not set explicitly again then.

blivet/devicefactory.py
```diff
@@ -661,7 +662,7 @@ def _create_device(self):
 
             fmt = get_format(self.fstype,
                              mountpoint=self.mountpoint,
-                             min_luks_entropy=self.min_luks_entropy,
+                             min_luks_entropy=luks_data.min_entropy,
```
vpodzime: If it is, however, global, it should be passed as as an argument.

blivet/populator/populator.py
```diff
@@ -73,6 +73,9 @@ def __init__(self, conf=None, passphrase=None, luks_dict=None):
             :type luks_dict: dict
 
         """
+
+        luks_data.luks_dict = luks_dict
```
vpodzime: This should be ``luks_data.devs`` I guess?

blivet/static_data/luks_data.py
```diff
@@ -0,0 +1,91 @@
...
+        self.__min_entropy = 0
+        self.__passphrases = []
+
+    @property
+    def devs(self):
+        return self.__devs
+
+    @devs.setter
+    def devs(self, value):
+        self.__devs = value
```
vpodzime: I think ``devs`` is not a good name for this. Maybe it doesn't have to be publicly available like this at all? I believe all that's needed is a method that would return a passphrase for a given device.

japokorn:
Uploaded new and generally more awesome version. As per comments:
@dwlehman:
 * `autopart.py`: `min_luks_entropy` is in that case defined as an optional method parameter
 * `luks_data.py`: Rogue space removed
@vpodzime:
 * `luks_data.py`: TODO removed, added comments
 * `blivet.py`, `devicefactory.py`: 
     * Kept to support backwards-compatibility
     * Will be changed/removed in blivet 3.0

dwlehman:
I see now that you removed the docs for `min_luks_entropy` but not the actual argument handling. Why not use the global value like you're doing elsewhere, or at least use that as a default value for the kwarg?

blivet/autopart.py
```diff
@@ -450,16 +445,13 @@ def do_reqpart(storage, requests):
...
+def do_autopart(storage, data, min_luks_entropy=None):
     """ Perform automatic partitioning.
 
         :param storage: a :class:`~.Blivet` instance
         :type storage: :class:`~.Blivet`
         :param data: kickstart data
         :type data: :class:`pykickstart.BaseHandler`
-        :param min_luks_entropy: minimum entropy in bits required for
-                                 luks format creation
-        :type min_luks_entropy: int
```
vpodzime: the docstring should stay just being updated with what ``None`` means

blivet/autopart.py
```diff
@@ -494,15 +486,17 @@ def do_autopart(storage, data, min_luks_entropy=0):
...
-    devs = _schedule_implicit_partitions(storage, disks, min_luks_entropy)
+    devs = _schedule_implicit_partitions(storage, disks)
     log.debug("candidate disks: %s", disks)
     log.debug("devs: %s", devs)
 
     if disks == []:
         raise NotEnoughFreeSpaceError(_("Not enough free space on disks for "
                                         "automatic partitioning"))
+    if min_luks_entropy is not None:
+        luks_data.min_entropy = min_luks_entropy
```
vpodzime: I think this needs to happen before ``_schedule_implicit_partitions()`` is called.

blivet/deviceaction.py
```diff
@@ -581,7 +582,7 @@ def execute(self, callbacks=None):
 
         if isinstance(self.device.format, luks.LUKS):
             # LUKS needs to wait for random data entropy if it is too low
-            min_required_entropy = self.device.format.min_luks_entropy
+            min_required_entropy = luks_data.min_entropy
```
vpodzime: I think this needs to check if ``format.min_luks_entropy`` is set or not and use the global value only if not.

blivet/formats/luks.py
```diff
@@ -245,7 +238,7 @@ def _create(self, **kwargs):
                                     key_file=self._key_file,
                                     cipher=self.cipher,
                                     key_size=self.key_size,
-                                    min_entropy=self.min_luks_entropy)
+                                    min_entropy=luks_data.min_entropy)
```
vpodzime: This all needs to stay as is if we want to support per-device (more precisely per-format) entropy threshold.

blivet/autopart.py
japokorn: Fixed.

blivet/autopart.py
japokorn: Fixed.

blivet/deviceaction.py
japokorn: Fixed.

blivet/formats/luks.py
japokorn: Fixed.

blivet/autopart.py
vpodzime: Thanks, but it should go next to the other parameters not to the block of text below. ;)

vpodzime:
Looks good to me other than the nitpick above.

PR #406, opened 2016-05-03
Retry umount and log output of lsof if it fails

bcl:
This is a copy of how lorax handles umount. Try 3 times, sleep 1s
between retries and log the output of lsof if the logging level is debug
or higher.

_This is related to rhbz#1315541 trying to see if retrying the umount helps_

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -6,12 +6,14 @@ import os
 import shutil
 import selinux
 import subprocess
+from subprocess import CalledProcessError
 import re
 import sys
 import tempfile
 import uuid
 import hashlib
 import warnings
+from time import sleep
 from decimal import Decimal
 from contextlib import contextmanager
 from functools import wraps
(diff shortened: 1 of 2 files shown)


blivet/util.py
```diff
@@ -116,12 +118,39 @@ def mount(device, mountpoint, fstype, options=None):
...
-    try:
-        rc = run_program(["umount", mountpoint])
-    except OSError:
-        raise
+def umount(mountpoint, lazy=False, maxretry=3, retrysleep=1.0):
+    """ Unmount the given mountpoint.
 
+        :param str mountpoint: Path to unmount
+        :param bool lazy: When True do a lazy (-l) umount
+        :param int maxretry: Number of times to try the umount
```
vpodzime: According to the logic of the code below, isn't this more ``maxtry`` than ``maxretry``?

blivet/util.py
```diff
@@ -116,12 +118,39 @@ def mount(device, mountpoint, fstype, options=None):
...
+        :param float retrysleep: Seconds to sleep between retries
+        :returns: rc of last umount attempt
+        :rtype: int
+        :raises CalledProcessError: if all the retries fail
+        :raises OSError: if command cannot be executed
+    """
+    cmd = ["umount"]
+    if lazy: cmd += ["-l"]
+    cmd += [mountpoint]
```
vpodzime: Any reason why the two lines above don't just call ``.append()``?

blivet/util.py
```diff
@@ -116,12 +118,39 @@ def mount(device, mountpoint, fstype, options=None):
...
+    count = 0
+    while maxretry > 0:
```
vpodzime: What is this condition good for? If ``maxretry`` (or ``maxtry``) is ``0``, the function can just return early.

bcl:
No particular reasons, the code was copied from lorax.

dwlehman:
Looks okay, I suppose.

PR #407, opened 2016-05-04
Ignore pylint being confused with our mixins' properties

vpodzime:
Pylint somehow fails to figure out that 'parent_lv' is a property of the mixin
class and reports it to be a method which doesn't have any 'size' member.


dwlehman:
Looks good, provided you add another space before that comment to satisfy `python3-pep8`.

PR #409, opened 2016-05-06
Add the P_ keyword to xgettext.

dashea:
The plural string, "RAID level %(raidLevel)s requires that device have
at least %(minMembers)d member(s)." was not being included for
translation.

This also affects 1.20-devel but I don't know how much you want to care about Fedora string freezes.


vpodzime:
Looks good to me.

dashea:
I hit the big green button to push to 2.0-devel, but it looks like 2.1/2.2/3.0/master are more behind 2.0 than I'm comfortable with being responsible for, so I'll let someone else merge what needs merging.

PR #410, opened 2016-05-06
Add an Action interface to the dbus service.

dwlehman:
This adds interface `com.redhat.Blivet1.Action`.

It also changes things a bit: When an action is added that causes a device or format to be removed from the `DeviceTree` , that device (or format) is moved to a separate subtree for things whose destruction is pending. For example, if `sda1` had object path `/com/redhat/Blivet1/Devices/4` and we scheduled an action to destroy it, that object would be moved to `/com/redhat/Blivet1/RemovedDevices/4`. Once that action is executed and the device is really gone, we will also remove it from the dbus interface.

This also adds a `Commit` method to `DBusBlivet`, which might come in handy at some point ;)


tests/dbus_test.py
```diff
@@ -92,39 +94,41 @@ def setUp(self, *args):
         self._device_id = random.randint(0, 500)
         self._format_id = random.randint(501, 1000)
         self.obj = DBusDevice(Mock(name="StorageDevice", id=self._device_id,
-                                   parents=[], children=[], format=Mock(id=self._format_id)))
+                                   parents=[], children=[]),
+                              Mock(name="OjectManager"))
```
vpodzime: typo?

blivet/dbus/blivet.py
```diff
@@ -32,21 +31,24 @@
 from .object import DBusObject
 
 
+def sorted_object_paths_from_list(obj_list):
+    devices = sorted(obj_list, key=lambda d: d.object_path)
+    devices.sort(key=lambda d: d.id)
+    return [d.object_path for d in devices]
```
vpodzime: I think this shouldn't refer to "devices" if it's supposed to be a generic function. If not, it should be named like ``sorted_devices_from_list``.

blivet/dbus/blivet.py
```diff
@@ -76,42 +78,35 @@ def properties(self):
...
+        added = DBusDevice(device, self._manager)
         self._manager.add_object(added)
 
     def _format_removed(self, fmt):
-        removed_object_path = DBusFormat.get_object_path_by_id(fmt.id)
-        removed = self._dbus_formats[removed_object_path]
+        removed = self._manager.get_object_by_id(fmt.id)
         self._manager.remove_object(removed)
-        del self._dbus_formats[removed_object_path]
+        removed.remove_from_connection()
```
vpodzime: This is defined in the next commit.

blivet/dbus/object.py
```diff
@@ -39,6 +40,22 @@ def _init_dbus_object(self):
...
+    def present(self):
+        """ Is this object present in blivet's current view? """
+        return self._present
+
+    def set_presence(self, state):
+        """ Indicate whether the object is in blivet's current view. """
+        conn = self.connection
+        self.remove_from_connection()
+        self._present = state
+        self.add_to_connection(conn, self.object_path)
```
vpodzime: Any reason why this is not a setter for the ``present`` property?

blivet/dbus/blivet.py
vpodzime: Oh, is it a method of the ``dbus.Object`` class?

blivet/dbus/blivet.py
```diff
@@ -208,3 +208,8 @@ def InitializeDisk(self, object_path):
         self.RemoveDevice(object_path)
         device = self._get_device_by_object_path(object_path)
         self._blivet.initialize_disk(device)
+
+    @dbus.service.method(dbus_interface=BLIVET_INTERFACE)
+    def Commit(self):
+        """ Commit pending changes to disk. """
+        self._blivet.do_it()
```
vpodzime: I know this is a bit preliminary, but shouldn't we catch exceptions here and transform them into DBus exceptions?

blivet/dbus/object.py
dwlehman: Yes, I initially hesitated because I'm tired of properties that do too much in the setter/getter but I don't think one really does too much now that it's done.

blivet/dbus/blivet.py
dwlehman: Yes, I need to do this throughout the service.

blivet/dbus/blivet.py
dwlehman: True.

dwlehman:
Last push should address all of your comments.

vpodzime:
Looks good to me now.

PR #414, opened 2016-05-09
Improve documentation of the udev.resolve_glob() function

vpodzime:
Otherwise people may be confused what it actually does.


snbueno:
Looks ok to me.

PR #416, opened 2016-05-11
Fix protected status for extended partitions

vojtechtrefny:
Extended partitions should be protected also when at least one of
its logical partitions is protected.

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -660,6 +660,20 @@ class PartitionDevice(StorageDevice):
         self.disk.format.commit()
         self.update_size()
 
+    @property
+    def protected(self):
+        protected = super().protected
+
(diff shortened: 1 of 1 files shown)


blivet/devices/partition.py
```diff
@@ -660,6 +660,20 @@ def resize(self):
         self.disk.format.commit()
         self.update_size()
 
+    @property
+    def protected(self):
+        protected = super().protected
+
+        # extended partition is protected also when one of its logical partitions is protected
+        if self.is_extended:
+            return any(part.protected for part in self.disk.children if part.is_logical) or protected
```
vpodzime: ``protected`` should be checked first here because if it's ``True`` we would avoid the more expensive check of all disk partitions.

vpodzime:
Looks good to me otherwise.

PR #418, opened 2016-05-12
Ignore errors activating unknown swap partitions (#1330763)

bcl:
When installing alongside other systems (eg. Debian) their swap may not
be usable so ignore any UnknownSwapErrors that happen while activating
swaps.

Resolves: rhbz#1330763

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -71,6 +71,7 @@ from pykickstart.constants import AUTOPART_TYPE_LVM, CLEARPART_TYPE_ALL, CLEARPA
 
 from .storage_log import log_exception_info, log_method_call
 from .errors import DeviceError, DirtyFSError, FSResizeError, FSTabTypeMismatchError, UnknownSourceDeviceError, StorageError, UnrecognizedFSTabEntryError
+from .errors import UnknownSwapError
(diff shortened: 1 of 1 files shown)


blivet/__init__.py
```diff
@@ -2575,6 +2576,9 @@ def turnOnSwap(self, rootPath=""):
                 try:
                     device.setup()
                     device.format.setup()
+                except UnknownSwapError as e:
+                    log.warn("Failed to activate swap on %s, skipping it.", device.path)
+                    break
```
vpodzime: Wow, this is one of the weirdest ``while`` cycles I've ever seen. But that's not a problem of this change. :)

vpodzime:
Looks good to me.

dwlehman:
The change looks good. Is there a corresponding pull request for `2.0-devel`?

bcl:
> The change looks good. Is there a corresponding pull request for 2.0-devel?

Not yet, I'm waiting on feedback from the reporter. I don't think it does us much good if it's going to crash after rebooting.

bcl:
Pushed to rhel7-branch, v2.0 is going to need more work, looks like libblockdev doesn't make a distinction between the types of swap failures.

PR #419, opened 2016-05-15
ARM platforms: support both msdos and gpt partitions

nullr0ute:
More details in the commit message but we need to be able to support both msdos and GPT partitions due to various low end Single Board Computers such as the Pine64. It would be really good to have this in Fedora 24 too, I think I have got the spots needed to support msdos and GPT on both ARMv7 and aarch64, I couldn't see any spots in anaconda itself.


nullr0ute:
Did this completely wrong, sorry

PR #420, opened 2016-05-15
ARM platforms: support both msdos and gpt partitions

nullr0ute:
We need to be able to support both msdos and GPT partitions due to various low end Single Board Computers such as the Pine64. It would be really good to have this in Fedora 24 too, I think I have got the spots needed to support msdos and GPT on both ARMv7 and aarch64, I couldn't see any spots in anaconda itself.

A lot of new aarch64 single board computer devices need old style
msdos partition types due to restrictions in booting where the SoC
boots off a u-boot, that supports uEFI booting, written to the area
that is occupied by a GPT partition table.

Similarly a number of ARMv7 devices support booting from GPT
partitions where the u-boot binary is available in a EEPROM or nand.

### DIFF
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -274,6 +274,7 @@ class MacEFI(EFI):
 
 class Aarch64EFI(EFI):
     _non_linux_format_types = ["vfat", "ntfs"]
+    _disklabel_types = ["msdos", "gpt"]
 
 
 class PPC(Platform):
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me.

nullr0ute:
Do you want a separate PR for the F-24/1.20 branch?

vpodzime:
> Do you want a separate PR for the F-24/1.20 branch?

That would be nice. And actually, this should be a PR for the 2.0-devel branch so I'm closing it now. I'll take care about porting the 1.20-devel version to 2.0-devel and all the way upwards.

PR #422, opened 2016-05-16
Add Factory method to DBusBlivet

dwlehman:


examples/dbus_client.py
```diff
@@ -15,4 +15,5 @@
 objects = object_manager.GetManagedObjects()
 for object_path in blivet.ListDevices():
     device = objects[object_path]['com.redhat.Blivet1.Device']
-    print(device['Name'], device['Type'], device['Size'], device['FormatType'])
+    fmt = objects[device['Format']]['com.redhat.Blivet1.Format']
+    print(device['Name'], device['Type'], device['Size'], fmt['Type'])
```
vpodzime: Would be nice to show how to get the format object and some info from it, but that can be added later.

blivet/dbus/blivet.py
```diff
@@ -218,3 +226,27 @@ def Commit(self):
             raise dbus.exceptions.DBusException('%s.%s' % (BUS_NAME, e.__class__.__name__),
                                                 "An error occured while committing the "
                                                 "changes to disk: %s" % str(e))
+
+    @dbus.service.method(dbus_interface=BLIVET_INTERFACE, in_signature='ita{sv}', out_signature='o')
+    def Factory(self, device_type, size, kwargs):
```
vpodzime: I still haven't decided if sizes should be passed as ``uint64``s or strings over DBus. Using ``uint64`` is common, but we may as well be the first ones doing "the step to the future".

blivet/devices/storage.py
```diff
@@ -710,11 +710,13 @@ def _set_format(self, fmt):
             elif fmt.min_size and fmt.min_size > self.size:
                 raise errors.DeviceError("device is too small for new format")
 
-        callbacks.format_removed(device=self, fmt=self._format)
-        self._format = fmt
-        self._format.device = self.path
-        self._update_netdev_mount_option()
-        callbacks.format_added(device=self, fmt=self._format)
+        change = self._format != fmt
+        if change:
```
vpodzime: Any reason for the extra variable here? Looks to me like ``if self._format != format`` would do and be clear enough on its own.

vpodzime:
Looks good to me other than the nitpicks above.

blivet/devices/storage.py
dwlehman: Originally I made the variable because I was only making the callbacks conditional, which meant two separate branches based on initial state. Now that it's just the one branch/block I don't need `change`. Good catch.

blivet/dbus/blivet.py
dwlehman: I started out with size as a string, but went with `uint64` to avoid issues related to localization.

vpodzime: Fair enough. I can say that localization and sizes as strings is a nightmare

PR #423, opened 2016-05-17
Tests for crucial methods of core device classes.

dwlehman:
This adds testing of `create`, `destroy`, `setup`, `teardown` methods for `StorageDevice`, `DiskDevice`, `PartitionDevice`, `LVMVolumeGroupDevice`, `LVMLogicalVolumeDevice`, and `MDRaidArrayDevice`. It doesn't increase coverage by much, but it does test some critical logic in the device classes.

### DIFF
--- a/tests/devices_test/device_methods_test.py
+++ b/tests/devices_test/device_methods_test.py
@@ -1,0 +1,404 @@
+import unittest
+from unittest.mock import patch, Mock, PropertyMock
+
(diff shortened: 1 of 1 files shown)


dwlehman:
Added some fixes for `pylint` and `pep8`.

snbueno:
As far as I can tell, these all look ok.

PR #424, opened 2016-05-18
Fix a typo when checking whether we're using an FBA DASD. (#1233438)

snbueno:
Resolves: rhbz#1233438


dwlehman:
How about `arg.type.startswith("FBA")` instead?

snbueno:
That's fine with me, but I can't test it until tomorrow since I don't have direct access to an FBA DASD. And I've learned that I should always test this. I'll update the PR once I have.

vpodzime:
Looks good to me.

PR #426, opened 2016-05-23
Remember VG name even if it seems to have no PVs

vpodzime:
If udev knows about some VG (or more precisely LVs from such VG), but we don't
see any PVs of that VG, we still need to remember the VG's name because trying
to create a new one with the same name would/could result into a failure as LVM
may know (better) about the VG.

Resolves: rhbz#1245038


vojtechtrefny:
Looks good to me.

PR #427, opened 2016-05-27
Improved handling of corrupt or otherwise unsupported disklabels

dwlehman:
This is a total reversal from the "unusable storage" dialog I added last year. The idea here is that blivet should not be constrained by what disklabels parted supports. If there are partitions in the system for it, we should be able to identify them and whatever it built on them. We should also be able to completely clear them so that users can get to a known-good state. Here are the main points of what this patch set does:

- include partitions and their descendants even if parted doesn't support the disklabel
  - we cannot support modifying the partition table in this case
  - we can, however, use the devices as-is or completely wipe the disk to start over
- make a better effort to find the disk a partition belongs on, regardless of whether `udev`/`blkid` provide expected data

The corresponding anaconda piece will hide all devices built on an unsupported disklabel from the user while allowing them to clear the disk or just ignore it. Kickstart's `clearpart` should also work as expected now with unsupported disklabels.

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -215,7 +215,7 @@ class ActionList(object, metaclass=SynchronizedMeta):
 
         log.info("resetting parted disks...")
         for device in devices:
-            if device.partitioned:
+            if device.partitioned and device.format.supported:
                 device.format.reset_parted_disk()
 
             if device.original_format.type == "disklabel" and \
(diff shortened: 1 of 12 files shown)


dwlehman:
jenkins, test this please

vpodzime:
Looks good to me.

PR #428, opened 2016-05-27
Improved handling of corrupt or otherwise unsupported disklabels (rhel7)

dwlehman:
Same as #427 except for `rhel7-branch`, which means I had to drag over a few minor things to make it all work the same.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -69,10 +69,16 @@ class DiskLabel(DeviceFormat):
         self._origPartedDisk = None
         self._alignment = None
         self._endAlignment = None
+        self._supported = True
 
         if self.partedDevice:
-            # set up the parted objects and raise exception on failure
-            self.updateOrigPartedDisk()
+            # set up the parted objects
+            try:
+                self.updateOrigPartedDisk()
+            except Exception as e:  # pylint: disable=broad-except
+                self._supported = False
+                self._labelType = kwargs.get("labelType") or ""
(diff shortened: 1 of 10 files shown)


blivet/formats/disklabel.py
```diff
@@ -409,30 +427,15 @@ def endAlignment(self):
...
-            disk_length = read_int_from_sys("%s/size" % disk_root)
-            sector_size = read_int_from_sys("%s/queue/logical_block_size" % disk_root)
-            partition_names = [n for n in os.listdir(disk_root) if n.startswith(disk_name)]
-            used_sectors = 0
-            for partition_name in partition_names:
-                partition_root = sys_block_root + partition_name
-                partition_length = read_int_from_sys("%s/size" % partition_root)
-                used_sectors += partition_length
-
-            free = Size((disk_length - used_sectors) * sector_size)
```
vpodzime: Can we really get rid of all this?

vpodzime:
Looks good to me other than the double-safety check above.

blivet/formats/disklabel.py
dwlehman: Well, it is possible that sometimes it is used. However, it cannot handle `device-mapper` or `md`, and there are many other critical pieces of `DiskLabel` functionality that would not be available if we fail to get a `parted.Disk` instance. I think this one area of partial support in the absence of a `parted.Disk` is not very useful given how many other areas that become unusable in its absence.

vpodzime: Makes sense.

PR #430, opened 2016-05-27
Fix a few minor bugs

dwlehman:
This is the rhel7-branch version of #429 plus a little cleanup of recursive teardown so that we don't try to tear down inactive lvs (it looks especially ugly when the pv is md since then lvm denies any knowledge of the vg).


snbueno:
These look ok as well.

PR #432, opened 2016-06-02
Do not use deprecated pyudev methods

vojtechtrefny:
'pyudev.Device.from_sys_path' and 'pyudev.Device.from_device_file'
are deprecated since pyudev 0.18.


vpodzime:
These both look good to me.

dwlehman:
Looks good to me too.

PR #435, opened 2016-06-08
Round down to nearest MiB value when writing ks parittion info. (#1257997)

snbueno:
On s390x in particular, some partition alignment issue is causing fractional
sizes to be reported. Pykickstart doesn't take anything except int values for
partition info, hence the call to roundToNearest.

This change only affects the data that is written to ks.cfg.

Resolves: rhbz#1257997


vpodzime:
Looks good to me too.

PR #436, opened 2016-06-09
Fixes for calculations of space taken from a VG by LVs

vpodzime:
These fix a few issues with internal LVs, metadata and data part sizes etc. The change in behavior that we no longer include the thin pool's padding into its ``vg_space_used`` for existing thin pools is for the better because that way we can report consistent numbers as LVM for VG's free space.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -1040,7 +1040,7 @@ class LVRequest(Request):
 
         # Round up to nearest pe. For growable requests this will mean that
         # first growth is to fill the remainder of any unused extent.
-        self.base = int(lv.data_vg_space_used // lv.vg.pe_size)
+        self.base = int(lv.size // lv.vg.pe_size)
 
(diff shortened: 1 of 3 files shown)


blivet/partitioning.py
```diff
@@ -1040,7 +1040,7 @@ def __init__(self, lv):
 
         # Round up to nearest pe. For growable requests this will mean that
         # first growth is to fill the remainder of any unused extent.
-        self.base = int(lv.data_vg_space_used // lv.vg.pe_size)
+        self.base = int(lv.size // lv.vg.pe_size)
```
dwlehman: What is the reason for this change?

vpodzime: The fact that it makes sense. :) ``lv.size`` is a better value for ``base`` than ``data_vg_space_used`` which also includes the data part of the cache (if any) and potentially some other things (in future).

PR #437, opened 2016-06-10
fcoe: don't use dcb for autoconnecting of bnx2x and bnx2fc (#1261703)

rvykydal:
Also use fcoemon instead of fipvlan for bnx2fc.

Resolves: rhbz#1261703


vpodzime:
Looks good to me.

rvykydal:
Pushed

PR #441, opened 2016-06-20
Fix udev.resolve_glob to match paths too

vojtechtrefny:
Resolve_glob currently matches devices using only names but
Anaconda uses it only with paths. This patch changes behaviour
of resolve_glob to match devices using both path or name.

Related: rhbz#1200833

-------
Anaconda actually uses `resolve_glob` only with path -- https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/kickstart.py#L196 -- so this never worked as intended,

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -112,23 +112,40 @@ def resolve_devspec(devspec, sysname=False):
         return ret.sys_name if sysname else device_get_name(ret)
 
 def resolve_glob(glob):
+    """ Find all devices (names) matching by path or name
+
+        :param str glob: glob to match device names or paths against
+        :returns: list of device names matching :param:`glob`
+
+    """
     import fnmatch
-    ret = []
 
-    if not glob:
-        return ret
+    paths = []
 
-    for dev in get_devices():
-        name = device_get_name(dev)
+    if glob.startswith("/dev"):
+        dpath = "/".join(glob.split("/")[:-1])
+        fglob = glob.split("/")[-1]
+    else:
+        dpath = "/dev"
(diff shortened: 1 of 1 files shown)


blivet/udev.py
```diff
@@ -112,23 +112,40 @@ def resolve_devspec(devspec, sysname=False):
...
     import fnmatch
-    ret = []
 
-    if not glob:
-        return ret
+    paths = []
 
-    for dev in get_devices():
-        name = device_get_name(dev)
+    if glob.startswith("/dev"):
```
vpodzime: Might be better to check for ``"/dev/"`` here since the code below relies on the fact that there's the ``/`` after ``/dev``

blivet/udev.py
```diff
@@ -112,23 +112,40 @@ def resolve_devspec(devspec, sysname=False):
...
-        if fnmatch.fnmatch(name, glob):
-            ret.append(name)
-        else:
-            for link in device_get_symlinks(dev):
-                if fnmatch.fnmatch(link, glob):
-                    ret.append(name)
+    for fname in os.listdir(dpath):
+        if fnmatch.fnmatch(fname, fglob):
+            fpath = os.path.realpath(dpath + "/" + fname)
+            paths.append(fpath)
```
vpodzime: This ``for`` cycle is what ``glob.glob()`` does, AFAICT.

vpodzime: Except that the symlink would have to be resolved afterwards.

vpodzime:
Apart from the above comments I think function is worth adding tests.

PR #442, opened 2016-06-20
Fix udev.resolve_glob to match device path too

vojtechtrefny:
Related: rhbz#1200833

-------
Much simpler version of #441

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -120,8 +120,9 @@ def resolve_glob(glob):
 
     for dev in get_devices():
         name = device_get_name(dev)
+        path = device_get_devname(dev)
 
-        if fnmatch.fnmatch(name, glob):
+        if fnmatch.fnmatch(name, glob) or fnmatch.fnmatch(path, glob):
             ret.append(name)
         else:
             for link in device_get_symlinks(dev):
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -286,6 +286,29 @@ def device_is_loop(info):
     return (device_get_name(info).startswith("loop") and
             os.path.isdir("%s/loop" % device_get_sysfs_path(info)))
 
+def device_is_realdisk(info):
+    """ Return True if the udev device looks like a disk.
+
+        :param info: udevdb device entry
+        :type info: dict
+        :returns: whether the device is a disk
+        :rtype: bool
+
+        We want exclusive_disks to operate on anything that could be
+        considered a directly usable disk, ie: fwraid array, mpath, or disk.
+        Unfortunately, since so many things are represented as disks by
+        udev/sysfs, we have to define what is a disk in terms of what is
+        not a disk.
+    """
+    return (device_is_disk(info) and
+            not (device_is_cdrom(info) or
+                 device_is_partition(info) or
+                 device_is_dm_partition(info) or
+                 device_is_dm_lvm(info) or
+                 device_is_dm_crypt(info) or
+                 (device_is_md(info) and
+                  not device_get_md_container(info))))
+
 def device_get_serial(udev_info):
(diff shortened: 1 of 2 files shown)


vpodzime:
Looks good to me.

vojtechtrefny:
Update: New commit -- added `disks_only` argument to `resolve_glob` for disk-only results.

dwlehman:
What is the disks_only argument for? It seems like a strange argument for this particular function.

vojtechtrefny:
The `resolve_glob` is used in Anaconda during parsing kickstart options like `ignored-disks` etc. so it makes sense to match just disks. They don't have devicetree there and it's quite complicated to filter disks using just the name and udev.

M4rtinK:
>The resolve_glob is used in Anaconda during parsing kickstart options like ignored-disks etc. so it makes sense to match just disks. They don't have devicetree there and it's quite complicated to filter disks using just the name and udev.

Yep, there is no devicetree in the kickstart parsing phase, but we do need to know if a device specification is a disk or not for `clearpart` or `ignoredisk`.

vojtechtrefny:
Added a new method to get the udev device/info from name.

blivet/udev.py
```diff
@@ -47,6 +47,15 @@ def get_device(sysfs_path):
 
     return dev
 
+def get_device_by_name(name, subsystem="block"):
+    try:
+        dev = pyudev.Device.from_name(global_udev, subsystem, name)
```
dwlehman: Note that this will only match "system" names, eg: `md0`, `dm-1`, `sda`.

dwlehman:
I hate to rain on your parade, but...

I'm wondering if it would make more sense for anaconda to do a combination of `resolve_glob` and filtering based on `device_is_disk`, which leads me to point out that `udev.device_is_disk` is woefully inadequate for your purposes here. Any device-mapper device will get a `True` from that function. `Populator._udev_device_is_disk` has what you're after, but I think you'd have to move it into `blivet.udev` for this all to work.

vojtechtrefny:
@dwlehman: Would it make more sense to change `udev.device_is_disk` to make all the checks that `Populator._udev_device_is_disk` does (probably just on master, not for rhel7-branch, because it's an API change)? We use it (`udev.device_is_disk`) only from populator and populator.helpers.disk and we  always filter out that "non-disk" matches. Or is a separate method (in udev.py) better?

dwlehman:
That definitely seems like the right thing to do (move `Populator._udev_device_is_disk` to `udev.device_is_disk`), and I think it will actually work.

vojtechtrefny:
Added a new function `udev.device_is_realdisk` that will do all the checks `Populator._udev_device_is_disk` does on master. I'll create a separate PR for master/3.0-branch that will change behaviour `udev.device_is_disk` there.

M4rtinK:
The udev.device_is_realdisk() function works fine for my usecase (https://github.com/rhinstaller/anaconda/pull/682), thanks! :)

blivet/udev.py
```diff
@@ -286,6 +286,29 @@ def device_is_loop(info):
     return (device_get_name(info).startswith("loop") and
             os.path.isdir("%s/loop" % device_get_sysfs_path(info)))
 
+def device_is_realdisk(info):
```
dwlehman: You should change `DeviceTree.udevDeviceIsDisk` to call this function as well so the code isn't in two places.

dwlehman:
Looks good otherwise.

blivet/udev.py
vojtechtrefny: Fixed.

vpodzime:
Looks good to me.

PR #443, opened 2016-06-22
Fix not enough free space after creating extended partition (#1252350)

vojtechtrefny:
When adding fourth partition extended partition is automatically
created. Because the extended partition takes 1 MiB of the free space
we might not have enough free space for the logical partition.
This patch shrinks the logical partition if this problem occurs.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -916,7 +916,13 @@ def allocate_partitions(storage, disks, partitions, freespace):
         if part_type == parted.PARTITION_EXTENDED and \
            part_type != _part.req_part_type:
             log.debug("creating extended partition")
-            add_partition(disklabel, free, part_type, None)
+            ext = add_partition(disklabel, free, part_type, None)
+
+            # extedned partition took all free space - make the size request smaller
+            if aligned_size > (ext.geometry.length - disklabel.alignment.grainSize) * disklabel.sector_size:
+                log.debug("not enough free space after creating extended "
+                          "partition - shrinking the logical partition")
+                aligned_size = aligned_size - (disklabel.alignment.grainSize * disklabel.sector_size)
 
             # now the extended partition exists, so set type to logical
             part_type = parted.PARTITION_LOGICAL


blivet/partitioning.py
```diff
@@ -916,7 +918,16 @@ def allocate_partitions(storage, disks, partitions, freespace):
         if part_type == parted.PARTITION_EXTENDED and \
            part_type != _part.req_part_type:
             log.debug("creating extended partition")
-            add_partition(disklabel, free, part_type, None)
+            ext = add_partition(disklabel, free, part_type, None)
+
+            # extedned partition took all free space -- make the logical partition 1 MiB smaller
+            if aligned_size > (ext.geometry.length) * LINUX_SECTOR_SIZE - EXTENDED_METADATA_SIZE:
```
vpodzime: The parentheses don't make such sense here as they are.

blivet/partitioning.py
```diff
@@ -916,7 +918,16 @@ def allocate_partitions(storage, disks, partitions, freespace):
...
+                log.debug("not enough free space after creating extended "
+                          "partition - shrinking the logical partition")
+                if _part.req_start_sector is None:
+                    aligned_size = align_size_for_disklabel(aligned_size - EXTENDED_METADATA_SIZE, disklabel)
+                else:
+                    aligned_size -= EXTENDED_METADATA_SIZE
```
vpodzime: I think this whole ``if`` is worth a comment on what's going on here.

blivet/devices/partition.py
```diff
@@ -50,6 +50,8 @@
 # in case the default partition size doesn't fit
 FALLBACK_DEFAULT_PART_SIZE = Size("256MiB")
 
+EXTENDED_METADATA_SIZE = Size("1 MiB")
+
```
dwlehman: I think we always use a full alignment grain for the extended metadata. If I'm right, hard-coding this to be `1 MiB` is going to cause problems at some point.

blivet/partitioning.py
dwlehman: This should also be using the alignment grain size, since that is what we use when allocating the extended partition (except when the user specified the start sector for it).

vojtechtrefny: I'm using the `align_size_for_disklabel` to align the new size (after the "extedned metada" subtraction). I thought it should be enough or not?

blivet/devices/partition.py
vojtechtrefny: So using grainSize instead of 1 MiB should be enough here? I wasn't able to find any reliable source with information about the actual metadata size for extended partitions and 1 MiB is the space between start of extended partition and start of first logical one. (And I didn't realize this is the same as the grain size, so it makes sense to use grain size instead of some "random" number.)

dwlehman:
https://github.com/rhinstaller/blivet/blob/2.0-devel/blivet/partitioning.py#L412 is where we make space for extended/logical partition metadata:

```python
        if part_type == parted.PARTITION_LOGICAL:
            # make room for logical partition's metadata
            start += alignment.grainSize
```

That saves us the trouble of having to re-align afterwards since we align the start sector just before the above code.

dwlehman:
Actually, there is a very slight chance for a failure there if the optimal alignment grain size is larger than the requested logical partition size and the minimal alignment grain size is smaller than 1 MiB. Maybe instead of just using an alignment grain (in the code I referenced in the last comment as well as in your new code) we should start with 1 MiB and align that up to the next minimal alignment boundary.

vojtechtrefny:
I've made a much simpler version. This will just check if there is enough free space for the logical partition + grain size and make the "size request" smaller if there isn't enough free space. `add_partition` should make take care of the alignment. Worst case scenario is we will fail to add the logical partition but that's the same situation we have now. And this at least fixes the original bug in blivet-gui.

vpodzime:
Looks good to me now.

PR #445, opened 2016-06-23
Cache and reuse data about multipath members

vpodzime:
With the new function provided by libblockdev we can easily get the list of
multipath member devices once and then query it instead of querying every device
whether it is an mpath member or not.

Note: This is an enhancement so it's targeting the _2.1-devel_ branch.

### DIFF
--- a/blivet/events/handler.py
+++ b/blivet/events/handler.py
@@ -24,6 +24,7 @@ from ..errors import DeviceError, EventHandlingError
 from ..devices import DM_MAJORS, MD_MAJORS
 from .. import udev
 from ..threads import SynchronizedMeta
+from ..static_data import mpath_members
 
 from .changes import data as event_data
 from .changes import record_change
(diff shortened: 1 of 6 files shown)


dwlehman:
I think you forgot to `git add` the new file in `static_data/` (or maybe I just overlooked it).

vpodzime:
> I think you forgot to git add the new file in static_data/ (or maybe I just overlooked it).

Ooops. Should be fixed now.

dwlehman:
Will this cover the case of a newly-added multipath member device? You can take a look at `EventHandlerMixin._handle_add_event` to see what code is run when a disk is added.

dwlehman:
I guess it would be easy enough to just call the `is_multipath_member` function or whatever it's called in that case.

vpodzime:
Done now.

vojtechtrefny:
Looks good to me.

blivet/events/handler.py
```diff
@@ -130,6 +131,7 @@ def _handle_add_event(self, event):
         if device is None and self._event_device_is_physical_disk(event):
             log.info("disk %s was added", udev.device_get_name(event.info))
             self.handle_device(event.info)
+            mpath_members.update_cache(udev.device_get_devname(event.info))
```
dwlehman: Shouldn't we do this _before_ calling `handle_device`? (Doesn't it rely on the cache being up-to-date?)

vpodzime: Good catch, thanks!

vpodzime:
Fixed.

vpodzime:
Jenkins, test this, please.

vpodzime:
Force-pushed to trigger jenkins tests (there was something hanged somewhere between GH and our jenkins).

vpodzime:
Jenkins, test this, please.

PR #447, opened 2016-06-24
2.0 devel port fcoemon patches from rhel

rvykydal:
porting rhel7-branch commits:

commit 2ed433f1d906accf4f77660ce0ff036797f43662
with fixed newline handling (don't eat them anymore)
squashed commit 8e0c82664c207e3aa388efcfa80f224d8f7bcc90
commit 29fe26da5fb702e19e2e01ecc3351d296be1f283

### DIFF
--- a/blivet/fcoe.py
+++ b/blivet/fcoe.py
@@ -90,8 +90,10 @@ class FCoE(object):
             log.info("No FCoE EDD info found: %s", buf.rstrip())
             return
 
-        log.info("FCoE NIC found in EDD: %s", val)
-        self.add_san(val, dcb=True, auto_vlan=True)
+        dcb = self._iface_driver(val) not in ("bnx2x", "bnx2fc")
+
+        log.info("FCoE NIC found in EDD: %s, using dcb: %s", val, dcb)
+        self.add_san(val, dcb=dcb, auto_vlan=True)
 
     def startup(self):
         if self.started:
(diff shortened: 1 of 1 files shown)


dwlehman:
The `fcoe` tools could really use some usability work. This code should be shrinking -- not growing.

blivet/fcoe.py
```diff
@@ -155,30 +204,46 @@ def write(self, root):
...
+        else:
+            anaconda_cfg = """FCOE_ENABLE="yes"
+DCB_REQUIRED="yes"
+AUTO_VLAN="yes"
+MODE="fabric"
+"""
+            lines = anaconda_cfg.splitlines(True)
+
+        with open(os.path.join(cfg_dir, "cfg-%s" % nic), "w") as new_cfg:
+            new_cfg.write("# Generated by Anaconda installer\n")
```
dwlehman: I get the sense that this code might belong in anaconda rather than blivet.

rvykydal: Well, I must try - could we just remove the comment for this patch?

I have troubles figuring out how this code (and the related code, which is the code for activating fcoe targets) should be moved to anaconda. The write_nic_fcoe_cfg is just part of activating an fcoe target in anaconda/blivet in this approach (write config file + resrtart fcoe service). I think the code belongs to blivet as the code in the same way iscsi.py and zfcp.py belong here? Another question is that the approach might not be the right one. So perhaps we should try to change it to be actually (not sure if it is currently is) working in non-installer environment?  Try to use fcoe tools (or libs if they exist) for that instead of the write config + restart fcoe service approach. Seems like quite a big task to be solved separately. The service restart was approved by fcoe savvy people, I was also looking for a way using fcoe-tools (fcoeadm) but didn't succeed at the time I was working on the patch for rhel 7.

dwlehman:
This module, along with `iscsi` and `zfcp`, is in serious need of a refactor. I don't like seeing this code keep changing without actually improving, but I suppose I shouldn't wait for someone else to do that.

dwlehman:
Jenkins, test this please.

PR #451, opened 2016-06-30
Restore the correct default partition table type for AArch64 EFI.

damarlin:
Signed-off-by: David A. Marlin <dmarlin@redhat.com>
Resolves: rhbz#1349130


damarlin:
This restores the correct default partition table type (GPT) for AArch64 EFI, which was changed by commit cee66d95.

vpodzime:
Looks good to me.

vpodzime:
@damarlin, could you please also create a PR for the 2.0-devel branch? Thanks!

PR #454, opened 2016-07-04
Use libmount for MountsCache

vojtechtrefny:
Use libmount.Table for parsing /proc/self/mountinfo instead
'manual' parsing from python.

### DIFF
--- a/blivet/mounts.py
+++ b/blivet/mounts.py
@@ -19,78 +19,28 @@
 #
 # Red Hat Author(s): Vojtech Trefny <vtrefny@redhat.com>
 #
-from collections import defaultdict
-from .udev import resolve_devspec
+
+import libmount
+import functools
+
 from . import util
-from .devicelibs import btrfs
 
 import logging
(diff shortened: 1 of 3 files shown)


blivet/mounts.py
```diff
@@ -131,47 +92,21 @@ def is_mountpoint(self, path):
...
+        """ Computes the MD5 hash on /proc/self/mountinfo and updates the cache on change
         """
 
-        md5hash = util.md5_file("/proc/mounts")
+        md5hash = util.md5_file("/proc/self/mountinfo")
 
         if md5hash != self.mounts_hash:
             self.mounts_hash = md5hash
-            self._get_active_mounts()
+            self.mountpoints = libmount.Table("/proc/self/mountinfo")
```
vpodzime: The file path is a good candidate for a constant.

vpodzime:
Looks good to me otherwise. The test failure is weird, but seems to be unrelated to these changes.

PR #456, opened 2016-07-06
Force lvremove in LVMLogicalVolumeDevice._destroy

didib:
Resolves: rhbz#1352884


vpodzime:
This indeed fixes the issue, but it's quite a heavy-hammer approach. Let's wait and see what more we can identify in the bug report.

vojtechtrefny:
This should be already fixed in https://github.com/storaged-project/blivet/pull/602 -- the problem was the LV was not deactivated before removing.

PR #460, opened 2016-07-06
Fix mpath edge case in disk filtering

dwlehman:
People should be able to set `exclusive_disks = ["mpatha"]` (via `ignoredisk --onlyuse=mpatha` in kickstart) without having to list the member disks.

### DIFF
--- a/tests/devicetree_test.py
+++ b/tests/devicetree_test.py
@@ -11,8 +11,10 @@ from blivet.actionlist import ActionList
 from blivet.errors import DeviceTreeError
 from blivet.udev import trigger
 from blivet.devicelibs import lvm
+from blivet.devices import DiskDevice
 from blivet.devices import StorageDevice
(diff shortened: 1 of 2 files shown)


tests/devicetree_test.py
```diff
@@ -337,10 +337,10 @@ def test_hide_ignored_disks(self):
         self.assertTrue(mpatha in tree.devices)
 
         tree.exclusive_disks = ["sda", "sdb"]
-        with patch.object(tree, "hide"):
+        with patch.object(tree, "hide") as hide:
             tree._hide_ignored_disks()
-            tree.hide.assert_any_call(mpatha)
-            tree.hide.assert_any_call(sdc)
+            hide.assert_any_call(mpatha)
+            hide.assert_any_call(sdc)
```
vpodzime: Looks like these changes should be squashed to the previous commit.

vpodzime:
Looks good to me otherwise.

PR #461, opened 2016-07-08
Don't require existing origin when modeling snapshot

jkonecny12:
You can't add snapshot to a model when this check is enabled. I don't know why you want this but it looks as a huge drawback to me.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -1168,8 +1168,6 @@ class LVMSnapshotMixin(object):
 
         if self.origin and not isinstance(self.origin, LVMLogicalVolumeDevice):
             raise ValueError("lvm snapshot origin must be a logical volume")
-        if self.origin and not self.origin.exists:
-            raise ValueError("lvm snapshot origin volume must already exist")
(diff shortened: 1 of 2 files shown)


blivet/devices/lvm.py
```diff
@@ -1168,10 +1168,6 @@ def _init_check(self):
 
         if self.origin and not isinstance(self.origin, LVMLogicalVolumeDevice):
             raise ValueError("lvm snapshot origin must be a logical volume")
-        if self.origin and not self.origin.exists:
-            raise ValueError("lvm snapshot origin volume must already exist")
-        if self.vorigin and not self.exists:
-            raise ValueError("only existing vorigin snapshots are supported")
```
vpodzime: The second ``if`` should stay as is, we don't want to support creation of snapshots with vorigins.

vpodzime:
The tests needs to be adapted to this change. See the jenkins console output for details.

jkonecny12:
Updated

vpodzime:
Looks good to me.

PR #462, opened 2016-07-08
Allow creation of a new LV from other LVs

vpodzime:
Thin and cache pools can be created from two other LVs (data and metadata). We
don't support creation of standalone cache pools now, but we do support creation
of standalone thin pools and thus we should also support their creation from
other LVs.

A new high-level function added to the API can save us from a lot of headache
caused by people doing crazy things and ending up with inconsistent devicetree.


dwlehman:
Looks good to me.

PR #463, opened 2016-07-11
Adapt to the old-new storaged API paths

vpodzime:
storaged now provides the API compatible with UDisks2.


vojtechtrefny:
Looks good to me.

dwlehman:
Are they removing the storaged paths?

dwlehman:
I think it would be quite a lot of effort, if not impossible, to provide two distinct namespaces from the same dbus service. Presumably they're removing the storaged ones.

vpodzime:
> I think it would be quite a lot of effort, if not impossible, to provide two distinct namespaces from the same dbus service. Presumably they're removing the storaged ones.

Exactly.

vpodzime:
I just want to wait with this for the actual change to happen in Rawhide (should be any time soon).

PR #464, opened 2016-07-19
Fix udev.resolve_glob and udev.device_is_disk

vojtechtrefny:
I didn't realize we need #450 on 2.x too. I'll remove `device_is_realdisk` from 3.0 after merging this.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -322,6 +322,31 @@ def device_is_loop(info):
             os.path.isdir("%s/loop" % device_get_sysfs_path(info)))
 
 
+@util.deprecated("3.0", "udev.device_is_disk provides same functionality in 3.0")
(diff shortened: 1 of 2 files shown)


blivet/udev.py
```diff
@@ -322,6 +322,31 @@ def device_is_loop(info):
             os.path.isdir("%s/loop" % device_get_sysfs_path(info)))
 
 
+@util.deprecated("3.0", "")
```
vpodzime: This should suggest a replacement for the user code or explain why this function is no longer needed.

vpodzime:
Looks good to me otherwise.

blivet/udev.py
vojtechtrefny: Fixed.

PR #465, opened 2016-07-22
Creation of LVs from non-existing LVs

vpodzime:
The first patch enables and implements creation of new LVs from non-existing LVs - IOW, makes sure the internal LVs are created first. The other two patches just fix minor issues, I've spotted while working on the feature.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -614,8 +614,6 @@ class LVMLogicalVolumeBase(DMDevice, RaidDevice):
 
         self._from_lvs = from_lvs
         if self._from_lvs:
-            if any(not lv.exists for lv in self._from_lvs):
-                raise ValueError("Conversion of LVs only supported for existing LVs")
             if exists:
                 raise ValueError("Only new LVs can be created from other LVs")
             if size or maxsize or percent:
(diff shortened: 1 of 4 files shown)


vojtechtrefny:
Looks good to me.

blivet/devices/lvm.py
```diff
@@ -1471,6 +1475,13 @@ def vg_space_used(self):
             space += Size(blockdev.lvm.get_thpool_padding(space, self.vg.pe_size))
         return space
 
+    def _pre_create(self):
+        if self._from_lvs:
+            # make sure all the LVs this LV should be created from exist
+            for lv in self._from_lvs:
+                if not lv.exists:
+                    lv.create()
```
dwlehman: So actions won't be scheduled for creation of the component LVs?

vpodzime: Normally, such actions should be scheduled and the component LVs should already be created when this code is executed. The creation of the component LVs should only happen here as a last resort. E.g. in cases where some code doesn't use actions (like tests).

dwlehman: I think we should consider the tests a form of documentation and therefore require the same procedure when testing as in production. This is handy, but I think it is better to be consistent.

vpodzime: Makes sense.

vpodzime:
I'm closing this PR in favor of #468 which requires all the patches from here anyway.

PR #467, opened 2016-07-25
Suggest container names based on current hostname in installer (#1359…

rvykydal:
…631)

We stopped setting ksdata's hostname from hostname looked up by anaconda (from
dhcp or dns lookup) which is same as the current installer hostname in most
cases, so use the current hostname for suggesting container names to keep
former behaviour.

Resolves: rhbz#1359631

Related: rhbz#1290858

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -35,6 +35,7 @@ _sysroot = ROOT_PATH
 shortProductName = 'blivet'
 productName = 'blivet'
 ERROR_RAISE = 0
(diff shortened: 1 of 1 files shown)


vpodzime:
Looks good to me for the *rhel7-branch*. I'd prefer some cleaner solution on the other branches though.

rvykydal:
jenkins, test this please

rvykydal:
Updated for failing pylint tests

rvykydal:
Pushed

PR #468, opened 2016-07-26
Actions for creating LVs from other LVs

vpodzime:
The first three patches are the same as in the PR #465 and may be ignored here. The next 5 patches tweak the little details we need to make actions for creating LVs from other LVs work. The last patch adds tests for the newly-supported functionality.

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -291,9 +291,20 @@ class DeviceTreeBase(object, metaclass=SynchronizedMeta):
                 raise DeviceTreeError("device is already in the tree")
 
         if action.is_create and action.is_device:
+            # if adding an LV constructed from other LVs, we need to remove the
+            # LVs it's supposed to be constructed from from the device tree
+            if isinstance(action.device, LVMLogicalVolumeDevice) and action.device.from_lvs:
+                for lv in action.device.from_lvs:
+                    if lv in self._devices:
+                        self._remove_device(lv)
             self._add_device(action.device)
         elif action.is_destroy and action.is_device:
             self._remove_device(action.device)
+            # if removing an LV constructed from other LVs, we need to put the
+            # LVs it's supposed to be constructed from back into the device tree
+            if isinstance(action.device, LVMLogicalVolumeDevice) and action.device.from_lvs:
+                for lv in action.device.from_lvs:
+                    self._add_device(lv, new=False)
         elif action.is_create and action.is_format:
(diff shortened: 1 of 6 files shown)


blivet/devicetree.py
```diff
@@ -203,6 +203,12 @@ def _remove_device(self, dev, force=None, modparent=True):
                        device.disk == dev.disk:
                         device.update_name()
 
+        if isinstance(dev, LVMLogicalVolumeDevice) and dev.from_lvs:
+            # if removing an LV constructed from other LVs, we need to put the
+            # LVs it's supposed to be constructed from back into the device tree
+            for lv in dev.from_lvs:
+                self._add_device(lv, new=False)
```
dwlehman: I wonder if this should instead go in `DeviceTreeBase._cancel_action`.

blivet/devicetree.py
```diff
@@ -152,6 +152,13 @@ def _add_device(self, newdev, new=True):
             if parent not in self._devices:
                 raise DeviceTreeError("parent device not in tree")
 
+        if isinstance(newdev, LVMLogicalVolumeDevice) and newdev.from_lvs:
+            # if adding an LV constructed from other LVs, we may need to remove
+            # the LVs it's supposed to be constructed from from the device tree
+            for lv in newdev.from_lvs:
+                if lv in self._devices:
+                    self._remove_device(lv)
+
```
dwlehman: If the code to do the reverse moves to `cancel_action` we'll obviously want to move this code into `_register_action`.

blivet/devicetree.py
vpodzime: I'm not sure about this. It makes more sense to me to make sure that whenever the device is added/removed it takes care about its components no matter if it is an action cancellation or (possibly) something else.

dwlehman: When would you remove such a device other than in the context of cancelling an action? The only example I can think of is trying to clear out a devicetree, in which case we should certainly not be quietly adding devices back into the tree.

dwlehman:
I don't think we should implicitly create internal LVs as part of creating the top-level LV. We don't do this anywhere else in blivet.

blivet/devicetree.py
vpodzime: OK, makes sense. I'll move this to ``_cancel_action`` and the chunk below to ``_register_action``

vpodzime:
The test failure is due to some weird issue with pykickstart.

dwlehman:
Looks good to me.

PR #469, opened 2016-07-26
Broke backwards API compatibility of LUKS handling

japokorn:
Removed backwards API compatibility
API optimized for LUKS data singleton

### DIFF
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -64,30 +64,20 @@ def parted_exn_handler(exn_type, exn_options, exn_msg):
 
 
 class PopulatorMixin(object, metaclass=SynchronizedMeta):
-    def __init__(self, passphrase=None, luks_dict=None, disk_images=None):
+    def __init__(self, disk_images=None):
         """
-            :keyword passphrase: default LUKS passphrase
-            :keyword luks_dict: a dict with UUID keys and passphrase values
-            :type luks_dict: dict
             :keyword disk_images: dictoinary of disk images
             :type list: dict
-
         """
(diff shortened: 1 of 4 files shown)


blivet/populator/populator.py
```diff
@@ -64,16 +64,12 @@ def parted_exn_handler(exn_type, exn_options, exn_msg):
...
-            :keyword luks_dict: a dict with UUID keys and passphrase values
-            :type luks_dict: dict
             :keyword disk_images: dictoinary of disk images
             :type list: dict
-
         """
-        self.reset(passphrase=passphrase, luks_dict=luks_dict, disk_images=disk_images)
+        self.reset(passphrase=None, luks_dict={}, disk_images=disk_images)
 
     def reset(self, passphrase=None, luks_dict=None, disk_images=None):
```
dwlehman: Shouldn't this method also drop the `passphrase` and `luks_dict` arguments?

dwlehman:
I think you'll have to stop passing luks-related arguments to `DeviceTree.__init__` and/or `DeviceTree.reset` from `Blivet`.

blivet/populator/populator.py
japokorn: The original idea was to keep those so calling reset can set passphrase/luks_dict without additional method call. Removed anyway, since apparently it is not needed so often.

japokorn:
Argument passing fixed.

dwlehman:
Looks good except for one small thing: I think you should call `luks_data.reset` from `Populator.reset` instead of `DeviceTreeBase.reset`.

japokorn:
luks_data.reset moved to PopulatorMixin, fixed

vojtechtrefny:
Looks good to me now.

PR #470, opened 2016-07-26
Ensure biosboot shows up in kickstart (#1242666)

kellinm:
Added support to detect the biosboot partition.

The biosboot partition, unlike other bootloaders, is not considered a stage1
device and is not mountable therefore it never actually showed up. Detection
of the biosboot partition required a special case that does not impact the
internal bootloader device properties which would have wreaked havoc
elsewhere in the codebase.

Resolves: rhbz#1242666

### DIFF
--- a/tests/blivet_test.py
+++ b/tests/blivet_test.py
@@ -18,6 +18,7 @@ class BlivetTestCase(unittest.TestCase):
         in the kickstart data
         '''
 
+        # prepboot test case
         with patch('blivet.Blivet.bootLoaderDevice', new_callable=PropertyMock) as mockBootLoaderDevice:
             with patch('blivet.Blivet.mountpoints', new_callable=PropertyMock) as mockMountpoints:
                 # set up prepboot partition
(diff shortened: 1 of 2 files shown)


tests/blivet_test.py
```diff
@@ -18,22 +18,44 @@ def test_bootloader_in_kickstart(self):
...
+            with patch('blivet.Blivet.mountpoints', new_callable=PropertyMock) as mockMountpoints:
+                # set up biosboot partition
+                biosboot_device_obj = PartitionDevice("biosboot_partition_device")
+                biosboot_device_obj.size = Size('1MiB')
+                biosboot_device_obj.format = formats.getFormat("biosboot")
+
+                biosboot_blivet_obj = Blivet()
+
+                # mountpoints must exist for updateKSData to run
+                mockDevices.return_value = [ biosboot_device_obj ]
```
dwlehman: Please remove the spaces around "biosboot_device_obj" here.

tests/blivet_test.py
```diff
@@ -18,22 +18,44 @@ def test_bootloader_in_kickstart(self):
...
                 # initialize ksdata
-                test_ksdata = returnClassForVersion()()
-                blivet_obj.ksdata = test_ksdata
-                blivet_obj.updateKSData()
+                prepboot_ksdata = returnClassForVersion()()
+                prepboot_blivet_obj.ksdata = prepboot_ksdata
+                prepboot_blivet_obj.updateKSData()
 
-        self.assertTrue("part prepboot" in str(blivet_obj.ksdata))
+        self.assertTrue("part prepboot" in str(prepboot_blivet_obj.ksdata))
```
vpodzime: While modifying this line, change it to ``self.assertIn`` please. The same applies to a similar line below.

kellinm:
Updated based on @vpodzime 's feedback.

vpodzime:
Looks good to me now. We will need a *2.0-devel* version of this change too.

PR #471, opened 2016-07-27
Fix checks when changing fslabel using actions

vojtechtrefny:
Found this when adding support for relabel using `ActionConfigureFormat` to blivet-gui. Checks (using `write_label` with `dry_run=True`) are done in constructor of the actions, but `self.label` is set in `apply`, so these two checks will always fail in dry run mode.


vpodzime:
Looks good to me.

dwlehman:
Looks good to me too.

PR #438, opened 2016-06-13
Make use of the new info provided for LVs

vpodzime:
To speed things up by not doing extra calls to libblockdev and to make things more reliable by using better source of information.

PR #472, opened 2016-07-27
Use the new way extra information is provided for LVs

vpodzime:
Instead of doing multiple calls to libblockdev for every LV to find out what its
potential pool, data and metadata LVs are, we can use the information from the
overall information about LVs libblockdev now provides.


vpodzime:
This is the first patch from the pull #438 which is now properly supported by both *libblockdev* and *lvm*. The other patch from there would still not work reliably.

dwlehman:
Looks good to me.

PR #474, opened 2016-07-28
Fix removing PVs after removing VG (#1358067)

vojtechtrefny:
Set `vgName` in lvmpv format to `None` when removing the vg, so we can remove the pv too.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -1106,7 +1106,7 @@ class LVMInternalLogicalVolumeDevice(LVMLogicalVolumeDevice):
             raise ValueError("new size must of type Size")
 
         if not self.takes_extra_space:
-            if size <= self.parent_lv.size:
+            if size <= self.parent_lv.size:  # pylint: disable=no-member
                 self._size = size
             else:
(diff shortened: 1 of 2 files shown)


dwlehman:
I'm not sure that "cannot replace active format" error makes any sense to raise in any case.

vojtechtrefny:
Added a new patch to ignore pylint errors. I think these are some bugs in pylint, because it just started failing without changes.

vojtechtrefny:
@dwlehman The exception message is a little misleading, "remove" would be probably better than "replace" because we're trying to remove the PV not "replace" it.

dwlehman:
It would be a real error to remove an active format, but we're only changing the model -- not actually removing anything. This was a paranoid check from when I first wrote blivet and didn't know exactly which conditions were going to be relevant. I tend to think we can remove that check/raise completely.

vojtechtrefny:
I think it isn't that paranoid. At least in this case -- we probably don't want to remove a PV that is part of a VG.

vpodzime:
Looks good to me.

dwlehman:
Although I still don't think the exception raised makes sense, what you are doing does look correct.

PR #480, opened 2016-08-02
Check a device is a DASD before doing DASD-specific things. (#1353667)

snbueno:
This will weed out non-DASD devices and prevent tb's occurring when we
try to check for DASD-specific attrs.

Resolves: rhbz#1353667

### DIFF
--- a/blivet/devicelibs/dasd.py
+++ b/blivet/devicelibs/dasd.py
@@ -114,6 +114,10 @@ DASD_FORMAT_CDL = 2
 
 def is_ldl_dasd(device):
     """Determine whether or not a DASD is LDL formatted."""
+    if not device.startswith("dasd"):
+        # not a dasd; bail
+        return False
+
     device = "/dev/%s" % (device,)
 
     f = open(device, "r")


blivet/devicelibs/dasd.py
```diff
@@ -114,6 +114,10 @@ class dasd_info(ctypes.Structure):
 
 def is_ldl_dasd(device):
     """Determine whether or not a DASD is LDL formatted."""
+    statusfile = "/sys/block/%s/device/status" % (device,)
+    if not os.path.isfile(statusfile):
+        return False
```
dwlehman: Only DASD have `/sys/class/block/<sysname>/device/status` ? It seems like either the commit message is inaccurate or the test is too inclusive.

snbueno: As far as s390x devices, yes. Considering is_ldl_dasd isn't going to run on non-s390x arches, I thought this was reasonable, as it's used elsewhere (make_unformatted_dasd_list I think).

dwlehman: Could you achieve the same result by checking that device type is "dasd" before calling `is_ldl_dasd`? (I assume it's being called from anaconda)

snbueno: I did consider that, but it's honestly cleaner and easier if it's done here.

Anaconda is passing a list of device names, not objects, so at best it could only really see if each of those names was in storage.devicetree.dasd before calling is_ldl_dasd. That might be a "meh, fine" level of ugliness, except the line that is collecting the list of LDL DASDs is:

`ldldasds = [d for d in self.selected_disks if is_ldl_dasd(d)]`

dwlehman: I'd be fine with it if the "test if it's a dasd" were more direct and/or convincing. I just see this causing problems at some point down the road. I'd even be satisfied if the test changed to `if device.startswith("dasd") and ...`

snbueno: That's doable.

vpodzime:
Looks good to me. Do you think it's also worth a change in the libblockdev's ``is_ldl_dasd()`` function?

snbueno:
@vpodzime , yes that's a good idea. I'll submit a PR.

PR #484, opened 2016-08-03
Allign the LV's size up to set the base size for growing

vpodzime:
That's what the LV would take from the VG if not grown at all.


dwlehman:
Looks good.

vpodzime:
Jenkins, test this, please.

PR #485, opened 2016-08-05
Device tags

dwlehman:
_I'm trying to sort out what the set of tags should be, so:
1. the set of tags is still in flux
2. the code doesn't currently match what I've written below_

Add a read/write property to `Device` containing a `set` of string tags. The primary use-case is disk selection, where users want to control which disks are used for various purposes. Ultimately, this could be exposed through `pykickstart` to add more powerful disk selection. The initial set of tags, which will be set automatically in the `Device` classes, is:

- local
    - rotational
    - removable
- remote
    - multipath

I had initially included basic device type tags, eg: `lvm`, `iscsi`, but I don't really see a use-case for them. Does anyone have use-cases in mind?

I am undecided on several questions:
1. should every tag have an inverse? (eg: local/remote, ssd/rotational)
2. what about `raid`? (I don't know if we can even reliably determine this aside from easy cases like dmraid)

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -40,6 +40,7 @@ from ..fcoe import fcoe
 import logging
 log = logging.getLogger("blivet")
 
+from .lib import Tags
 from .storage import StorageDevice
 from .container import ContainerDevice
 from .network import NetworkStorageDevice
@@ -94,6 +95,23 @@ class DiskDevice(StorageDevice):
                                sysfs_path=sysfs_path, parents=parents,
                                serial=serial, model=model,
                                vendor=vendor, bus=bus)
+
+        try:
+            ssd = int(util.get_sysfs_attr(self.sysfs_path, "queue/rotational")) == 0
+        except TypeError:  # get_sysfs_attr returns None from all error paths
+            ssd = False
+
+        self.tags.add(Tags.local)
+        if ssd:
+            self.tags.add(Tags.ssd)
+        if bus == "usb":
+            self.tags.add(Tags.usb)
+        if self.removable:
+            self.tags.add(Tags.removable)
+
+    def _clear_local_tags(self):
+        local_tags = set([Tags.local, Tags.ssd, Tags.usb, Tags.removable])
+        self.tags = self.tags.difference(local_tags)
 
     def __repr__(self):
         s = StorageDevice.__repr__(self)
@@ -204,6 +222,7 @@ class DMRaidArrayDevice(DMDevice, ContainerDevice):
         super(DMRaidArrayDevice, self).__init__(name, fmt=fmt, size=size,
                                                 parents=parents, exists=True,
                                                 sysfs_path=sysfs_path)
+        self.tags.add(Tags.local)
 
     @property
     def devices(self):
@@ -332,6 +351,17 @@ class MultipathDevice(DMDevice):
         else:
             self.parents.append(parent)
 
+    def _add_parent(self, parent):
+        super()._add_parent(parent)
+        if Tags.remote not in self.tags and Tags.remote in parent.tags:
+            self.tags.add(Tags.remote)
+
+    def _remove_parent(self, parent):
+        super()._remove_parent(parent)
+        if Tags.remote in self.tags and Tags.remote in parent.tags and \
+           not any(p for p in self.parents if Tags.remote in p.tags and p != parent):
+            self.tags.remove(Tags.remote)
+
     def _setup(self, orig=False):
         """ Open, or set up, a device. """
         log_method_call(self, self.name, orig=orig, status=self.status,
@@ -406,6 +436,8 @@ class iScsiDiskDevice(DiskDevice, NetworkStorageDevice):
                       self.node.iface,
                       self.nic)
 
+        self._clear_local_tags()
+
     def dracut_setup_args(self):
         if self.ibft:
             return set(["iscsi_firmware"])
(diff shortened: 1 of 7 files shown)


blivet/devices/disk.py
```diff
@@ -95,6 +95,23 @@ def __init__(self, name, fmt=None,
                                serial=serial, model=model,
                                vendor=vendor, bus=bus)
 
+        try:
+            ssd = not bool(int(util.get_sysfs_attr(self.sysfs_path, "queue/rotational")))
```
vpodzime: I'd rather see ``int(...) == 0`` here.

vpodzime:
I think the tags should be defined somewhere as constants (or maybe as an ``Enum``?). Otherwise I see bugs like ``"netwokr"`` and similar biting us in the future.

dwlehman:
I had imagined it being possible to combine tags, but it occurs to me that would involve parsing of boolean expressions based on tags. That's more complexity than I had in mind for this feature. Can anyone see a use-case that would not be satisfied by single-tag specification? I'm open to allowing inversion, eg: `!ssd`. Is there any reason to require a special character indicating the string is a label (eg: `@ssd`)? I was thinking of using special parameters and attributes to avoid type confusion, eg:

`new_partition(size=Size('10G'), disk_tag='ssd', ...)`

or

`new_partition(size=Size('10G'), disk_tag='!ssd', ...)`

The tag would be stored in a `req_disk_tags` attribute, which would be resolved at partition allocation time so that it handles changes in disk selection gracefully. My sense is that there is no need for `@` given the dedicated params/attrs.

vpodzime:
I agree that there's no need for any special character given the dedicated params/attrs. As for the combinations, I think we can allow passing multiple tags in the ``OR`` way - for e.g. something like *SSD or NVMe* as ``["ssd", "nvme"]``. Other boolean expressions would be an overkill, I think. And I don't see any valid use case for the inversions. People usually care about where their data end up not where they don't end up. Or is it just my naive idea?

dwlehman:
I decided to see what the kickstart community has to offer in the way of use-cases by posting a solicitation, [here](https://www.redhat.com/archives/kickstart-list/2016-August/msg00012.html).

blivet/devices/__init__.py
```diff
@@ -19,7 +19,7 @@
 # Red Hat Author(s): David Lehman <dlehman@redhat.com>
 #
 
-from .lib import device_path_to_name, device_name_to_disk_by_path, ParentList
+from .lib import device_path_to_name, device_name_to_disk_by_path, ParentList, Tags
```
vpodzime: Is this a leftover from something or intentional change to make ``Tags`` importable from ``devices``?

blivet/devices/disk.py
```diff
@@ -95,6 +96,23 @@ def __init__(self, name, fmt=None,
...
+        self.tags.add(Tags.local)
+        if ssd:
+            self.tags.add(Tags.ssd)
+        if bus == "usb":
+            self.tags.add(Tags.usb)
+        if self.removable:
+            self.tags.add(Tags.removable)
+
+    def _clear_local_tags(self):
+        local_tags = set([Tags.local, Tags.ssd, Tags.usb, Tags.removable])
```
vpodzime: This can be written as ``{Tags.local, Tags.ssd, Tags.usb, Tags.removable}`` which avoids creation of the temporary list. But I know it looks quite confusing due to the same notation for dicts.

blivet/devices/disk.py
```diff
@@ -332,6 +351,17 @@ def add_parent(self, parent):
...
 
+    def _add_parent(self, parent):
+        super()._add_parent(parent)
+        if Tags.remote not in self.tags and Tags.remote in parent.tags:
+            self.tags.add(Tags.remote)
+
+    def _remove_parent(self, parent):
+        super()._remove_parent(parent)
+        if Tags.remote in self.tags and Tags.remote in parent.tags and \
+           len(p for p in self.parents if Tags.remote in p.tags) < 2:
```
vpodzime: ``not any(p for p in self.parents if Tags.remote in p.tags and p != parent)`` would be better readable for me here, but that might be just a matter of personal preference.

blivet/devices/lib.py
```diff
@@ -27,6 +28,21 @@
 LINUX_SECTOR_SIZE = Size(512)
 
 
+class Tags(str, Enum):
```
vpodzime: Nice, I didn't know this is possible.

blivet/devices/lib.py
```diff
@@ -27,6 +28,21 @@
...
+    #   - All local parents makes a local device.
+    #   - Or should the rules be really simple and uniform, ignoring these truths?
+    #
+    local = 'local'
+    remote = 'remote'
+
+    removable = 'removable'
+
+    ssd = 'ssd'  # What about pmem, &c? Should this just be "fast"?
+    usb = 'usb'  # Should this just be "slow"?
```
vpodzime: I think we may figure these out later based on the user/developer feedback.

vpodzime:
Looks good to me other than the above nitpicks .

blivet/devices/disk.py
dwlehman: I like your version better, too.

blivet/devices/disk.py
dwlehman: The ambiguity of the `set` and `dict` initializer syntax really bothers me. It has led me to start using named constructors for the built-in types much more than I used to, but especially with `set`. I wonder if I'll get over it eventually like I did with `foo = 1 if bar else 2`.

blivet/devices/__init__.py
dwlehman: I'll remove it. I think I initially expected that it would be useful to have them importable from `devices`, but I can't remember why.

blivet/devices/disk.py
vpodzime: Yeah, I'm still on my way too.

PR #488, opened 2016-08-17
Don't crash if lvm refuses to activate an lv. (#1365758)

dwlehman:


vojtechtrefny:
Looks good to me.

blivet/devicetree.py
```diff
@@ -1552,7 +1552,14 @@ def addLV(lv):
...
                 self._addDevice(lv_device)
                 if flags.installer_mode:
-                    lv_device.setup()
+                    try:
+                        lv_device.setup()
+                    except LVMError:
+                        # The only scenario in which this is expected is an incomplete VG
+                        # and/or multiple VGs with the same name.
+                        log.warning("failed to activate lv %s -- ignoring it", lv_device.name)
+                        self._removeDevice(lv_device)
```
vpodzime: Do we really want to pretend the LV doesn't exist if we fail to activate it? There are more scenarios coming to my mind in which this may fail - like missing tools for thin metadata consistency checks, etc.

dwlehman: This is the only way I could see that doesn't involve adding new handling for devices that exist but for some reason cannot be activated. It's probably necessary to acknowledge the device exists, though -- with this patch we would lose the ability to remove the LV later.

dwlehman:
New version sets `lv_device.controllable` to False if `setup` fails. This should prevent further attempts to run `lvchange` on this LV while allowing us to remove it later if desired.

vpodzime:
> New version sets lv_device.controllable to False if setup fails. This should prevent further attempts to run lvchange on this LV while allowing us to remove it later if desired.

Nice. I didn't know we have an attribute like that. :) Looks good to me too now!

PR #489, opened 2016-08-17
Make sure a logical volume is active before formatting it

fabianorosas:
This is fixed in 1.x+ but 0.x still fails while formatting a deactivated logical volume.


vpodzime:
Looks good to me.

PR #492, opened 2016-08-23
Fix TypeError when calling blivet.reset()

vojtechtrefny:
API of devicetree.reset() has changed in #469

PR #493, opened 2016-08-25
Functional testing

vojtechtrefny:
The idea here is that Jenkins will run the `runvmtest.py` to start the tests -- it will start an existing VM (with empty disks and a LiveCD), make a snapshot of it and run the test (using ssh) reverting to the snapshot after each test run.

Few missing things:
- Some better summary for test results
- Dynamic test name detection -- there is a list of test names to run in `runvmtest` right now; I've tried to get the names automatically using `inspect` or `importlib` but without success


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
(First commit isn't part of this, it's the same one as in #492, I'll remove it before pushing.)

dwlehman:
My only criticism is that the new tests should take into account the fact that we will eventually want to run them on real hardware, whether that's through beaker or not. They're all storage-backed or functional tests; the difference is what they run on, but more importantly how they are initiated:

- running a vm
- running locally, as we would do on real hw or in beaker

I think these can be referred to as distinct test harnesses. `BlivetResetTestCase` is common to those two, so I think the harness code should either be a mixin class or should be somehow separated from the test code.

vojtechtrefny:
Added a better test result summary. And removed the "Fix TypeError..." commit.

vpodzime:
Looks good to me.

vojtechtrefny:
Added new commit to fix failing thin snapshot test case.

vpodzime:
Still looks good to me :)

PR #496, opened 2016-08-29
Use None as mountpoint for new snapshots

vojtechtrefny:
Default value for mountpoint is always None not an empty string.


vpodzime:
Looks good to me.

PR #498, opened 2016-08-30
Do not try to create format for snapshots in create_device

vojtechtrefny:
For snapshots, format already exists after creating the snapshot.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -812,7 +812,10 @@ class Blivet(object, metaclass=SynchronizedMeta):
             :rtype: None
         """
         self.devicetree.actions.add(ActionCreateDevice(device))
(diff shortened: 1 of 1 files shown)


blivet/blivet.py
```diff
@@ -812,7 +812,10 @@ def create_device(self, device):
             :rtype: None
         """
         self.devicetree.actions.add(ActionCreateDevice(device))
-        if device.format.type and not device.format_immutable:
+
+        is_snapshot = isinstance(LVMLogicalVolumeDevice) and device.is_snapshot_lv
```
vpodzime: ``isinstance()`` takes two arguments (which is what pylint is complaining about). Other than that this looks good to me.

vojtechtrefny: Oops, fixed.

PR #499, opened 2016-09-01
Disk tags as disk spec for new partitions

dwlehman:
This builds on #485 by doing two things:

1. add a constructor argument to `PartitionDevice` to allow passing tags as disk set specification
2. add handling in `partitioning.allocate_partitions` to resolve the tags to a disk set at partition allocation time

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -70,7 +70,7 @@ class PartitionDevice(StorageDevice):
                  size=None, grow=False, maxsize=None, start=None, end=None,
                  major=None, minor=None, bootable=None,
                  sysfs_path='', parents=None, exists=False,
-                 part_type=None, primary=False, weight=0):
+                 part_type=None, primary=False, weight=0, disk_tags=None):
         """
             :param name: the device name (generally a device node's basename)
             :type name: str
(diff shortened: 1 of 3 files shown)


vpodzime:
*"If multipath tags are passed..."*? Looks like you have been thinking about multipath too much lately ;)

blivet/devices/partition.py
```diff
@@ -187,6 +201,7 @@ def __init__(self, name, fmt=None, uuid=None,
             # XXX It might be worthwhile to create a shit-simple
             #     PartitionRequest class and pass one to this constructor
             #     for new partitions.
+            self.req_disk_tags = list(disk_tags or [])
```
vpodzime: ``list(disk_tags) or []``?

vpodzime:
Looks good to me otherwise.

blivet/devices/partition.py
dwlehman: I had that initially, but if `disk_tags` is `None` it'll raise `ValueError`.

dwlehman: Looking at it again, that does look pretty silly. I could do

```python
    self.req_disk_tags = list(disk_tags) if disk_tags is not None else list()
```

vpodzime: Oh, that's right. I like the above (last) version most.

PR #501, opened 2016-09-06
things almost forgotten

dwlehman:
This version of "Do not include..." is a bit clumsy in the name of safety. I don't know what the difference is between `self.size` and `self.req_size` at that point, but let's work that out in fedora.


vpodzime:
These all look good to me.

PR #507, opened 2016-09-13
2.0 packaging tweaks

dwlehman:
Port of #491 to 2.0-devel.

Does anyone know how I could get patches onto both 1.20-devel and 2.0-devel by merging?

### DIFF
--- a/python-blivet.spec
+++ b/python-blivet.spec
@@ -69,7 +69,6 @@ cp -a . %{py3dir}
 make
 
 %install
-rm -rf %{buildroot}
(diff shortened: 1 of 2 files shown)


vpodzime:
> Does anyone know how I could get patches onto both 1.20-devel and 2.0-devel by merging?

You mean to merge a single branch into those two branches? I don't think that's possible given the fact that 1.20-devel is also quite a lot forward to the point where those two branches diverged.

vpodzime:
Looks good to me.

PR #513, opened 2016-09-27
Remove several redundant recursive teardown calls.

dwlehman:
The most important one is definitely the one in `_find_existing_installations`, since it can trigger full setup and recursive teardown for each lv in vgs with an md pv.

The only one of these that I think may not be completely redundant is the one at the end of `PopulatorMixin.populate`, since anaconda runs that method directly (as opposed to using `Blivet.reset`) after adding network storage through the GUI. For that one, we could either leave it as-is or we could add an explicit `teardown_all` in anaconda after the `populate` call.

### DIFF
--- a/blivet/osinstall.py
+++ b/blivet/osinstall.py
@@ -250,7 +250,7 @@ def _find_existing_installations(devicetree):
 
         if not os.access(get_sysroot() + "/etc/fstab", os.R_OK):
             util.umount(mountpoint=get_sysroot())
-            device.teardown(recursive=True)
+            device.teardown()
             continue
 
         try:
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -418,6 +418,9 @@ class PopulatorMixin(object, metaclass=SynchronizedMeta):
 
     def teardown_disk_images(self):
         """ Tear down any disk image stacks. """
+        if not self.disk_images:
+            return
+
         self.teardown_all()
         for (name, _path) in self.disk_images.items():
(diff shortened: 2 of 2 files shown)


blivet/osinstall.py
```diff
@@ -250,7 +250,7 @@ def _find_existing_installations(devicetree):
 
         if not os.access(get_sysroot() + "/etc/fstab", os.R_OK):
             util.umount(mountpoint=get_sysroot())
-            device.teardown(recursive=True)
+            device.teardown()
```
vpodzime: Isn't this to prevent issues with the whole stack being activated in the ``device.setup()`` call above and the leaf ``device`` being teared down in the end leaving the rest of the stack active? Or is that no longer an issue? It definitely used to be.

blivet/populator/populator.py
```diff
@@ -455,9 +458,6 @@ def populate(self, cleanup_only=False):
             parted.clear_exn_handler()
             self._hide_ignored_disks()
 
-        if flags.installer_mode:
-            self.teardown_all()
```
vpodzime: Yeah, I believe this should go to anaconda.

blivet/osinstall.py
dwlehman: In `find_existing_installations` there is a `teardown_all` in a `finally` block around the call to `_find_existing_installations`, so everything should get deactivated regardless of outcome.

vpodzime: Oh, that's right! Thanks for the clarification.

PR #515, opened 2016-10-05
DeviceFactory constructor argument cleanup

dwlehman:
See the commit message for 72170b1f for the point. My thinking before I started was that this would be significant usability improvement for `DeviceFactory`.

Since it breaks the API (`Blivet.factory_device`, `devicefactory.get_device_factory`, and `DeviceFactory`) it has to go onto 3.0-devel.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -321,7 +321,9 @@ class DeviceFactory(object):
         self.fstype = None  # not included in default_settings b/c of special handling below
         self.min_luks_entropy = luks_data.min_entropy
 
+        # If a device was passed, update the defaults based on it.
         self.device = kwargs.get("device")
+        self._update_defaults_from_device()
 
         # Map kwarg 'name' to attribute 'device_name'.
         if "name" in kwargs:
@@ -347,6 +349,39 @@ class DeviceFactory(object):
         self.__actions = []
         self.__names = []
         self.__roots = []
+
+    def _update_defaults_from_device(self):
+        """ Update default settings based on passed in device, if provided. """
+        if self.device is None:
+            return
+
+        self.size = getattr(self.device, "req_size", self.device.size)
+        self.disks = getattr(self.device, "req_disks", self.device.disks[:])
+
+        self.fstype = self.device.format.type
+        self.mountpoint = getattr(self.device.format, "mountpoint", None)
+        self.label = getattr(self.device.format, "label", None)
+
+        # TODO: add a "raid_level" attribute to all relevant classes (or to RaidDevice)
+        if hasattr(self.device, "level"):
+            self.raid_level = self.device.level
+        elif hasattr(self.device, "data_level"):
+            self.raid_level = self.device.data_level
+
+        self.encrypted = isinstance(self.device, LUKSDevice)
+        self.device_name = getattr(self.device, "lvname", self.device.name)
+
+        if self.device and hasattr(self.device, "container"):
+            self.container_name = self.device.container.name
+            self.container_size = self.device.container.size_policy
+            if hasattr(self.device.container, "data_level"):
+                self.container_raid_level = self.device.container.data_level
+            elif (hasattr(self.device.container, "pvs") and
+                  len(self.device.container.pvs) == 1 and
+                  hasattr(self.device.container.pvs[0].raw_device, "level")):
+                self.container_raid_level = self.device.container.pvs[0].raw_device.level
+            self.container_encrypted = all(isinstance(p, LUKSDevice)
+                                           for p in self.device.container.parents)
 
     @property
     def encrypted(self):
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -125,10 +125,8 @@ def get_device_type(device):
     return device_type
 
 
-def get_device_factory(blivet, device_type=DEVICE_TYPE_LVM, size=None, **kwargs):
+def get_device_factory(blivet, device_type=DEVICE_TYPE_LVM, **kwargs):
     """ Return a suitable DeviceFactory instance for device_type. """
-    disks = kwargs.pop("disks", [])
-
     class_table = {DEVICE_TYPE_LVM: LVMFactory,
                    DEVICE_TYPE_BTRFS: BTRFSFactory,
                    DEVICE_TYPE_PARTITION: PartitionFactory,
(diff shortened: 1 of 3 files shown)


vpodzime:
Jenkins, test this, please.

blivet/devicefactory.py
```diff
@@ -137,9 +135,9 @@ def get_device_factory(blivet, device_type=DEVICE_TYPE_LVM, size=None, **kwargs)
                    DEVICE_TYPE_DISK: DeviceFactory}
 
     factory_class = class_table[device_type]
-    log.debug("instantiating %s: %s, %s, %s, %s", factory_class,
-              blivet, size, [d.name for d in disks], kwargs)
-    return factory_class(blivet, size, disks, **kwargs)
+    log.debug("instantiating %s: %s, %s, %s", factory_class,
+              blivet, [d.name for d in kwargs.get("disks", [])], kwargs)
```
vpodzime: The original code removed the ``"disks"`` key from ``kwargs`` while the new version keeps it in. Was the old version wrong?

blivet/devicefactory.py
```diff
@@ -250,12 +248,21 @@ class DeviceFactory(object):
...
+                         "raid_level": None,
+                         "encrypted": None,
+                         "container": None,
+                         "device": None,
+                         "container_name": None,
+                         "container_size": SIZE_POLICY_AUTO,
+                         "container_raid_level": None,
+                         "container_encrypted": None}
+
+    def __init__(self, *args, **kwargs):  # pylint: disable=unused-argument
```
vpodzime: What are ``*args`` good for here?

blivet/devicefactory.py
```diff
@@ -294,50 +301,70 @@ def __init__(self, storage, size, disks, fstype=None, mountpoint=None,
...
-        self.container_encrypted = container_encrypted
-
-        self.container_name = container_name
-        self.device_name = name
-
-        self.container_size = container_size
-
-        self.container = None
-        self.device = device
+        self.storage = kwargs["storage"]  # a Blivet instance
```
vpodzime: This shouldn't be just a kwarg if we actually require it.

blivet/devicefactory.py
```diff
@@ -294,50 +301,70 @@ def __init__(self, storage, size, disks, fstype=None, mountpoint=None,
             :keyword container_encrypted: whether to encrypt the container
             :type container_encrypted: bool
             :keyword container_size: requested container size
-            :type container_size: :class:`~.size.Size`
+            :type container_size: :class:`~.size.Size`, :const:`SIZE_POLICY_AUTO` (the default),
+                                  or :const:`SIZE_POLICY_MAX`
```
vpodzime: Also need to change a couple of ``:param ...`` to ``:keyword ...`` above.

blivet/devicefactory.py
```diff
@@ -294,50 +301,70 @@ def __init__(self, storage, size, disks, fstype=None, mountpoint=None,
...
+            self.container_name = kwargs["container_name"]
+        if "container_size" in kwargs:
+            self.container_size = kwargs["container_size"]
+        if "container_raid_level" in kwargs:
+            self.container_raid_level = kwargs["container_raid_level"]
+        if "container_encrypted" in kwargs:
+            self.container_encrypted = kwargs["container_encrypted"]
+
+        if "min_luks_entropy" in kwargs:
+            self.min_luks_entropy = kwargs["min_luks_entropy"]
```
vpodzime: Maybe this block could be simplified by creating a copy of the defaults and then calling ``defaults_copy.update(kwargs)`` followed by the for cycle with ``setattr`` above moved here?

blivet/devicefactory.py
```diff
@@ -294,50 +301,70 @@ def __init__(self, storage, size, disks, fstype=None, mountpoint=None,
...
+        self.storage = kwargs["storage"]  # a Blivet instance
+
+        self._encrypted = None
+        self._raid_level = None
+        self._container_raid_level = None
+
+        for setting, value in self._default_settings.items():
+            setattr(self, setting, value)
+
+        self.fstype = None  # not included in default_settings b/c of special handling below
```
vpodzime: What is the special handling below? I cannot find it.

blivet/devicefactory.py
```diff
@@ -1694,8 +1721,8 @@ class BTRFSFactory(DeviceFactory):
     child_factory_class = PartitionSetFactory
     child_factory_fstype = "btrfs"
 
-    def __init__(self, storage, size, disks, **kwargs):
-        super(BTRFSFactory, self).__init__(storage, size, disks, **kwargs)
+    def __init__(self, *args, **kwargs):
```
vpodzime: I think ``*args`` should be dropped here and in all the above ``__init__`` functions.

blivet/devicefactory.py
dwlehman: The old version grabbed "disks" from kwargs and passed it as a positional argument, which is one of the things this set changes.

blivet/devicefactory.py
dwlehman: Nothing in particular.

blivet/devicefactory.py
dwlehman: Good catch -- thanks.

blivet/devicefactory.py
dwlehman: It's close to the bottom of `__init__`, where we set `self.fstype` to the blivet default via a call to `Blivet.get_fstype` to choose a good default at runtime.

blivet/devicefactory.py
dwlehman: Yeah, it probably could. The only reason I didn't already do this is that the ctor parameter `name` maps to the attribute `device_name`. There may be a couple of others but, for the most part, what you suggest seems possible.

blivet/devicefactory.py
vpodzime: Oh, I overlooked it being passed that way, sorry for the noise.

blivet/devicefactory.py
vpodzime: Still don't see a difference between setting it to ``None`` here and having it in the defaults.

dwlehman: I left it out of the dict because the below block prevents unit testing based on the dict from passing.

dwlehman:
FYI I'm planning to squash "Simplify override..." from those last two commits into the "kwargs" commit on Monday.

dwlehman:
Okay, I squashed them. Ready for review.

blivet/devicefactory.py
vpodzime: Makes sense, thanks!

blivet/devicefactory.py
```diff
@@ -348,6 +350,39 @@ def __init__(self, storage, **kwargs):
         self.__names = []
         self.__roots = []
 
+    def _update_defaults_from_device(self):
+        """ Set up default settings based on device, if provided, or internal defaults. """
```
vpodzime: I find the docstring a bit misleading as the method doesn't seem to update the default settings based on internal defaults. Maybe just drop the *"or internal defaults." part?

PR #516, opened 2016-10-12
Device name now checked only for new devices

japokorn:
Device name check moved from Device to StorageDevice. Now is the name checked only when
device does not exist.

Added tests to check this behavior.

### DIFF
--- a/tests/devices_test/device_names_test.py
+++ b/tests/devices_test/device_names_test.py
@@ -26,6 +26,30 @@ class DeviceNameTestCase(unittest.TestCase):
         for name in bad_names:
             self.assertFalse(sd.is_name_valid(name))
 
+        # Check that name validity check is omitted (only) when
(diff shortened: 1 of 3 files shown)


tests/devices_test/device_names_test.py
```diff
@@ -26,6 +26,29 @@ def test_storage_device(self):
...
+                self.fail("Device name check failed when it shouldn't")
+
+        for name in bad_names:
+            try:
+                StorageDevice(name, exists=True)
+            except ValueError as e:
+                if ' is not a valid name for this device' in str(e):
+                    self.fail("Device name checked on already existing device")
+
+            with self.assertRaises(ValueError):
```
vpodzime: This could use ``self.assertRaisesRegexp()`` and check the message too.

japokorn:
Pushed.

PR #518, opened 2016-10-27
Fix iscsi problems: type of arg passed to storaged, store auth info when logging in with it

AdamWill:
These commits fix a couple of problems in iSCSI support. The first is simple: the type of the `port` arg doesn't match the type string (`q` is an int). The second is more complicated; when we moved from libiscsi to storaged the `node` objects were changed from libiscsi `PyIscsiNode` instances to a new namedtuple class called `NodeInfo`...but as a consequence of that, they no longer provide any access to the usernames and passwords used to authenticate to the target, which both  blivet and anaconda actually expect. This restores that information in a different form (it just gets stuffed into the `NodeInfo` tuple) and adapts blivet to the change; I'll send a corresponding PR for anaconda.

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -365,7 +365,7 @@ class iSCSI(object):
             if r_password:
                 auth_info["r_password"] = GLib.Variant("s", r_password)
 
-            args = GLib.Variant("(sqa{sv})", (ipaddr, port, auth_info))
+            args = GLib.Variant("(sqa{sv})", (ipaddr, int(port), auth_info))
             nodes, _n_nodes = self._call_initiator_method("DiscoverSendTargets", args)
 
             found_nodes = _to_node_infos(nodes)
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -66,8 +66,29 @@ def has_iscsi():
     return True
 
 
-NodeInfo = namedtuple("NodeInfo", ["name", "tpgt", "address", "port", "iface"])
 TargetInfo = namedtuple("TargetInfo", ["ipaddr", "port"])
+
+
+class NodeInfo(object):
+    """Simple representation of node information."""
+    def __init__(self, name, tpgt, address, port, iface):
+        self.name = name
+        self.tpgt = tpgt
+        self.address = address
+        self.port = port
+        self.iface = iface
+        # These get set by log_into_node, but *NOT* _login
+        self.username = None
+        self.password = None
+        self.r_username = None
+        self.r_password = None
+
+    @property
+    def conn_info(self):
+        """The 5-tuple of connection info (no auth info). This form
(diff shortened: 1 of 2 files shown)


blivet/iscsi.py
```diff
@@ -239,7 +242,7 @@ def _login(self, node_info, extra=None):
             extra = dict()
         extra["node.startup"] = GLib.Variant("s", "automatic")
 
-        args = GLib.Variant("(sisisa{sv})", tuple(list(node_info) + [extra]))
+        args = GLib.Variant("(sisisa{sv})", node_info[:5] + (extra,))
```
vpodzime: Nice, I didn't know that one can concatenate tuples like this.

blivet/iscsi.py
```diff
@@ -414,6 +417,14 @@ def log_into_node(self, node, username=None, password=None,
...
             if not self._mark_node_active(node):
                 log.error("iSCSI: node not found among discovered")
+            if username:
+                node.username = username
+            if password:
+                node.password = password
+            if r_username:
+                node.r_username = r_username
+            if r_password:
+                node.r_password = r_password
```
vpodzime: A don't this can work. A ``NamedTuple`` is immutable, AFAICT. If we need to do this, we need to change the ``NodeInfo`` into a traditional class.

AdamWill: ah, you're right. As I said in the bug report, I didn't test it with an actual authenticated iSCSI target yet. Hmm, guess I get to think about the design again. More terrible ideas pending!

AdamWill:
Well, I guess this, then. If you think it's worth adding `getAuth` and `setAuth` I could do that, but I don't really think it is.

blivet/iscsi.py
```diff
@@ -66,10 +66,28 @@ def has_iscsi():
...
+        self.address = address
+        self.port = port
+        self.iface = iface
+        # These get set by log_into_node: note _login does *NOT*
+        self.username = None
+        self.password = None
+        self.r_username = None
+        self.r_password = None
+
+    def infotuple(self):
```
vpodzime: One last double-nitpick -- I think this should be a ``conn_info`` ``@property``.

vpodzime:
> Well, I guess this, then. If you think it's worth adding getAuth and setAuth I could do that, but I don't really think it is.

I don't think we need that.

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Jenkins, test this, please.

blivet/iscsi.py
AdamWill: someone (I think kparal) was very rude about my habit of using read-only properties once, so I tend to avoid introducing them to other people's code, though I still use them all over my own =) But if you want it that way, sure, can do.

PR #520, opened 2016-10-27
iSCSI: set `iscsi.initiator_set` when setting initiator

AdamWill:
The iSCSI class has an `initiator_set` property whose meaning
feels a bit slippery these days. It has always been set to
True in `__init__()` if iBFT is active, right after we get the
initiator name from the firmware. Prior to 9280eff7, it was
also set true by `startup()` after it wrote out INITIATOR_FILE.
In 9280eff7, that was removed, without any kind of replacement.
Now `initiator_set` will never be True unless iBFT is being
used.

This is a problem because `iscsi.write()` checks if it's True,
and immediately bails if it isn't. The result of this is that
when you do an iSCSI install with anaconda, the contents of
`/var/lib/iscsi` from the installer environment are no longer
copied in the installed system.

I wasn't entirely sure how to fix this, but this seems like the
simplest way, at least: just have the `initiator` setter set
`initiator_set` to True after it sets `self._initiator`. I'm
not sure if it might be better to ditch this flag entirely, or
make it a property that just returns True if `self._initiator`
is set, or what. I'm not really sure why we have it any more,
it kinda feels a bit vestigial.

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -124,7 +124,6 @@ class iSCSI(object):
         # This list contains nodes discovered through iBFT (or other firmware)
         self.ibft_nodes = []
         self._initiator = ""
-        self.initiator_set = False
         self.started = False
         self.ifaces = {}
 
@@ -134,7 +133,6 @@ class iSCSI(object):
             try:
                 initiatorname = self._call_initiator_method("GetFirmwareInitiatorName")[0]
                 self._initiator = initiatorname
-                self.initiator_set = True
             except Exception:  # pylint: disable=broad-except
                 log_exception_info(fmt_str="failed to get initiator name from iscsi firmware")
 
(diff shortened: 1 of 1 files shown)


AdamWill:
I tested this and confirmed that, if you use an updates.img with all the other patches I submitted plus this one, the install works and the installed system's `/var/lib/iscsi` is populated.

vpodzime:
Again thanks a lot for you work on these issues!

blivet/iscsi.py
```diff
@@ -172,6 +170,13 @@ def _call_initiator_method(self, method, args=None):
                                    connection=self._connection)
 
     @property
+    def initiator_set(self):
+        """True if initiator is set at our level."""
+        if self._initiator == "":
+            return False
+        return True
```
vpodzime: This can be reduced to ``return self._initiator == ""``.

vpodzime:
Jenkins, this is ok to test.

PR #523, opened 2016-11-10
Fix detection of 'macefi' partitions (#1393846)

AdamWill:
368a4db6 lost a crucial condition in the detection of 'macefi'
partitions in the transition to the 'populator helper' design.
Previously we checked that the parted partition 'name' (which
is a GPT property, for GPT partitions) matched the expected
value according to the macefi format, which basically means we
will only detect partitions created by a previous anaconda run
as 'macefi' (because that name is a very specific one which is
only created by anaconda in the first place).

In the transition, that condition was lost, and now we treat
any device with an HFS+ filesystem that's over 50MiB in size
as a 'macefi' device, which means we mount it at /boot/efi and
try to write all kinds of stuff to it. Not surprisingly, this
borks the install. Fortunately, HFS+ filesystems are mounted
read-only unless they have journalling disabled, so this won't
result in us messing up people's OS X partitions with any luck.

### DIFF
--- a/blivet/populator/helpers/boot.py
+++ b/blivet/populator/helpers/boot.py
@@ -54,6 +54,16 @@ class MacEFIFormatPopulator(BootFormatPopulator):
     _type_specifier = "macefi"
     _base_type_specifier = "hfsplus"
 
+    @classmethod
+    def match(cls, data, device):
+        fmt = formats.get_format(cls._type_specifier)
+        try:
+            return (super().match(data, device) and
+                    device.parted_partition.name == fmt.name)
+        except AttributeError:
+            # just in case device.parted_partition has no name attr
+            return False
+
 
 class AppleBootFormatPopulator(BootFormatPopulator):
     _type_specifier = "appleboot"
--- a/tests/populator_test.py
+++ b/tests/populator_test.py
@@ -1114,7 +1114,22 @@ class MDFormatPopulatorTestCase(FormatPopulatorTestCase):
             self.assertEqual(array.name, array_name)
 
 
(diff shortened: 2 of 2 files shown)


dwlehman:
Jenkins, please test this.

dwlehman:
Jenkins, test this please.

AdamWill:
So I wrote tests...which was a good thing as there was a blatant error in the original (not enough args for the `super.match()` call, because I copied it out of a different file where `match()` takes fewer...)

So this fixes that, and protects against `device.parted_partition` not having a `name` attribute (which I don't *think* is likely to happen, but can't hurt to be safe), and adds the tests.

Testing this was a bit awkward, this is my best effort. I used a stub to mock `parted_partition` because it's a bit deep to comfortably do with `mock.patch`. For the actual test we only need the `name` attribute, so I had a kinda neat approach to just fake it in-line with a `namedtuple`. But it turns out that when `partition.parted_partition` exists, the getter for the `partition.size` property will call `partition.parted_partition.getLength()`...so we have to fake that up too. In true Inception style, let's just make it return `partition._size` (converted)...

If anyone can think of a better way, please yell. This was the best I could manage without just entirely replacing `test_match` for `MacEFIFormatPopulatorTestCase`.

AdamWill:
so this has a +1 from @dwlehman and reporter has confirmed the fix in the bug, can someone merge it? We probably want to take https://github.com/rhinstaller/blivet/pull/524 also before doing another build, though.

vpodzime:
Jenkins, this is ok to test.

PR #524, opened 2016-11-11
Fix "unknown" SAS device sysfs parsing.

vathpela:
Since the regexp matches the device type as well as the identifying
numbers, we need to pull the numbers from match groups 2 and 3, not 1
and 2.

Resolves: rhbz#1394026

### DIFF
--- a/blivet/devicelibs/edd.py
+++ b/blivet/devicelibs/edd.py
@@ -316,8 +316,8 @@ class EddEntry(object):
                         self.sas_address = int(sas_match.group(1), base=16)
                         self.sas_lun = int(sas_match.group(2), base=16)
                     elif unknown_match:
-                        self.sas_address = int(unknown_match.group(1), base=16)
-                        self.sas_lun = int(unknown_match.group(2), base=16)
+                        self.sas_address = int(unknown_match.group(2), base=16)
+                        self.sas_lun = int(unknown_match.group(3), base=16)
                     else:
                         log.warning("edd: can not match interface for %s: %s",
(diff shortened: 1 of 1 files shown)


AdamWill:
LGTM. Some background from IRC:

(adamw) if i had to take a magical carpet ride into the mind of 1-year-ago pjones, i'd *guess* you noticed from the code that it's theoretically possible for any of the different types ('SAS', 'SATA', 'RAID', I20' etc) to come up with 'unknown' instead of 'device' or 'sas_address' or 'identity_tag' or whatever
(pjones) well, it comes from http://lxr.free-electrons.com/source/drivers/firmware/edd.c?v=4.4#L130
(pjones) Oh, this is probably the kernel still not believing in the SAS EDD type correctly.
(adamw) so you threw in a catch-all regex at the end of the list with ^(\S*) instead of a specific device type at the start of the regex
(adamw) but you forgot to bump the match group numbers for that extra group later when using the capture groups
(adamw) but that's just me trying to make inferences from how the code looks
(pjones) you may be right.
(pjones) I suspect the bigger issue is that hpa never acked/responded to/applied http://www.kernelhub.org/?msg=849667&p=2

vpodzime:
Looks good to me and I know it is a blocker hot-fix, but could you also please add a new test for the "unknown" SAS device case? Thanks!

vpodzime:
Jenkins, this is ok to test.

vpodzime:
Also please provide a PR for the *2.1-devel* branch not for *master*.

PR #525, opened 2016-11-18
Require BlockDev 2.0 in the gi.require_version() call

vpodzime:
We now have libblockdev-2.0. Unfortunately, the useless function
gi.require_version() is stupid and doesn't support anything like ">= 1.0", which
is what we really need. So unless we want to add more useless code trying both
"1.0" and "2.0", we are doomed to fail on systems that have libblockdev-1.x
installed even if that's just fine for us. To prevent this let's also bump the
requires in the spec file.

Resolves: rhbz#1395791

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -75,7 +75,7 @@ log_bd_message = lambda level, msg: program_log.info(msg)
 
 import gi
 gi.require_version("GLib", "2.0")
-gi.require_version("BlockDev", "1.0")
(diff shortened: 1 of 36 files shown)


vpodzime:
Please note that the tests are doomed to fail on this as the jenkins slaves don't have libblockdev 2.0 installed (which I'll change, but only right before this is merged to prevent failures of all other PRs).

vpodzime:
Jenkins, test this, please.

PR #526, opened 2016-11-18
Change how we run e2fsck to check ext filesystems

vpodzime:
The '-p' option means "Automatically repair ("preen") the file system.  This
option will cause e2fsck to automatically fix any filesystem problems that can
be safely fixed without human intervention." which is something we really
shouldn't do as part of reset()/populate(). We should use '-n' instead "Open the
filesystem read-only, and assume an answer of `no' to all questions." which
guaranties no changes to be made on the file system.

We might want to add the '-p' functionality back at some point, but it needs to
be explicitly triggered by the user code (e.g. Anaconda). I think we need to add
a 'clean' property and a 'repair' method to the formats.FS class so that users
can see where the problem is (if any) and explicitly trigger a safe fixup
attempt if they want to.

Related: rhbz#1170803

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -296,8 +296,6 @@ class FS(DeviceFormat):
         """ Update this filesystem's current and minimum size (for resize). """
 
         #   This method ensures:
-        #   * If there are fsck errors, self._resizable is False.
-        #       Note that if there is no fsck program, no errors are possible.
         #   * If it is not possible to obtain the current size of the
         #       filesystem by interrogating the filesystem, self._resizable
         #       is False (and self._size is 0).
@@ -317,32 +315,26 @@ class FS(DeviceFormat):
         self._min_instance_size = Size(0)
         self._resizable = self.__class__._resizable
 
-        # We can't allow resize if the filesystem has errors.
+        # try to gather current size info
+        self._size = Size(0)
         try:
-            self.do_check()
-        except FSError:
-            self._resizable = False
-            raise
-        finally:
-            # try to gather current size info anyway
-            self._size = Size(0)
-            try:
-                if self._info.available:
-                    self._current_info = self._info.do_task()
-            except FSError as e:
-                log.info("Failed to obtain info for device %s: %s", self.device, e)
-            try:
-                self._size = self._size_info.do_task()
-                self._min_instance_size = self._size
-            except (FSError, NotImplementedError) as e:
-                log.warning("Failed to obtain current size for device %s: %s", self.device, e)
+            if self._info.available:
+                self._current_info = self._info.do_task()
+        except FSError as e:
+            log.info("Failed to obtain info for device %s: %s", self.device, e)
+        try:
+            self._size = self._size_info.do_task()
+        except (FSError, NotImplementedError) as e:
+            log.warning("Failed to obtain current size for device %s: %s", self.device, e)
+        else:
+            self._min_instance_size = self._size
 
-            # We absolutely need a current size to enable resize. To shrink the
-            # filesystem we need a real minimum size provided by the resize
-            # tool. Failing that, we can default to the current size,
-            # effectively disabling shrink.
-            if self._size == Size(0):
-                self._resizable = False
+        # We absolutely need a current size to enable resize. To shrink the
+        # filesystem we need a real minimum size provided by the resize
+        # tool. Failing that, we can default to the current size,
+        # effectively disabling shrink.
+        if self._size == Size(0):
+            self._resizable = False
 
         try:
             result = self._minsize.do_task()
--- a/blivet/tasks/fsck.py
+++ b/blivet/tasks/fsck.py
@@ -114,7 +114,9 @@ class Ext2FSCK(FSCK):
                     128: "Shared library error."}
 
     ext = availability.E2FSCK_APP
-    options = ["-f", "-p", "-C", "0"]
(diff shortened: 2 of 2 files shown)


AdamWill:
I don't think this really addresses what people care about in RHBZ 1170803. Avoiding modifying the filesystem is fine, but what people really *complain* about is that just running the check blocks the install process potentially for a very long time, especially when it runs on a disk you weren't planning to do anything with at all in the first place. We really need to stop running this check at all except when it's actually *needed* - before a resize operation.

vpodzime:
@AdamWill I understand that. But there's no other way to find out whether the file system is clean or not. And blivet needs that information to provide reliable data/info about the file system. I think we need some kind of fast, cheap and noninvasive check + indicate to user (in anaconda) that the file system needs a proper check so that they can tell if we should run it or not.

AdamWill:
"And blivet needs that information to provide reliable data/info about the file system."

Um. Does it? In the bug we came to the agreement (I thought) that the check was specifically added because in a new upstream release of e2fsprogs, they'd made it so `resize2fs -P` (which we use to estimate the minimum possible size of ext partitions) would not work unless the filesystem had been checked recently. Upstream agreed that that was a mistake and changed it.

The setup you describe would probably be neat if it were possible, but we've never had anything like it before and AFAIK this check was not added for any greater reason than '`resize2fs -P` needs it', and that is no longer true.

vpodzime:
@AdamWill My understanding is that because it's not possible to resize an unclean file system blivet needs to check it to decide whether to mark it resizable or not. Which is described in the sources:

```
# We can't allow resize if the filesystem has errors.
try:
    self.do_check()
except FSError:
    self._resizable = False
    raise
```

blivet/formats/fs.py
```diff
@@ -274,32 +274,25 @@ def update_size_info(self):
         self._min_instance_size = Size(0)
         self._resizable = self.__class__._resizable
 
-        # We can't allow resize if the filesystem has errors.
+        # try to gather current size info
```
mulkieran: Assuming this goes in, method comments in header should be updated to match code changes.

vpodzime: Good catch, thanks!

vpodzime: Fixed

AdamWill:
We need to do something with this...anyone?

vpodzime:
> We need to do something with this...anyone?

Actually, I just rebased the latest version of the changes to current *2.1-devel* and it looks like it fixes the issue(s) for me. Or am I missing something? I do admit I got lost a bit in all these things.

blivet/tasks/fsck.py
```diff
@@ -114,7 +114,9 @@ class Ext2FSCK(FSCK):
                     128: "Shared library error."}
 
     ext = availability.E2FSCK_APP
-    options = ["-f", "-p", "-C", "0"]
+    # "Force checking even if the file system seems clean." (we might get false results otherwise)
+    # + "Open the filesystem read-only, and assume an answer of `no' to all questions."
+    options = ["-f", "-n"]
```
dwlehman: You're also turning off the progress output. Why do that?

blivet/formats/fs.py
```diff
@@ -317,32 +315,25 @@ def update_size_info(self):
...
         try:
-            self.do_check()
-        except FSError:
+            if self._info.available:
+                self._current_info = self._info.do_task()
+        except FSError as e:
+            log.info("Failed to obtain info for device %s: %s", self.device, e)
+        try:
+            self._size = self._size_info.do_task()
+            self._min_instance_size = self._size
```
dwlehman: Seems a bit odd that setting `self._min_instance_size` is in the `try` block.

blivet/formats/fs.py
```diff
@@ -317,32 +315,25 @@ def update_size_info(self):
...
+        except (FSError, NotImplementedError) as e:
+            log.warning("Failed to obtain current size for device %s: %s", self.device, e)
+
+        # We absolutely need a current size to enable resize. To shrink the
```
dwlehman: I'm not completely sure this is true, but it's probably a good indicator of having the tools and the state required to resize.

blivet/tasks/fsck.py
vpodzime: What is the point of having it? In the logs...

blivet/formats/fs.py
vpodzime: Well, it should only be (re)set here if we are updating ``self._size``

blivet/formats/fs.py
vpodzime: Yeah, it's an old comment that I just moved.

vpodzime:
> We just talked about this, and I'm not sure it makes sense to have XFS._resizable default to False. Are there any conditions (that this code handles) under which we would determine that we cannot resize an existing XFS filesystem?

I'm not aware of any. XFS does a FS check when being mounted. We are not doing anything like that here. And we probably never mark XFS as resizable (or at least I'm not aware of any code possibly doing so).

blivet/tasks/fsck.py
dwlehman: Touche, sir.

blivet/formats/fs.py
dwlehman: I would have put it in an `else` block to isolate/identify the code that might raise an exception, but it's not important.

vpodzime: Yeah, that's a good idea.

PR #530, opened 2016-12-22
Shallow copy another alignment property (#1408282)

AdamWill:
Several blivet classes use custom __deepcopy__ methods to avoid
deep copying some parted objects which we know can't be deep
copied. Unfortunately the list of attributes excepted from deep
copying for `DiskLabel` was missing one parted Alignment object,
and with Python 3.6, we blow up trying to deepcopy that. This
fixes the problem by adding that attribute to the list of ones
we know have to be shallow copied.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -90,7 +90,8 @@ class DiskLabel(DeviceFormat):
             We can't do copy.deepcopy on parted objects, which is okay.
         """
         return util.variable_copy(self, memo,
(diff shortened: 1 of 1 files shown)


AdamWill:
FWIW I think https://github.com/rhinstaller/blivet/commit/93c805d74b6ee893e8f9cf59663207ffd45268c3 initially introduced this problem, because it changed the set of alignment attributes that `DiskLabel` instances have, and missed one of them out of the `shallow` list when updating that.

vojtechtrefny:
Jenkins, ok to test.

PR #535, opened 2017-01-05
Updated calls to avoid log spamming

japokorn:
Updated deprecated get_info calls that caused
excessive logging of warnings

### DIFF
--- a/blivet/populator/helpers/partition.py
+++ b/blivet/populator/helpers/partition.py
@@ -57,13 +57,13 @@ class PartitionDevicePopulator(DevicePopulator):
                 return device
 
         disk = None
-        sys_name = udev.device_get_partition_disk(self.data)
-        if sys_name:
-            disk_name = udev.resolve_devspec(sys_name)
+        disk_name = udev.device_get_partition_disk(self.data)
+        if disk_name:
             disk = self._devicetree.get_device_by_name(disk_name)
             if disk is None:
                 # create a device instance for the disk
-                disk_info = next((i for i in udev.get_devices() if i.sys_name == sys_name), None)
+                disk_info = next((i for i in udev.get_devices()
+                                  if udev.device_get_name(i) == disk_name), None)
                 if disk_info is not None:
                     self._devicetree.handle_device(disk_info)
                     disk = self._devicetree.get_device_by_name(disk_name)


M4rtinK:
Thanks a lot - this should hopefully make the Anaconda logs (more?) readable again. :)

vpodzime:
> Thanks a lot - this should hopefully make the Anaconda logs (more?) readable again. :)

LOL :)

blivet/populator/populator.py
```diff
@@ -118,7 +118,7 @@ def _udev_device_is_disk(self, info):
             udev/sysfs, we have to define what is a disk in terms of what is
             not a disk.
         """
-        return udev.device_is_realdisk(info)
+        return udev.device_is_disk(info)
```
vpodzime: This looks quite suspicious, please double-check it's no change in the behavior.

japokorn: You are right. Deprecation warning popping out is targeted to version 3.0. This part of code is to be left unchanged.

japokorn:
Jenkins, test this please.

japokorn:
Jenkins, test this please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Is this really done? I mean, are these the only places in the codebase where we use the deprecated API?

M4rtinK:
I guess I'll check with the branch in updates image. :)

M4rtinK:
So I've pulled the 2.1-devel-udev_log_spam branch to an updates image and it does seem like most of the udev warning spam is still there (log from a manual installation):
https://paste.fedoraproject.org/550675/86492662/

I *think* some of the warnings that did show up before are not there anymore, but the overall amount still has not really decreased & the log is still not very readable.

dwlehman:
My hunch is this will require a change everywhere we access `info` or `data` as a dict, whether by key lookup (eg: `info['ID_FS_TYPE']`) or by `dict.get` (eg: `info.get('ID_FS_TYPE', '')`). Most of them will be in `blivet/udev.py` but there will certainly be others sprinkled throughout `blivet/populator/`.

vpodzime:
> My hunch is this will require a change everywhere we access info or data as a dict, whether by key lookup (eg: info['ID_FS_TYPE']) or by dict.get (eg: info.get('ID_FS_TYPE', '')). Most of them will be in blivet/udev.py but there will certainly be others sprinkled throughout blivet/populator/.

Let me repeat my suggestion once more here -- why don't we return the dicts (``properties``) instead of original objects right when we iterate over them? We don't need the objects themselves anywhere, do we?

PR #540, opened 2017-02-03
Loop devices w/o backing file are now ignored

japokorn:
Populator now skips all detached loop devices (i.e. without backing file)


japokorn:
Jenkins, test this please.

PR #541, opened 2017-02-03
Loop devices w/o backing file are now ignored

japokorn:
Populator now skips all detached loop devices (i.e. without backing file)

### DIFF
--- a/blivet/populator/helpers/disk.py
+++ b/blivet/populator/helpers/disk.py
@@ -46,6 +46,7 @@ class DiskDevicePopulator(DevicePopulator):
         return (udev.device_is_disk(data) and
                 not udev.device_is_cdrom(data) and
                 not udev.device_is_partition(data) and
+                not udev.device_is_loop(data) and
                 not udev.device_is_dm(data) and
                 not (udev.device_is_md(data) and not udev.device_get_md_container(data)))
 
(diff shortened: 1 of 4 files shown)


dwlehman:
I have a couple of other ideas for how this could be resolved. I'm not convinced any one is better than the others (or yours), but I want to list them anyway:

1. Change `udev.device_is_loop` to only check the name, adding an additional check for the `loop/` subdir or backing file in the loop device handler. This would prevent detached loop devices from being treated like disks, which is the behavior I saw that led to this work.
2. Your approach, but checking for the `loop/` subdir instead of calling the libblockdev function to check for a backing file. I don't really have a preference here, just stating alternatives. Checking for the dir is cheaper, but it also probably provides less certainty what the situation is.
3. Adding something to the disk device populator helper to prevent loop devices from being handled there.

vpodzime:
> Your approach, but checking for the loop/ subdir instead of calling the libblockdev function to check for a backing file. I don't really have a preference here, just stating alternatives. Checking for the dir is cheaper, but it also probably provides less certainty what the situation is.

This is what libblockdev does - https://github.com/rhinstaller/libblockdev/blob/master/src/plugins/loop.c#L100 I don't think checking for dir existence from python is cheaper. :)

japokorn:
After further examination I have discovered that the check is already implemented but not working correctly.
Loop device was being skipped but recognized later as a disk device by check in DiskDevicePopulator.match().
By updating the check, detached loop device no longer makes it into devicetree. All previous changes were reverted.

japokorn:
Jenkins, test this please.

japokorn:
Implemented requested changes and updated tests accordingly.

japokorn:
pushed

PR #542, opened 2017-02-07
Consider non-efi AARCH64 machine to behave like am ARM machine

superna9999:
If an non-efi AARCH64 machine le booted, the system architecture fails to be determined.

Consider a non-efi AARCH64 to be like a non-efi ARM machine.

### DIFF
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -468,6 +468,8 @@ def get_platform():
             return omapARM()
         else:
             return ARM()
+    elif arch.is_aarch64():
+        return ARM()
     else:
         raise SystemError("Could not determine system architecture.")
 


blivet/platform.py
```diff
@@ -468,6 +468,8 @@ def get_platform():
             return omapARM()
         else:
             return ARM()
+    elif arch.is_aarch64():
+        return ARM()
```
vpodzime: Please create a new ``Aarch64NonEFI`` (or a better name) subclass of the ``ARM`` class and use that here. And don't forget to add a descriptive docstring to that new class explaining why it's just a subclass of ``ARM``.

vpodzime:
Jenkins, this is ok to test.

pholica:
I've hit this when I run virt-install without --boot uefi.

vpodzime:
@superna9999, could you please update the PR?

blivetghoul:

vpodzime:
I guess this should be just closed now?

vojtechtrefny:
> I guess this should be just closed now?

Yes, I agree. @superna9999 feel free to reopen this PR and rebase to latest 3.3-devel if you think this is still relevant.

PR #546, opened 2017-02-22
FBA DASD should use the msdos disk label type (#1214407)

poncovka:
There can be created up to three partitions on FBA DASD with the msdos
disk label type instead of only one, so we should prefer that.

Resolves: rhbz#1214407

### DIFF
--- a/blivet/devicelibs/dasd.py
+++ b/blivet/devicelibs/dasd.py
@@ -144,6 +144,30 @@ def is_ldl_dasd(device):
         return False
     return True
 
+def is_fba_dasd(device):
+    """Is the device a FBA DASD?"""
+    if not device.startswith("dasd"):
+        # not a dasd; bail
+        return False
+
+    device = "/dev/%s" % (device,)
+    f = open(device, "r")
+
+    # poorly check if this is even a block device...
+    arg = blksize()
+    rc = fcntl.ioctl(f.fileno(), BLKSSZGET, arg, True)
+    if rc < 0:
+        return False
+
+    # alright, it's a block device, so get some info about DASD...
+    arg = dasd_info()
+    rc = fcntl.ioctl(f.fileno(), BIODASDINFO2, arg)
+    if rc < 0:
+        return False
+
+    # check if we're on an FBA DASD
+    return arg.type.startswith('FBA')
+
 def get_dasd_ports():
     """ Return comma delimited string of valid DASD ports. """
     ports = []
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -117,36 +117,26 @@ class Platform(object):
              "descriptions": {k: _(v) for k, v in self._boot_descriptions.items()}}
         return d
 
-    def requiredDiskLabelType(self, device_type):
-        # pylint: disable=unused-argument
-        return None
-
     def bestDiskLabelType(self, device):
         """The best disklabel type for the specified device."""
         if flags.testing:
             return self.defaultDiskLabelType
 
         parted_device = parted.Device(path=device.path)
+        label_type = self.defaultDiskLabelType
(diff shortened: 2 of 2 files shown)


blivet/platform.py
```diff
@@ -351,12 +352,20 @@ def setPlatformBootPartition(self):
         return [PartSpec(mountpoint="/boot", size=Size("1GiB"),
                          weight=self.weight(mountpoint="/boot"), lv=False)]
 
-    def requiredDiskLabelType(self, device_type):
-        """The required disklabel type for the specified device type."""
-        if device_type == parted.DEVICE_DASD:
```
dwlehman: Since you're removing the only reason for `requiredDiskLabelType` to exist, please remove it completely -- thanks for removing an old platform-specific hack!

poncovka: Fixed.

poncovka:
I have the pull request for `2.1-devel` ready here https://github.com/poncovka/blivet/commits/2.1-devel-fba, but it requires some changes in `libblockdev` that are here https://github.com/poncovka/libblockdev/commit/e48944e8f89151b6aca91c098d026df999a9d151. However, the testing is more complicated in this case, so it will take some time to create pull requests.

blivet/devicelibs/dasd.py
```diff
@@ -144,6 +144,30 @@ def is_ldl_dasd(device):
         return False
     return True
 
+def is_fba_dasd(device):
+    """Is the device a FBA DASD?"""
+    if not device.startswith("dasd"):
+        # not a dasd; bail
+        return False
+
+    device = "/dev/%s" % (device,)
```
dwlehman: I know it will not matter in this case, but I would prefer if you passed in the full path instead of the name so the device class can handle building the full path.

blivet/devicelibs/dasd.py
```diff
@@ -144,6 +144,30 @@ def is_ldl_dasd(device):
...
+    # poorly check if this is even a block device...
+    arg = blksize()
+    rc = fcntl.ioctl(f.fileno(), BLKSSZGET, arg, True)
+    if rc < 0:
+        return False
+
+    # alright, it's a block device, so get some info about DASD...
+    arg = dasd_info()
+    rc = fcntl.ioctl(f.fileno(), BIODASDINFO2, arg)
+    if rc < 0:
```
dwlehman: Is there any likelihood that `open` or either of those `ioctl`s will raise an exception?

blivet/devicelibs/dasd.py
poncovka: The `is_fba_dasd` function is copy-pasted from `is_ldl_dasd`. I could change  `is_fba_dasd`, but then `is_ldl_dasd` should be changed as well. However, I would prefer not to do that since it is for RHEL and s390x is not so easy to test. 

Nevertheless, I can do that in a pull request for `2.1-devel` branch (https://github.com/rhinstaller/blivet/pull/552), since the `bd_s390_dasd_is_fba` function is already prepared for that (https://github.com/rhinstaller/libblockdev/pull/170). I should just call `blockdev.s390.dasd_is_fba(device.path)`, right?

What do you think?

blivet/devicelibs/dasd.py
poncovka: We never get exceptions from `is_ldl_dasd` in anaconda. And `is_fba_dasd` is copy-pasted from it. I can change the function to catch the exceptions, but see the comment above.

blivet/devicelibs/dasd.py
dwlehman: Fixing it on `2.1-devel` while minimizing changes for `rhel7-branch` sounds wise. (Also, you are correct about using `device.path`.)

poncovka: Fixed on `2.1-devel`.

PR #547, opened 2017-02-22
Add HW RAID properties to DiskDevice

dwlehman:
This adds a few properties to `DiskDevice` describing RAID characteristics where applicable. The data is taken from [`lsm.Client.volume_raid_info`](https://libstorage.github.io/libstoragemgmt-doc/doc/py_api_user_guide.html#query-volume-raid-information----lsmclientvolumeraidinfo) and converted to blivet data types.

A useful followup would be to use `DiskDevice.raid_system` as `DiskDevice.description` when available since it should convey much more information than the existing method. This would be a freebie for anaconda usability.

### DIFF
--- a/blivet/devicelibs/disk.py
+++ b/blivet/devicelibs/disk.py
@@ -1,0 +1,111 @@
+from collections import namedtuple
+
+from . import raid
+from .. import errors
+from .. import util
+from ..size import Size
+
+lsm = None
+
+_HBA_PLUGIN_URIS = ("hpsa://", "megaraid://")
+LSMInfo = namedtuple('HBAVolumeInfo', ['system', 'nodes', 'raid_type', 'raid_stripe_size', 'raid_disk_count'])
+""" .. class:: LSMInfo
+
+        .. attribute:: system (str): descriptive name of HBA unit
+        .. attribute:: nodes (list[str]): list of device node paths for the volume
+        .. attribute:: raid_type (:class:`~.devicelibs.raid.RAIDLevel` or None): RAID level
+        .. attribute:: raid_stripe_size (:class:`~.size.Size` or None): stripe size
+        .. attribute:: raid_disk_count (int or None): number of disks in the RAID set
+"""
+volumes = dict()
+_raid_levels = dict()
+
+
+class _LSMRAIDLevelStub(raid.RAIDLevel):
+    def __init__(self, name):
+        self._name = name
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def names(self):
+        return [self.name]
+
+    @property
+    def min_members(self):
+        return 0
+
+    def has_redundancy(self):
+        return False
+
+    def is_uniform(self):
+        return False
+
+
+class _LSMDependencyGuard(util.DependencyGuard):
+    error_msg = "libstoragemgmt functionality not available"
+
+    def _check_avail(self):
+        global lsm
+        if lsm is None:
+            try:
+                import lsm  # pylint: disable=redefined-outer-name
+            except ImportError:
+                lsm = None
+
+        return lsm is not None
+
+_lsm_required = _LSMDependencyGuard()
+
+
+def _update_lsm_raid_levels():
+    """ Build a mapping of lsm.RAID_TYPE->blivet.devicelibs.raid.RAIDLevel """
+    global _raid_levels
+    _raid_levels = dict()
+    lsm_raid_levels = dict((k, v) for (k, v) in lsm.Volume.__dict__.items() if k.startswith("RAID_TYPE_"))
+    for constant_name, value in lsm_raid_levels.items():
+        name = constant_name[len("RAID_TYPE_"):]
+        try:
+            level = raid.get_raid_level(name)
+        except errors.RaidError:
+            level = _LSMRAIDLevelStub(name)
+
+        _raid_levels[value] = level
+
+
+def _get_lsm_raid_level(lsm_raid_type):
+    """ Return a blivet.devicelibs.raid.RAIDLevel corresponding the lsm-reported RAID level."""
+    return _raid_levels.get(lsm_raid_type, _raid_levels.get(lsm.Volume.RAID_TYPE_UNKNOWN))
+
+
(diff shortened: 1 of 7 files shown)


blivet/devicelibs/disk.py
```diff
@@ -0,0 +1,110 @@
...
+        return False
+
+
+class _LSMDependencyGuard(util.DependencyGuard):
+    error_msg = "libstoragemgmt functionality not available"
+
+    def _check_avail(self):
+        global lsm
+        try:
+            import lsm  # pylint: disable=redefined-outer-name
```
dwlehman: It is provided by `libstoragemgmt-python`. I am trying to avoid adding any hard dependencies, and made an effort to arrange things to function without failure whether or not libstoragemgmt is installed (or its service running). I am open to argument, of course, if you think it should be a hard requirement for all blivet users.

dwlehman: You're correct that the patch supports MegaRAID and SmartArray only. There is a pull request that would add a generic interface for local HBA to `lsm` [here](https://github.com/libstorage/libstoragemgmt/issues/238), although that will not necessarily change the supported hardware list.

nnDarshan: Not having a hard dependency makes sense, as not all users will need raid details.

vpodzime: Shouldn't this first test ``if lsm is None:``? I guess the import is more or less a no-op too if run multiple times, but not sure.

blivet/devicelibs/disk.py
```diff
@@ -0,0 +1,110 @@
...
+        return lsm is not None
+
+_lsm_required = _LSMDependencyGuard()
+
+
+def _update_lsm_raid_levels():
+    """ Build a mapping of lsm.RAID_TYPE->blivet.devicelibs.raid.RAIDLevel """
+    global _raid_levels
+    _raid_levels = dict()
+    lsm_raid_levels = dict((k, v) for (k, v) in lsm.Volume.__dict__.items() if k.startswith("RAID_TYPE_"))
```
vpodzime: I believe this can be written as ``{k: v for (k, v) in...}``

blivet/devicelibs/disk.py
```diff
@@ -0,0 +1,110 @@
...
+            stripe_size = None
+            disk_count = None
+            if caps.supported(lsm.Capabilities.VOLUME_RAID_INFO):
+                raid_info = client.volume_raid_info(vol)[:3]
+                raid_level = _get_lsm_raid_level(raid_info[0])
+                stripe_size = Size(raid_info[1])
+                disk_count = raid_info[2]
+
+            info = LSMInfo(system.name, nodes, raid_level, stripe_size, disk_count)
+            volumes.update(dict((node, info) for node in nodes))
```
vpodzime: Same here with the "dict-comprehension"

blivet/devicelibs/disk.py
dwlehman: Yeah, it probably should check first.

tests/util_test.py
```diff
@@ -113,3 +115,35 @@ def test_requires_property(self):
...
+        with self.assertLogs("blivet", level="WARNING") as cm:
+            self.assertEqual(self._test_dependency_guard_non_critical(), None)
+        self.assertTrue(TestDependencyGuard.error_msg in "\n".join(cm.output))
+
+        with self.assertRaises(errors.DependencyError):
+            self._test_dependency_guard_critical()
+
+        with unittest.mock.patch.object(_requires_something, '_check_avail', return_value=True):
+            self.assertEqual(self._test_dependency_guard_non_critical(), True)
+            self.assertEqual(self._test_dependency_guard_critical(), True)
```
dwlehman: I've become a big fan of `mock` lately -- unit testing in python is __much__ harder without it.

blivet/devicelibs/disk.py
dwlehman: Yeah, I've just become fond of the explicit constructor calls -- especially for `dict` and `list`. I think when they introduced the `set` initialization syntax that looked like `dict`'s it made me want to be more explicit.

tests/devicelibs_test/disk_test.py
```diff
@@ -0,0 +1,91 @@
...
+        with patch("blivet.devicelibs.disk._lsm_required._check_avail", return_value=True):
+            with patch("blivet.devicelibs.disk.lsm") as _lsm:
+                _lsm.Volume.RAID_TYPE_RAID0 = sentinel.RAID_TYPE_RAID0
+                _lsm.Volume.RAID_TYPE_OTHER = sentinel.RAID_TYPE_OTHER
+                _lsm.Capabilities.VOLUME_RAID_INFO = sentinel.VOLUME_RAID_INFO
+                _lsm.Client.capabilities.side_effect = client_capabilities
+                _lsm.Client.volumes.return_value = lsm_volumes
+                disklib.update_volume_info()
+                for lvol, bvol in zip(lsm_volumes, disklib.volumes):
+                    self.assertEqual(bvol.system, _lvol[0])
```
tasleson: I believe `_lvol` should be `lvol`

tests/devicelibs_test/disk_test.py
```diff
@@ -0,0 +1,91 @@
...
+                    if client_capabilities(Mock(), bvol.system):
```
tasleson: I believe you are missing import for `Mock`, e.g.  `from unittest.mock import patch, sentinel, Mock, Size`

tests/devicelibs_test/disk_test.py
```diff
@@ -0,0 +1,91 @@
...
+                        self.assertEqual(bvol.raid_level, disklib._get_lsm_raid_level(lvol[1]))
+                        self.assertEqual(bvol.stripe_size, Size(lvol[2]))
```
tasleson: I believe you are missing: `from blivet.size import Size` for 'Size` here

tasleson:
@dwlehman LGTM

dwlehman:
#564 is this pull request rebased onto `2.2-devel`. Closing this since it shouldn't go onto `2.1-devel`.

PR #552, opened 2017-03-09
FBA DASD should use the msdos disk label type

poncovka:
There can be created up to three partitions on FBA DASD with the msdos
disk label type instead of only one, so we should prefer that.

Also the method requiredDiskLabelType can be removed, because it is not
needed anymore.

This pull request is based on https://github.com/rhinstaller/blivet/pull/546 and requires a new version of libblockdev 
https://github.com/rhinstaller/libblockdev/pull/170.

### DIFF
--- a/blivet/platform.py
+++ b/blivet/platform.py
@@ -123,34 +123,24 @@ class Platform(object):
              "descriptions": dict((k, _(v)) for k, v in self._boot_descriptions.items())}
         return d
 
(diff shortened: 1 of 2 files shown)


vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

PR #553, opened 2017-03-13
Introduce the discard_new flag and use it

vpodzime:
Instead of the old non-specific 'installer_mode' flag we no
longer have on 3.0-devel and master.


vpodzime:
This is a fixup for changes in PR #550 which are incompatible with *3.0-devel* and *master* branches.

vpodzime:
Jenkins, test this, please.

PR #555, opened 2017-03-14
Fix detection of linear MD RAID

vojtechtrefny:
Resolves: rhbz#1372414


vojtechtrefny:
Commit message updated.

PR #565, opened 2017-03-28
Ignore disks labeled OEMDRV in autopartitioning.

rvykydal:
Resolves: rhbz#1412022

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -202,6 +202,11 @@ def storageInitialize(storage, ksdata, protected):
 
     # kickstart uses all the disks
     if flags.automated_install:
+        # ignore OEMDRV disk
+        oemdrv = storage.devicetree.resolveDevice("LABEL=OEMDRV")
+        if oemdrv and not oemdrv in ksdata.ignoredisk.ignoredisk:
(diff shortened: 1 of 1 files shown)


blivet/__init__.py
```diff
@@ -202,6 +202,11 @@ def storageInitialize(storage, ksdata, protected):
 
     # kickstart uses all the disks
     if flags.automated_install:
+        # ignore OEMDRV disk
```
vpodzime: Why only on automated installs?

rvykydal: I think because it resolves reporters / customer's case. Do you think we should apply it to all installs? If so, to me it seems that just moving the patched code before the automated_install if block should do the job then.

vpodzime: Yeah, i think we should be always doing this.

rvykydal:
We should set ignored disks already in anaconda before initializing storage.
The patch doesn't work for multiple reasons (https://bugzilla.redhat.com/show_bug.cgi?id=1412022#c11).
Closing this PR and opening https://github.com/rhinstaller/anaconda/pull/1009

PR #566, opened 2017-04-03
Reserve space for thin pool in the VG not the thin pool itself

vpodzime:
As described in the first commit (788d7d0) and in the [bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1131247)

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -33,6 +33,7 @@ from .devicelibs import raid
 from .partitioning import SameSizeSet
 from .partitioning import TotalSizeSet
 from .partitioning import do_partitioning
+from .autopart import AUTOPART_THPOOL_RESERVE
 from .size import Size
 from .static_data import luks_data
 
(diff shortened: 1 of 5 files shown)


blivet/devices/lvm.py
```diff
@@ -1485,19 +1512,30 @@ def lvs(self):
...
+            self._chunk_size = Size(blockdev.LVM_DEFAULT_CHUNK_SIZE)
+            log.debug("Using default chunk size: %s", self._chunk_size)
+
+        self._metadata_size = Size(blockdev.lvm.get_thpool_meta_size(self._size,
+                                                                     self._chunk_size,
+                                                                     100))  # snapshots
+        log.debug("Recommended metadata size: %s", self._metadata_size)
+
+        log.debug("Adjusting size from %s to %s", self.size, self.size - self._metadata_size)
+        self.size = self.size - self._metadata_size
```
dwlehman: The idea here being that taking metadata space from the requested lv has less potential to cause a difficult situation since users might, for example, expect to be able to allocate lvs whose combined size equals the total vg size, right?

blivet/devicefactory.py
```diff
@@ -1482,21 +1479,15 @@ def _get_total_space(self):
...
-                pad = Size(blockdev.lvm.get_thpool_padding(self.device.size, self._pesize))
-                log.debug("old device size: %s ; old pad: %s", self.device.size, pad)
-                size -= pad
-                log.debug("size cut to %s to omit old device padding", size)
+        # black maths (to make sure there's AUTOPART_THPOOL_RESERVE.percent reserve)
+        size_with_reserve = size * (1 / (1 - (AUTOPART_THPOOL_RESERVE.percent / 100)))
+        if size_with_reserve - size > AUTOPART_THPOOL_RESERVE.min:
+            size = size_with_reserve
+        else:
+            size = size + AUTOPART_THPOOL_RESERVE.min
```
dwlehman: This is basically mimicking logic in `autoset_md_size` so we can correctly anticipate its results?

blivet/devices/lvm.py
vpodzime: Exactly. And also to make the code for growing a thin pool simpler because the metadata size depends on (is a function of) the thin pool's size. This way we may at worse end up creating a slightly over-provisioned thin pool, but that shouldn't be a problem, especially with that big reserved space in the VG.

blivet/devicefactory.py
vpodzime: You mean the logic of ``thpool_reserve``? Ideally it should be the VG saying *"I want this much more"*, but that would be (is going to be) quite a big change of logic which I'm trying to avoid here. So this just instead tries to be clever enough to give the VG enough space for it to have it's reserved space.

And now that I see it, it should also reflect ``AUTOPART_THPOOL_RESERVE.max`` because it may quite easily happen that somebody will use device factory for a VG bigger than 500 GiB.

dwlehman:
Jenkins, test this please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
Jenkins, test this, please.

vpodzime:
> Looks okay to me once the temporary commit is replaced by a proper change in libblockdev.

Done.

PR #567, opened 2017-04-03
Fix unit tests run as root (rhel7-branch)

dwlehman:


dwlehman:
Jenkins, please test this.

dwlehman:
Jenkins, test this please.

dwlehman:
I enabled testing as root prior to triggering the ci run of record.

tests/devicelibs_test/btrfs_test.py
```diff
@@ -164,19 +168,5 @@ def testSubvolume(self):
...
-
-    def testSmallDevice(self):
-        """ Creation of a smallish device will result in an error if the
-            data and metadata levels are specified differently, but not if
-            they are unspecified.
-        """
-        with self.assertRaises(BTRFSError):
-            btrfs.create_volume(self.loopDevices, data="single", metadata="dup")
-        self.assertEqual(btrfs.create_volume(self.loopDevices), 0)
-
```
vpodzime: Do we really want to just drop this test?

blivet/devicelibs/mdraid.py
```diff
@@ -343,6 +343,9 @@ def mddetail(device):
 
     info = {}
     for (name, colon, value) in (line.strip().partition(":") for line in lines):
+        if name and name[0] not in string.letters:
+            continue
```
vpodzime: Maybe an example of a bad line could be useful here.

tests/formats_test/selinux_test.py
```diff
@@ -33,10 +36,9 @@ def testMountingExt2FS(self):
         an_fs.unmount()
         os.rmdir(mountpoint)
 
-        self.assertEqual(root_selinux_context[1], 'system_u:object_r:file_t:s0')
+        self.assertEqual(root_selinux_context[1], DEFAULT_CONTEXT)
```
vpodzime: So what is this test actually trying to verify?

dwlehman: It's useless. At some point I think it was expected that you'd get a different context while the thing was mounted, but only if `flags.installer_mode` was True. Now it appears that selinux doesn't care about this mountpoint at all. There are two clear options:

1. remove this test completely
2. rewrite the test to use mock to directly verify that we're calling the appropriate selinux function where appropriate (bearing in mind that python2-mock is an epel package)

tests/devicelibs_test/btrfs_test.py
dwlehman: I played with it for a bit and couldn't find a size that was right on the line like the test expects and btrfs is not worth the time it would take at this point.

blivet/devicelibs/mdraid.py
dwlehman: Good idea -- will do.

tests/devicelibs_test/btrfs_test.py
vpodzime: Fair enough. :)

tests/formats_test/selinux_test.py
vpodzime: I think we can get rid of this test.

blivet/devicelibs/mdraid.py
dwlehman: In doing the legwork to reproduce and thus obtain the sample output, it occurred to me that I could make a much "simpler" _change_ to the parsing code instead of adding this hack. While this sounds nice, it is a change on `rhel7-branch`, so I'm not sure the risk is worth taking for the sake of nicer code.

Here's what it looks like (diff from this pull request branch):

```patch
diff --git a/blivet/devicelibs/mdraid.py b/blivet/devicelibs/mdraid.py
index 33c1ec0..4906719 100644
--- a/blivet/devicelibs/mdraid.py
+++ b/blivet/devicelibs/mdraid.py
@@ -342,10 +342,7 @@ def mddetail(device):
         raise MDRaidError("mddetail failed for %s: %s" % (device, e))
 
     info = {}
-    for (name, colon, value) in (line.strip().partition(":") for line in lines):
-        if name and name[0] not in string.letters:
-            continue
-
+    for (name, colon, value) in (line.strip().partition(" : ") for line in lines):
         value = value.strip()
         name = name.strip().upper()
         if colon and value and name:
```

Opinions?

vpodzime: Looks good to me. I don't we are interested in any other values than those separated by ``" : "``.

dwlehman:
Jenkins, test this please.

dwlehman:
Jenkins, test this please.

PR #568, opened 2017-04-03
Fix unit tests that only run as root

dwlehman:
This is the only fix needed to be able to run unit tests as root on `3.0-devel` and `master`.


dwlehman:
Jenkins, test this please.

dwlehman:
The test run of record ran `sudo make pylint pep8 test` (and passed).

PR #569, opened 2017-04-03
Fix root-only unit tests on 2.1-devel

dwlehman:
Apparently pointing blivet's tests at an uncompiled anaconda git clone is not the same as having the anaconda-core package installed -- go figure.

### DIFF
--- a/tests/formats_test/luks_test.py
+++ b/tests/formats_test/luks_test.py
@@ -49,6 +49,10 @@ class LUKSTestCase(loopbackedtestcase.LoopBackedTestCase):
         self.fmt.update_size_info()
         self.assertEqual(self.fmt.current_size, new_size)
 
+    def tearDown(self):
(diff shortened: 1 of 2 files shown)


tests/formats_test/luks_test.py
```diff
@@ -49,6 +49,10 @@ def test_resize(self):
         self.fmt.update_size_info()
         self.assertEqual(self.fmt.current_size, new_size)
 
+    def tearDown(self):
+        self.fmt.teardown()
+        super(LUKSTestCase, self).tearDown()
```
vpodzime: This should ideally be added with ``add_cleanup()``, but that's a general issue of these tests so I'm okay with being consistent here.

dwlehman: Good point, though. I'll something to the backlog to fix it properly.

PR #570, opened 2017-04-06
Reserve space for thin pool(s) to grow in a VG

vpodzime:
This is a *rhel7-branch* version of the PR #566 addressing [rhbz#1131247](https://bugzilla.redhat.com/show_bug.cgi?id=1131247)

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -25,6 +25,7 @@ import copy
 import pprint
 import re
 import time
+from collections import namedtuple
 
 # device backend modules
 from ..devicelibs import lvm
@@ -45,6 +46,9 @@ from .storage import StorageDevice
 from .container import ContainerDevice
 from .dm import DMDevice
 from .md import MDRaidArrayDevice
+
+ThPoolReserveSpec = namedtuple("ThPoolReserveSpec", ["percent", "min", "max"])
+""" A namedtuple class for specifying restrictions of space reserved for a thin pool to grow """
 
 class LVMVolumeGroupDevice(ContainerDevice):
     """ An LVM Volume Group """
@@ -102,6 +106,7 @@ class LVMVolumeGroupDevice(ContainerDevice):
         self.peFree = util.numeric_type(peFree)
         self.reserved_percent = 0
         self.reserved_space = Size(0)
+        self._thpool_reserve = None
 
         # this will have to be covered by the 20% pad for non-existent pools
         self.poolMetaData = 0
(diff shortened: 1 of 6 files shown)


vpodzime:
I still need to adapt the unittests.

vpodzime:
> I still need to adapt the unittests.

Or not. :) I forgot basically all the devicefactory tests are missing on the rhel7-branch. But at least I'm obviously not breaking anything else.

blivet/devices/storage.py
```diff
@@ -243,7 +243,7 @@ def _setTargetSize(self, newsize):
         if not isinstance(newsize, Size):
             raise ValueError("new size must of type Size")
 
-        if self.maxSize and newsize > self.maxSize:
+        if self.maxSize and newsize and newsize > self.maxSize:
```
dwlehman: When does the factory set a device's size to 0? My recollection is that the factory's code for adjusting a container to deal w/ a removed device runs _after_ the device is removed.

blivet/devices/lvm.py
```diff
@@ -1156,11 +1175,17 @@ def __init__(self, name, parents=None, size=None, uuid=None,
                                                 percent=percent,
                                                 segType=segType)
 
-        self.metaDataSize = metadatasize or 0
         self.chunkSize = chunksize or 0
         self.profile = profile
         self._lvs = []
 
+        if metadatasize:
+            self.metaDataSize = metadatasize or 0
```
vpodzime: Actually, it should be moved before the ``if`` block to make sure ``self.metaDataSize`` is always set to something when ``__init__()`` finishes. Fixing.

blivet/devices/storage.py
vpodzime: This happens when the (last) thin LV is removed and the thin pool's size is adapted as part of ``_reconfigure_container()``. The calculated size for thin pool is ``0`` because it has no thin LVs and it now doesn't require any padding.

dwlehman: Hmm, okay. Does the newly-empty thin-pool get removed eventually? If so, from where?

vpodzime: Anaconda then calls this:
```
    def _remove_empty_parents(self, device):
        # if this device has parents with no other children, remove them too
        for parent in device.parents:
            if parent.kids == 0 and not parent.isDisk:
                self._destroy_device(parent)
```
That looks like a quite twisted logic/thing to me, but I'm not changing this in any way.

PR #571, opened 2017-04-06
Add a WWN attribute to disk-like device classes.

dwlehman:
One tricky bit is that I'm removing the `serial` constructor argument to `MultipathDevice`, which has until now been used as a stand-in for WWN. The `MultipathDevice` constructor is not listed as part of the public API, so I think this should be okay.

It's too bad that `DMRaidArrayDevice` and `MultipathDevice` do not derive from `DiskDevice` -- it would have made this (and #485) much simpler.

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -62,7 +62,7 @@ class DiskDevice(StorageDevice):
 
     def __init__(self, name, fmt=None,
                  size=None, major=None, minor=None, sysfs_path='',
-                 parents=None, serial=None, vendor="", model="", bus="",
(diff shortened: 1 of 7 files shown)


dwlehman:
Jenkins, test this please.

dwlehman:
Jenkins, test this please.

dwlehman:
Jenkins, test this please.

dwlehman:
Latest update added `DiskDevice.wwn` to the public API.

vpodzime:
> Latest update added DiskDevice.wwn to the public API.

Good point. Looks even better to me now.

PR #537, opened 2017-01-14
WIP: Allow setting file system UUIDs

aszlig:
Implements setting UUIDs via passing `new_uuid="..."` to the format.

This is similar to setting labels, but while labels are set by just passing the `label=something` argument, this uses `new_uuid=something` because first of all, an UUID is not optional and we also have an `uuid` argument that's already used by the device tree populator to pass the UUID from the udev attributes.

File systems that support setting UUIDs are now ext2/3/4, FAT, NTFS, XFS, JFS, ReiserFS, HFS+ and swap. Not supported is HFS, because it doesn't seem to have a notion for setting serial or any kind of volume ID.

PR #573, opened 2017-04-12
Fixes for thin snapshot creation

vpodzime:
These are fixes and enhancements required for valid (thin) snapshot creation, similar to PR #563, but this time for *2.1-devel* and other devel branches. It contains patches from the PR #537 for file system UUID configuration in their slightly modified versions incorporating the comments from the review process and with post-creation UUID configuration for FAT file systems removed because it seemed to be not working at all and it's not a big deal anyway.


vpodzime:
@aszlig, please take a look at these.

PR #576, opened 2017-04-26
Use structured logging for the anaconda logger

M4rtinK:
Otherwise all Blivet log messages in the Anaconda log messages
would have the "misc" prefix.
With "anaconda.blivet.autopart" the prefix will be "blivet.autopart".


vpodzime:
> Looks fine to me.

Do we really want to (unconditionally) use the anaconda logger(s) in blivet's autopartitioning code?

vpodzime:
This is definitely an improvement for the current situation.

PR #578, opened 2017-04-28
Fix resolve_devspec to fully support raid devices (#1445723)

poncovka:
Return md name if it matches a device name for the given device
specification, so the device names are consistent when we work
with udev module and a devicetree.

Resolves: rhbz#1445723


dwlehman:
Jenkins, test this please.

PR #579, opened 2017-05-01
Update md boot metadata version selection.

dwlehman:
This is a backport from master and has been verified to resolve the referenced bug. I don't see any risk, given that the behavior has not caused problems in recent fedora and the reporter verified the fix.


dwlehman:
Jenkins, test this please.

PR #580, opened 2017-05-03
Add RAID chunk size to the generated kickstart file

vojtechtrefny:
Resolves: rhbz#1447343


vojtechtrefny:
> Is it possible to only write the chunk size if/when the user specified a value for it?

The problem is with the "default" value for the chunk size -- if the chunk size is 512 KiB (`mdraid.MD_CHUNK_SIZE`) we don't know if this is because user wants that or if it is our default value.

I can add some flag/variable to check if user actually set the chunk size manually.

PR #581, opened 2017-05-03
Add RAID chunk size to the generated kickstart file

vojtechtrefny:
#580 for 2.1-devel

This also includes the original chunks size support commit, because it isn't (for some unknown reason) part of this branch -- it is on "newer" branches (2.2-devel, 3.0-devel and master) and also on the 2.1-release branch (that's why no one noticed) but not here.


vojtechtrefny:
Added new commit to fix pylint warnings -- this is a bug in pylint/astroid (see https://github.com/PyCQA/pylint/issues/1282) and we don't have the newest version on Fedora.

PR #582, opened 2017-05-03
Respect thin pool's min size when setting its req_size

vpodzime:
When setting thin pool's size based on the sum of its thin LVs we
need respect its minimum size. That can for example be its actual
size if it already exists.

Together with the recent changes for thin pool padding this

Resolves: rhbz#1435180


vpodzime:
This also fixes https://bugzilla.redhat.com/show_bug.cgi?id=1449963

PR #584, opened 2017-05-04
Do not remove manually created extended partitions (#1440150)

vojtechtrefny:
In installer mode do not try to remove empty extended partitions
created by blivet-gui. We should remove the extended partition
only if the PartitionDevice doesn't exist or if we previously
deleted all the logical partitions.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -330,6 +330,8 @@ def remove_new_partitions(disks, remove, all_partitions):
     log.debug("removing all non-preexisting partitions %s from disk(s) %s",
               ["%s(id %d)" % (p.name, p.id) for p in remove],
               [d.name for d in disks])
+
+    removed_logical = []
     for part in remove:
         if part.parted_partition and part.disk in disks:
             if part.exists:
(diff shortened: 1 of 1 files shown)


blivet/partitioning.py
```diff
@@ -340,16 +342,38 @@ def remove_new_partitions(disks, remove, all_partitions):
...
+                        return True
+                    else:
+                        return False
+
     for disk in disks:
         # remove empty extended so it doesn't interfere
         extended = disk.format.extended_partition
-        if extended and not disk.format.logical_partitions and \
-           (flags.installer_mode or
-                extended not in (p.parted_partition for p in all_partitions)):
```
dwlehman: What if the only change was to remove the `flags.installer_mode or` from the original code?

vojtechtrefny: That would definitely help. I wanted to keep as much of the original functionality as possible, because it we may want to delete the extended partition if `remove_new_partitions` actually removes the logical partitions.

dwlehman: Right. In the case where the extended was created automatically by blivet there will not be a corresponding `PartitionDevice` instance, so that should be covered by `extended not in (p.parted_partition for p in all_partitions)`.

PR #585, opened 2017-05-12
Updated calls to avoid log spamming

japokorn:
Fix of excessive deprecation warnings in the logs.

Due to nature of deprecation warnings origin (e.g. __getitem__ or
__iter__ methods) it was almost impossible to find and fix all of its
sources. Instead the Device class was substituted with its 'properties'
attribute which directly fixes the problem.
Couple of other minor changes was neccessary, though.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -24,36 +24,56 @@
 import os
 import re
 import subprocess
+import logging
+import pyudev
 
 from . import util
 from .size import Size
 from .flags import flags
 
-import pyudev
 global_udev = pyudev.Context()
-
-import logging
 log = logging.getLogger("blivet")
 
 INSTALLER_BLACKLIST = (r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram', r'ndblk', r'pmem')
 """ device name regexes to ignore when flags.installer_mode is True """
 
 
+def device_to_dict(device):
+    # Transform Device to dictionary
+    # Originally it was possible to use directly Device where needed,
+    # but it lead to unfixable excessive deprecation warnings from udev.
+    # Sice blivet uses Device.properties only (with couple of exceptions)
+    # this is a functional workaround. (japokorn May 2017)
+
+    result = dict(device.properties)
+    result["SYS_NAME"] = device.sys_name
+    result["SYS_PATH"] = device.sys_path
+    return result
+
+
 def get_device(sysfs_path):
     try:
-        dev = pyudev.Devices.from_sys_path(global_udev, sysfs_path)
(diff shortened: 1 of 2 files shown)


japokorn:
jenkins, test this please

blivet/udev.py
```diff
@@ -41,19 +41,31 @@
...
-        dev = pyudev.Devices.from_sys_path(global_udev, sysfs_path)
+        device = pyudev.Devices.from_sys_path(global_udev, sysfs_path)
     except pyudev.DeviceNotFoundError as e:
         log.error(e)
-        dev = None
+        result = None
+    else:
+        result = dict(device.properties)
+        result["SYS_NAME"] = device.sys_name
+        result["SYS_PATH"] = device.sys_path
```
dwlehman: I would suggest that you make a little function containing the three lines above, so you can reuse the code here and in `get_devices` below.

blivet/udev.py
```diff
@@ -101,7 +113,7 @@ def resolve_devspec(devspec, sysname=False):
             if device_get_uuid(dev) == devspec[5:]:
                 ret = dev
                 break
-        elif device_get_name(dev) == devname or dev.sys_name == devname:
+        elif device_get_name(dev) == devname or dev["SYS_NAME"] == devname:
```
dwlehman: I wonder if a new function called `device_get_sys_name` would be an improvement.

tests/populator_test.py
```diff
@@ -120,7 +120,7 @@ def test_get_helper(self, *args):
         """Test get_device_helper for dm devices."""
         device_is_dm = args[0]
         device_is_dm_lvm = args[6]
-        data = Mock()
+        data = {'SYS_PATH': 'dummy'}
```
dwlehman: Have you run the unit tests locally to verify this? I would be surprised if this change is sufficient.

M4rtinK:
@japokorn Works fine - the Anaconda log is now **much** more readable, thanks a lot! :)

Log with the patch applied: https://paste.fedoraproject.org/paste/ZWS-kjzm29z7MjZr1cDKLV5M1UNdIGYhyRLivL9gydE=

tests/populator_test.py
japokorn: I have. Local tests are passing. The return value of the affected function is mocked in all cases. This is just to prevent KeyError since that function is actually being executed before the mock happens.

blivet/udev.py
japokorn: Done.

dwlehman:
There are still a couple of issues pointed out by `pylint` and `pep8` that you can see in the CI console log.

japokorn:
Fixed pylint and pep8 related issues.

PR #586, opened 2017-05-12
Always mount & unmount an XFS file system when writing new UUID

vpodzime:
This is needed when the file system is not clean (e.g. if it was
created as a snapshot of some other mounted file system) which is
quite a common case. And since 'xfs_admin' exits with code 0 even
when it fails to set the UUID [1], we cannot do a second/fallback
try. The first one has to have the best chance to succeed, hence
this patch.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1450423

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -1076,19 +1076,15 @@ class XFS(FS):
            Raises an FSError if the UUID can not be set.
         """
 
+        # try to mount and umount the FS first to make sure it is clean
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #587, opened 2017-05-12
python2 compatibility, part 1

dwlehman:
This gets us to the point where we can run the unit tests under both python2 and python3 (on Fedora 25).

TODO:

- [x] work out a compatible way to run pylint under python2 or python3
- [ ] Makefile changes to run tests and pylint for both python versions
- [ ] rpm spec file changes to build for both python versions
- [ ] handle missing mock module in python2 where it isn't standard

### DIFF
--- a/blivet/populator/helpers/__init__.py
+++ b/blivet/populator/helpers/__init__.py
@@ -1,4 +1,5 @@
 import inspect as _inspect
+import six as _six
 
 from .devicepopulator import DevicePopulator
 from .formatpopulator import FormatPopulator
@@ -49,7 +50,7 @@ def get_device_helper(data):
         The helper lists are sorted according to priorities defined within each
         class. This function returns the first matching class.
     """
-    return next((h for h in _device_helpers if h.match(data)), None)
+    return _six.next((h for h in _device_helpers if h.match(data)), None)
 
 
 def get_format_helper(data, device):
@@ -58,4 +59,4 @@ def get_format_helper(data, device):
         The helper lists are sorted according to priorities defined within each
         class. This function returns the first matching class.
     """
-    return next((h for h in _format_helpers if h.match(data, device=device)), None)
+    return _six.next((h for h in _format_helpers if h.match(data, device=device)), None)
--- a/blivet/threads.py
+++ b/blivet/threads.py
@@ -24,6 +24,7 @@ from threading import RLock, current_thread, main_thread
 from types import FunctionType
 from abc import ABCMeta
 from six import raise_from, wraps
+import functools
 
 from .errors import ThreadError
 from .flags import flags
@@ -33,7 +34,7 @@ blivet_lock = RLock(verbose=flags.debug_threads)
 
 def exclusive(m):
     """ Run a callable while holding the global lock. """
-    @wraps(m)
+    @wraps(m, set(functools.WRAPPER_ASSIGNMENTS) & set(dir(m)))
     def run_with_lock(*args, **kwargs):
         with blivet_lock:
             if current_thread() == main_thread():
--- a/blivet/threads.py
+++ b/blivet/threads.py
@@ -24,6 +24,7 @@ from threading import RLock, current_thread, main_thread
 from functools import wraps
 from types import FunctionType
 from abc import ABCMeta
+from six import raise_from
 
 from .errors import ThreadError
 from .flags import flags
@@ -40,7 +41,7 @@ def exclusive(m):
                 exn_info = get_thread_exception()
                 if exn_info[1]:
                     clear_thread_exception()
-                    raise ThreadError("raising queued exception") from exn_info[1]
+                    raise_from(ThreadError("raising queued exception"), exn_info[1])
 
             return m(*args, **kwargs)
 
--- a/blivet/threads.py
+++ b/blivet/threads.py
@@ -20,16 +20,23 @@
 # Red Hat Author(s): David Lehman <dlehman@redhat.com>
 #
 
-from threading import RLock, current_thread, main_thread
+import threading
 from types import FunctionType
 from abc import ABCMeta
-from six import raise_from, wraps
+from six import raise_from, wraps, PY3
 import functools
 
 from .errors import ThreadError
 from .flags import flags
 
-blivet_lock = RLock(verbose=flags.debug_threads)
+blivet_lock = threading.RLock(verbose=flags.debug_threads)
+
+
+def _is_main_thread():
+    if PY3:
+        return threading.current_thread() == threading.main_thread()
+    else:
+        return threading.currentThread().name == "MainThread"
 
 
 def exclusive(m):
@@ -37,7 +44,7 @@ def exclusive(m):
     @wraps(m, set(functools.WRAPPER_ASSIGNMENTS) & set(dir(m)))
     def run_with_lock(*args, **kwargs):
         with blivet_lock:
-            if current_thread() == main_thread():
+            if _is_main_thread():
                 exn_info = get_thread_exception()
                 if exn_info[1]:
                     clear_thread_exception()
--- a/blivet/threads.py
+++ b/blivet/threads.py
@@ -21,10 +21,9 @@
 #
 
 from threading import RLock, current_thread, main_thread
-from functools import wraps
 from types import FunctionType
 from abc import ABCMeta
-from six import raise_from
+from six import raise_from, wraps
 
 from .errors import ThreadError
 from .flags import flags
--- a/tests/events_test.py
+++ b/tests/events_test.py
@@ -1,6 +1,6 @@
-import test_compat
+import test_compat  # pylint: disable=unused-import
 
-from six.moves.mock import Mock, patch
+from six.moves.mock import Mock, patch  # pylint: disable=no-name-in-module,import-error
 import time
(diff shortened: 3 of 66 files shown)


blivet/threads.py
```diff
@@ -40,7 +41,7 @@ def run_with_lock(*args, **kwargs):
                 exn_info = get_thread_exception()
                 if exn_info[1]:
```
tasleson: py2 it's OK to do `some_exception[N]`, but in py3 you need to do `some_exception.args[N]`, which also works for py2 as well.

blivet/threads.py
```diff
@@ -40,7 +41,7 @@ def run_with_lock(*args, **kwargs):
...
                     clear_thread_exception()
-                    raise ThreadError("raising queued exception") from exn_info[1]
```
tasleson: py2 it's OK to do `some_exception[N]`, but in py3 you need to do `some_exception.args[N]`, which also works for py2 as well.

tests/events_test.py
```diff
@@ -29,6 +30,7 @@ def testEventMask(self):
         device = "sdc"
         action = "add"
         mgr.handle_event(action, device)
+        time.sleep(1)
```
tasleson: Why are these needed for py2 and not for py3?

tasleson:
OK, I did review this, but only focused on the lines that were modified.  Good job keeping the changes for each commit simple.  Additionally I see that the `setup.py` doesn't have a requires section, so I'm guessing it's not utilized heavily for dependency validation, eg. missing `six` there now.

tests/events_test.py
dwlehman: These are not py2-specific. The failure was sporadic, only occurring maybe three times out of 25 or so runs, but it made me aware of the potential for a race.

blivet/threads.py
dwlehman:
```python
def get_thread_exception():
     return (_exception_thread, _thread_exception)
```

tasleson: Ah, never mind :-)

dwlehman:
Latest push fixed pep8 and pylint errors.

tests/test_compat.py
```diff
@@ -0,0 +1,46 @@
+# compat.py
```
vpodzime: ``test_compat.py``

tests/test_compat.py
```diff
@@ -0,0 +1,46 @@
...
+def setup():
+    global has_mock
+    add_move(mock_move)
+
+    try:
+        import six.moves.mock as _mock
+    except ImportError:
+        has_mock = False
+    else:
+        has_mock = True
```
vpodzime: What is ``has_mock`` good for? I take it an import from ``six.moves`` fails with ``ImportError`` if ``mock`` is not available. Isn't that enough?

blivet/threads.py
```diff
@@ -33,7 +34,7 @@
 
 def exclusive(m):
     """ Run a callable while holding the global lock. """
-    @wraps(m)
+    @wraps(m, set(functools.WRAPPER_ASSIGNMENTS) & set(dir(m)))
```
vpodzime: So ``six.wraps()`` does something else than Python 3's ``functools.wraps()``? What's the point of ``six.wraps()`` then?

blivet/populator/helpers/__init__.py
```diff
@@ -1,4 +1,5 @@
 import inspect as _inspect
+import six as _six
```
dwlehman: The goal was to keep the `blivet.populator.helpers` package namespace uncluttered.

tests/test_compat.py
dwlehman: Oops.

tests/test_compat.py
dwlehman: Well, we don't want the `ImportError` to cause the test run to end -- what we want is to adapt by only running the tests that do not require `mock`. Does that seem reasonable?

blivet/threads.py
dwlehman: From https://pythonhosted.org/six/index.html#six.wraps:

> @six.wraps(wrapped, assigned=functools.WRAPPER_ASSIGNMENTS, updated=functools.WRAPPER_UPDATES)
> This is exactly the functools.wraps() decorator, but it sets the __wrapped__ attribute on what it decorates as functools.wraps() does on Python versions after 3.2.

blivet/populator/helpers/__init__.py
vpodzime: Still don't see the point. It only makes difference for ``*`` imports, doesn't it? Maybe we should define ``__all__`` instead?

vpodzime: Still don't see the point. It only makes difference for ``*`` imports, doesn't it? Maybe we should define ``__all__`` instead?

tests/test_compat.py
vpodzime: Yeah, I just find 
```
try:
    import SOMETHING
except ImportError:
    has_something = False
else:
    has_something = True
```
more natural than ``from SOMETHING import has_something`` and then checking the boolean. But I definitely see the point in not doing this over and over again in all the places.

blivet/threads.py
vpodzime: https://docs.python.org/3/library/functools.html#functools.update_wrapper says:

> The optional arguments are tuples to specify which attributes of the original function are assigned directly to the matching attributes on the wrapper function and which attributes of the wrapper function are updated with the corresponding attributes from the original function.

So we want *everything* found in ``m`` to also be set on the wrapped function? This definitely looks like a change beyond just making this python-23 compatible.

dwlehman: The python2 version of wraps apparently doesn't automatically adjust the list of assignments for a method as opposed to a function. My understanding is that this change is a no-op on python3.

blivet/populator/helpers/__init__.py
dwlehman: Yeah, it's probably only relevant for `import *`. At some point I was noticing some junk in the namespaces in blivet from this type of import but, to be honest, I don't remember the exact details.

tests/test_compat.py
dwlehman: The review UI shows so little patch context I didn't realize which code we were discussing here. This is actually not supposed to be part of this pull request (it is not used by anything in it). This was a quick first pass at how to handle to possibility of a missing mock module in a way that requires minimal boilerplate code in the many test modules that use mock. I'll remove it from this PR.

blivet/threads.py
vpodzime: Interesting. I cannot see anything like that mentioned in the documentation. How / from where did you get the idea that something like this is necessary?

This surely is no big deal, I'm okay with this being merged as it is. :)

dwlehman: Before I added the additional arguments I saw errors like this when running under python2:

```
[dlehman@tsuga blivet]$ PYTHONPATH=.:tests:../pykickstart/ python2 -m unittest -v -f devices_test.dependencies_test
  File "/usr/lib64/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib64/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/usr/lib64/python2.7/unittest/__main__.py", line 12, in <module>
    main(module=None)
  File "/usr/lib64/python2.7/unittest/main.py", line 94, in __init__
    self.parseArgs(argv)
  File "/usr/lib64/python2.7/unittest/main.py", line 149, in parseArgs
    self.createTests()
  File "/usr/lib64/python2.7/unittest/main.py", line 158, in createTests
    self.module)
  File "/usr/lib64/python2.7/unittest/loader.py", line 130, in loadTestsFromNames
    suites = [self.loadTestsFromName(name, module) for name in names]
  File "/usr/lib64/python2.7/unittest/loader.py", line 91, in loadTestsFromName
    module = __import__('.'.join(parts_copy))
  File "/home/dlehman/src/blivet/tests/devices_test/dependencies_test.py", line 5, in <module>
    from blivet.deviceaction import ActionCreateDevice
  File "blivet/deviceaction.py", line 28, in <module>
    from .devices import StorageDevice
  File "blivet/devices/__init__.py", line 23, in <module>
    from .device import Device
  File "blivet/devices/device.py", line 37, in <module>
    class Device(util.ObjectID):
  File "/usr/lib/python2.7/site-packages/six.py", line 824, in wrapper
    return metaclass(cls.__name__, cls.__bases__, orig_vars)
  File "blivet/threads.py", line 77, in __new__
    obj = property(fget=exclusive(obj.__get__),
  File "blivet/threads.py", line 45, in exclusive
    @wraps(m)
  File "/usr/lib/python2.7/site-packages/six.py", line 792, in wrapper
    f = functools.wraps(wrapped, assigned, updated)(f)
  File "/usr/lib64/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'method-wrapper' object has no attribute '__module__'
[dlehman@tsuga blivet]$ 
```

dwlehman: (Identical traceback when using `functools.wraps` there, btw.)

vpodzime: I see.

PR #593, opened 2017-05-16
Various pylint fixes

vojtechtrefny:
We now have new pylint in Fedora and it has a lot of new awesome warnings and errors.

Test will still fail on this because pylint started writing the score to stdout and pocketlint thinks it is an error. https://github.com/rhinstaller/pocketlint/pull/16 should fix this


vojtechtrefny:
Jenkins, test this please.

PR #594, opened 2017-05-18
pylint: ignore some pylint warnings in tests

vojtechtrefny:
This code is only on `2.2-devel` so these errors/warnings were not fixed by #593


vojtechtrefny:
Jenkins, test this please.

PR #597, opened 2017-05-24
external dependency tests

dwlehman:
Am I being lazy by just leaving the existing tests there? I was tempted to rewrite them using mock and combine with the new ones I'm adding here.

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -131,7 +131,7 @@ class StorageDevice(Device):
 
         self._readonly = False
         self._protected = False
-        self.controllable = not flags.testing
+        self._controllable = True
 
         # Copy only the validity check from Device._set_name() so we don't try
         # to check a bunch of inappropriate state properties during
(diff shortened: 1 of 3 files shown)


blivet/devices/storage.py
```diff
@@ -190,6 +190,14 @@ def raw_device(self):
         """ The device itself, or when encrypted, the backing device. """
         return self
 
+    @property
+    def controllable(self):
+        return self._controllable and not flags.testing and not self.unavailable_type_dependencies()
+
+    @controllable.setter
+    def controllable(self, value):
+        self._controllable = value
```
vpodzime: Does the setter make sense here? This looks to me like it shouldn't be set "from outside".

dwlehman: There are a few places from the populator where we update the controllable property for some special cases.

vpodzime: Fair enough

PR #599, opened 2017-06-02
Round the recommended thpool metadata size to extents

vpodzime:
Otherwise LVM does it when creating the thin pool and we may end
up with not enough space in the VG to create the data part of the
pool (which happens afterwards).

Resolves: rhbz#1456528


vpodzime:
> Looks good. Is there anywhere else we need to do this rounding?

Not on the *rhel7-branch*, AFAIK.

PR #601, opened 2017-06-02
Fixes related to missing external deps

dwlehman:
This is stuff I encountered while testing the API demo on a freshly-installed F26 Alpha Server installation.

### DIFF
--- a/blivet/static_data/lvm_info.py
+++ b/blivet/static_data/lvm_info.py
@@ -37,7 +37,13 @@ class LVsInfo(object):
     @property
     def cache(self):
         if self._lvs_cache is None:
-            lvs = blockdev.lvm.lvs()
+            try:
+                lvs = blockdev.lvm.lvs()
+            except NotImplementedError:
+                log.error("libblockdev lvm plugin is missing")
+                self._lvs_cache = dict()  # pylint: disable=attribute-defined-outside-init
+                return self._lvs_cache
+
             self._lvs_cache = dict(("%s-%s" % (lv.vg_name, lv.lv_name), lv) for lv in lvs)  # pylint: disable=attribute-defined-outside-init
 
         return self._lvs_cache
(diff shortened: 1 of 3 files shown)


dwlehman:
Jenkins, test this please.

blivet/static_data/lvm_info.py
```diff
@@ -37,7 +37,13 @@ class LVsInfo(object):
     @property
     def cache(self):
         if self._lvs_cache is None:
-            lvs = blockdev.lvm.lvs()
+            try:
+                lvs = blockdev.lvm.lvs()
+            except blockdev.BlockDevNotImplementedError:  # pylint: disable=catching-non-exception
```
vpodzime: It should be possible to catch ``NotImplementedError`` here too (``BlockDevNotImplementedError`` should be inheriting from it)

blivet/static_data/lvm_info.py
```diff
@@ -56,8 +62,14 @@ class PVsInfo(object):
...
     def cache(self):
         if self._pvs_cache is None:
-            pvs = blockdev.lvm.pvs()
             self._pvs_cache = dict()  # pylint: disable=attribute-defined-outside-init
+
+            try:
+                pvs = blockdev.lvm.pvs()
+            except blockdev.BlockDevNotImplementedError:  # pylint: disable=catching-non-exception
+                log.error("libblockdev lvm plugin is missing")
+                return self._pvs_cache
```
vpodzime: Shouldn't this also initialize the cache to an empty dict if the plugin is missing like for ``lvs`` above? There's no point trying this multiple times (for every PV), right?

dwlehman: Yes, good catch.

dwlehman: Actually, it's not needed because the dict is initialized just above the changed code (within the same property).

vpodzime: I see it now. I got confused by two different ways the same is achieved for PVs and LVs.

PR #602, opened 2017-06-06
Make sure an LV is deactivated before removal

vpodzime:
Before and LV is removed, we may need to set its ancestors (VG
and its PVs) up which may trigger auto-activation of the LV
itself. And since we are supposed to always remove a deactivated
LV so we have to make sure the LV is deactivated before we try to
remove it.

Resolves: rhbz#1456821

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -792,6 +792,10 @@ class LVMLogicalVolumeDevice(DMDevice):
         # set up the vg's pvs so lvm can remove the lv
         self.vg.setupParents(orig=True)
 
+        # setting up VG's PVs may have caused this LV was automatically
+        # activated, make sure it is deactivated before removal
+        self.teardown()
+
     def _destroy(self):
         """ Destroy the device. """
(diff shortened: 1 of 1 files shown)


blivet/devices/lvm.py
```diff
@@ -792,6 +792,10 @@ def _preDestroy(self):
         # set up the vg's pvs so lvm can remove the lv
         self.vg.setupParents(orig=True)
 
+        # setting up VG's PVs may have caused this LV was automatically
+        # activated, make sure it is deactivated before removal
+        self._teardown()
```
vpodzime: Because the code tears the device down before calling ``destroy()`` and we really only want to call ``lvdeactivate()`` after the other invocations in ``_pre_destroy()``.

dwlehman: I worry about 1) a race where the lv isn't activated until between the `lvdeactivate` and the `lvremove` and 2) an auto-activated format such as swap (which would be taken care of by calling the full `teardown`, assuming no races).

PR #604, opened 2017-06-07
Adjust Makefile and spec file for Python 2 compatibility

vojtechtrefny:
Allow building blivet with Python 2 and add a Python 2 subpackage.

-----
I've tested this in Copr (build [here](https://copr.fedorainfracloud.org/coprs/vtrefny/test2/build/562282/)) and on my CentOS machine. Few notes:

- CentOS has `pyudev` 0.15 and we require >= 0.18, I've changed this only for the copr build.
- Translation canary doesn't work with Python 2
- `python2-pocketlint` is not ready yet, Ie created a [copr repo with it](https://copr.fedorainfracloud.org/coprs/vtrefny/pocketlint/)
- Version for _Conflicts_ and _Obsoletes_ for the `blivet-data` package needs to be adjusted on the release branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,21 @@
+PYTHON?=python3
+PKG_INSTALL?=dnf
+
 PKGNAME=blivet
 SPECFILE=python-blivet.spec
-VERSION=$(shell python3 setup.py --version)
+VERSION=$(shell $(PYTHON) setup.py --version)
 RPMVERSION=$(shell rpmspec -q --queryformat "%{version}\n" $(SPECFILE) | head -1)
 RPMRELEASE=$(shell rpmspec --undefine '%dist' -q --queryformat "%{release}\n" $(SPECFILE) | head -1)
 RC_RELEASE ?= $(shell date -u +0.1.%Y%m%d%H%M%S)
 RELEASE_TAG=$(PKGNAME)-$(RPMVERSION)-$(RPMRELEASE)
 VERSION_TAG=$(PKGNAME)-$(VERSION)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -61,7 +78,7 @@ check-requires:
 
 install-requires:
 	@echo "*** Installing the dependencies required for testing and analysis ***"
-	dnf install -y $(TEST_DEPENDENCIES)
+	yum install -y $(TEST_DEPENDENCIES)
```
dwlehman: The choice of package manager should probably be made up near the top where we decide on other variables, right?

python-blivet.spec
```diff
@@ -25,16 +34,34 @@ Source0: http://github.com/rhinstaller/blivet/archive/%{realname}-%{realversion}
...
-BuildRequires: python3-devel python3-setuptools
 
 %description
 The python-blivet package is a python module for examining and modifying
 storage configuration.
 
+%package -n %{realname}-data
+Summary: Data for the %{realname} python module.
+
+BuildRequires: systemd
```
dwlehman: What's this for?

Makefile
vojtechtrefny: Fixed

python-blivet.spec
vojtechtrefny: This is needed for the `%{_unitdir}` macro -- https://fedoraproject.org/wiki/Packaging:Systemd#.25files_section

vojtechtrefny:
Jenkins, test this please.

PR #606, opened 2017-06-13
New test for selinux context setting when mounting

japokorn:
Added new test for checking correct function parameters
when setting selinux context while mounting

### DIFF
--- a/tests/formats_test/selinux_test.py
+++ b/tests/formats_test/selinux_test.py
@@ -1,96 +1,53 @@
+# pylint: disable=unused-import
 import os
-import selinux
-import tempfile
+from unittest.mock import patch, ANY
 import unittest
+import selinux
 
 import blivet
-from tests import loopbackedtestcase
 import blivet.formats.fs as fs
-from blivet.size import Size
 
 
-@unittest.skipUnless(selinux.is_selinux_enabled() == 1, "SELinux is disabled")
-class SELinuxContextTestCase(loopbackedtestcase.LoopBackedTestCase):
-
+class SELinuxContextTestCase(unittest.TestCase):
     """Testing SELinux contexts.
     """
-
-    def __init__(self, methodName='run_test'):
-        super(SELinuxContextTestCase, self).__init__(methodName=methodName, device_spec=[Size("100 MiB")])
 
     def setUp(self):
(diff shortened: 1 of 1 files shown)


japokorn:
Implemented changes requested by dwlehman.

tests/formats_test/selinux_test.py
```diff
@@ -92,6 +94,31 @@ def test_mounting_xfs(self):
         an_fs.unmount()
         os.rmdir(mountpoint)
 
+    @patch("blivet.util.mount", return_value=0)
+    @patch.object(fs.FS, "_pre_setup", return_value=True)
+    @patch("os.access", return_value=True)
+    # pylint: disable=unused-argument
+    # pylint: disable=no-self-use
+    def test_mount_selinux_context(self, *args):
```
japokorn: Pylint disables needed:
-  `*args` cannot be ommitted, because it is used by patch decorators
- no `self` call triggers "This should be a function" warning, but the method has to be part of the class to be run.

tests/formats_test/selinux_test.py
```diff
@@ -1,96 +1,54 @@
...
 import unittest
+import selinux
 
 import blivet
-from tests import loopbackedtestcase
 import blivet.formats.fs as fs
 from blivet.size import Size
 
 
 @unittest.skipUnless(selinux.is_selinux_enabled() == 1, "SELinux is disabled")
```
dwlehman: I think you can remove this line now since we're not testing the actual selinux calls anymore.

japokorn:
All requested changes are now implemented.

tests/formats_test/selinux_test.py
```diff
@@ -1,96 +1,54 @@
...
 import unittest
+import selinux
 
 import blivet
-from tests import loopbackedtestcase
 import blivet.formats.fs as fs
-from blivet.size import Size
 
 
 @unittest.skipUnless(selinux.is_selinux_enabled() == 1, "SELinux is disabled")
```
dwlehman: Please remove the above line since these tests do not depend on selinux being enabled.

japokorn: removed

PR #611, opened 2017-06-21
Stop enforcing obsolete limits on partition count. (#1460668)

dwlehman:
Today I learned that the partition count limits in pyparted are in fact massively outdated (rhinstaller/pyparted#41). These days major/minor numbers are dynamically allocated so there is no limit except for space. There is a notable exception in mmcblk, which has a kernel config value for the number of minors per major. It defaults to 8. Good grief.


dwlehman:
Changes subsequent to review were limited to removing small amounts of unused code from the partitioning tests.

PR #617, opened 2017-07-20
No longer skip test w/o selinux; mocked

japokorn:
Rather than skiping the test, parts of it that need selinux were mocked

### DIFF
--- a/tests/formats_test/selinux_test.py
+++ b/tests/formats_test/selinux_test.py
@@ -13,8 +13,6 @@ class SELinuxContextTestCase(unittest.TestCase):
     """
 
     def setUp(self):
-        if not blivet.flags.selinux:
-            self.skipTest("SELinux disabled.")
         self.installer_mode = blivet.flags.installer_mode
         super(SELinuxContextTestCase, self).setUp()
 


tests/formats_test/selinux_test.py
```diff
@@ -13,14 +13,13 @@ class SELinuxContextTestCase(unittest.TestCase):
...
     def setUp(self):
-        if not blivet.flags.selinux:
-            self.skipTest("SELinux disabled.")
         self.installer_mode = blivet.flags.installer_mode
         super(SELinuxContextTestCase, self).setUp()
 
     @patch("blivet.util.mount", return_value=0)
     @patch.object(fs.FS, "_pre_setup", return_value=True)
     @patch("os.access", return_value=True)
+    @patch("selinux.lsetfilecon")
```
vpodzime: Don't we need to make sure the function is (not) being called or something?

dwlehman: You're already patching `selinux.lsetfilecon` below (from the original commit, I think). Adding this patch call for the whole method shouldn't be necessary unless there is something going on that needs to be better understood.

japokorn: Removed `selinux.lsetfilecon` patch.

japokorn:
pushed

PR #619, opened 2017-08-03
Fixed behavior when selinux is disabled

japokorn:
selinux flag is set to True for the duration of the test

### DIFF
--- a/tests/formats_test/selinux_test.py
+++ b/tests/formats_test/selinux_test.py
@@ -14,10 +14,9 @@ class SELinuxContextTestCase(unittest.TestCase):
 
     def setUp(self):
         self.installer_mode = blivet.flags.installer_mode
+        self.selinux = blivet.flags.selinux
+        super(SELinuxContextTestCase, self).setUp()
         self.addCleanup(self._clean_up)
-
-    def _clean_up(self):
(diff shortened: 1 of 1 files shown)


dwlehman:
This test should be performing the (fake) mount both with `flags.selinux` enabled and disabled to verify that the flag correctly controls whether we try to adjust the context on the mountpoint and `lost+found` directory.

japokorn:
I have moved tearDown to _clean_up. PR should be ready for closing now.

dwlehman:
Please rebase so there is no merge commit on the branch. Other than that, it looks good.

tests/formats_test/selinux_test.py
```diff
@@ -51,3 +52,8 @@ def test_mount_selinux_ext2fs(self):
     def test_mount_selinux_xfs(self):
         """ Test of correct selinux context parameter value when mounting XFS"""
         self.exec_mount_selinux_format(fs.XFS)
+
+    def _clean_up(self):
+        super(SELinuxContextTestCase, self).tearDown()
```
vojtechtrefny: Please remove this line. We don't want to call `tearDown` from `_clean_up` method.

japokorn:
Commits squashed

japokorn:
pushed

PR #623, opened 2017-08-16
Move anaconda-specific code to anaconda

dwlehman:


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

blivet/formats/disklabel.py
```diff
@@ -218,12 +218,71 @@ def parted_device(self):
...
+    @classmethod
+    def set_default_label_type(cls, labeltype):
+        cls._default_label_type = labeltype
+        log.debug("default disklabel has been set to %s", labeltype)
+
+    def _label_type_size_check(self, label_type):
+        if self.parted_device is None:
+            return False
+
+        _label = parted.freshDisk(device=self.parted_device, ty=label_type)
```
vpodzime: This should be just ``label``

blivet/formats/disklabel.py
```diff
@@ -218,12 +218,71 @@ def parted_device(self):
...
+        return self.parted_device.length < _label.maxPartitionStartSector
+
+    def _get_best_label_type(self):
+        label_type = self.__class__._default_label_type
+        label_types = self.__class__.get_platform_label_types()[:]
```
vpodzime: The ``__class__`` part here can be dropped, AFAICT.

blivet/formats/disklabel.py
```diff
@@ -218,12 +218,71 @@ def parted_device(self):
...
+            elif self.parted_device.type == parted.DEVICE_DASD:
+                # the device is DASD
+                return "dasd"
+
+        for lt in label_types:
+            if self._label_type_size_check(lt):
+                log.debug("selecting %s disklabel for %s based on size",
+                          label_type, os.path.basename(self.device))
+                label_type = lt
+                break
```
vpodzime: Why not just ``return lt`` here?

blivet/formats/disklabel.py
```diff
@@ -218,12 +218,71 @@ def parted_device(self):
...
+
+        return label_type
```
vpodzime: Do we really want to return the default label type even if it doesn't match the *size_check()* criteria?

blivet/formats/disklabel.py
dwlehman: Fixed.

blivet/formats/disklabel.py
dwlehman: We always want to return a label type. Even if none of the types pass the size test, we still need to return one.

blivet/formats/disklabel.py
dwlehman: I just prefer to have one return statement per method, aside from very-early returns due to precondition checking.

dwlehman: I see I returned early in the previous s390-specific block just above this one. It was just too much change to try to cram that extremely platform-specific logic into the single-return pattern.

vpodzime:
Still looks good to me.

PR #624, opened 2017-08-16
Mark the DBus interface as unstable.

dwlehman:
It's important to do this before 2.2 is released.

### DIFF
--- a/blivet/dbus/constants.py
+++ b/blivet/dbus/constants.py
@@ -18,8 +18,8 @@
 # Red Hat Author(s): David Lehman <dlehman@redhat.com>
 #
 
-BUS_NAME = "com.redhat.Blivet1"
-BASE_OBJECT_PATH = "/com/redhat/Blivet1"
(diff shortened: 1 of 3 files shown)


blivet/dbus/constants.py
```diff
@@ -19,7 +19,7 @@
 #
 
 BUS_NAME = "com.redhat.Blivet1"
-BASE_OBJECT_PATH = "/com/redhat/Blivet1"
+BASE_OBJECT_PATH = "/com/redhat/Blivet0"
```
vpodzime: I think we need a documentation for the DBus API and mention that it is unstable there. Also there's the ``release_notes.rst`` file which should mention this.

dwlehman:
I added some documentation of the DBus interface. Please take a look and see what you think:

  https://dlehman.fedorapeople.org/blivet-2.2-doc/  (Table of Contents)
  https://dlehman.fedorapeople.org/blivet-2.2-doc/dbus.html  (DBus Interface)

PR #625, opened 2017-08-18
Added support for device tags

japokorn:
added tags support for ks parameter 'ignoredisk'
added tests for 'ignoredisk' tags

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -34,6 +34,7 @@ from .errors import DeviceError, DeviceTreeError, StorageError
 from .deviceaction import ActionDestroyDevice, ActionDestroyFormat
 from .devices import BTRFSDevice, NoDevice, PartitionDevice
 from .devices import LVMLogicalVolumeDevice, LVMVolumeGroupDevice
+from .devices.lib import Tags
 from . import formats
 from .devicelibs import lvm
 from .events.changes import record_change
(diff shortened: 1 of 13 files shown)


japokorn:
Partial PR. Will be changed.
Contains: 
- support for ks command 'ignoredisk' tags
- tests for it

japokorn:
Closed due to commit merging problems.

blivet/devicetree.py
```diff
@@ -883,9 +884,25 @@ def unhide(self, device):
...
+            tags = [t[1:] for t in taglist if t.startswith("@")]
+            for tag in tags:
+                if tag not in Tags.__members__:
+                    raise ValueError("unknown ignoredisk tag '@%s' encountered" % tag)
+                if Tags(tag) in disk.tags:
+                    return True
+            return False
+
+        return ((self.ignored_disks and disk_in_taglist(disk, self.ignored_disks)) or
+                (self.exclusive_disks and not disk_in_taglist(disk, self.exclusive_disks)))
```
dwlehman: Separating the logic for matching devices by name from matching them by tags would make it easier to see what's going on in this code. You could also probably streamline it into something like this (starting after line 896):

```python
if not all(tag in Tags.__members__ for tag in tags):
    raise ValueError("invalid tag specified")
return any(Tags(tag) in disk.tags for tag in tags)
```

PR #627, opened 2017-08-29
Do file system check before resize (#1484575)

vpodzime:
Otherwise the resize usually fails.

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -427,10 +427,8 @@ class DeviceFormat(ObjectID, metaclass=SynchronizedMeta):
         if not self.device == "tmpfs" and not os.path.exists(self.device):
             raise FormatResizeError("device does not exist", self.device)
 
-        # The first minimum size can be incorrect if the fs was not
-        # properly unmounted. After do_check the minimum size will be correct
-        # so run the check one last time and bump up the size if it was too
-        # small.
+        self._pre_resize()
+
         self.update_size_info()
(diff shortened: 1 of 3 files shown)


AdamWill:
I am unsure if the last change - going back from `-n` to `-p` - belongs in this commit, per the discussion when it was changed from `-p` to `-n` in a43dc757eff7fe9417f72e25fc373d09439963de . It *may* be appropriate to change this back now, but shouldn't it be a separate PR where we can discuss and check for sure that it won't lead to unexpected access to filesystems the user didn't want us to touch?

vpodzime:
> shouldn't it be a separate PR where we can discuss and check for sure that it won't lead to unexpected access to filesystems the user didn't want us to touch?

I think it's okay to discuss it here. *ext* resize doesn't really work without the change so it's either all or nothing. Regarding the access to file systems, ``do_check()`` is only called in the ``_pre_resize()`` and ``_post_resize()`` hooks now so only when users explicitly say they want a file system to be resized. Which should be fine, I think. In the long term it might make sense to split things into two methods ``do_check()`` and ``do_repair()`` similarly how [*libblockdev* has these things separated](http://storaged.org/libblockdev/libblockdev-FS.html#bd-fs-repair).

AdamWill:
so, is this gonna be merged?

vpodzime:
> so, is this gonna be merged?

I think we should wait for @dwlehman's opinion on this.

PR #632, opened 2017-09-14
edd_test: don't run on non-x86

vojtechtrefny:
Since this doesn't make any sense on non-x86 machines anyway, and
apparently I've got some endian bugs in it (which don't really
matter...), just don't run the edd tests on non-x86.

------------
#276 rebased to 2.1-devel


vojtechtrefny:
Jenkins, test this please.

PR #638, opened 2017-10-05
DM device subsystem is now obtained by libblockdev

japokorn:
Changed way of obtaining DM device subsystem value
Now libblockdev function is called instead of UUID parsing

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -31,6 +31,11 @@ from . import util
 from .size import Size
 from .flags import flags
 
+import gi
+gi.require_version("BlockDev", "2.0")
+
+from gi.repository import BlockDev as blockdev
+
 global_udev = pyudev.Context()
 log = logging.getLogger("blivet")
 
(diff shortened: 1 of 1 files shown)


blivet/udev.py
```diff
@@ -592,15 +597,12 @@ def device_get_lv_type(info):
...
-    uuid = info.get("DM_UUID", "")
-    uuid_fields = uuid.split("-")
-    _subsystem = uuid_fields[0]
-    if _subsystem.lower().startswith("part") and len(uuid_fields) > 1:
-        # kpartx uses partN- as a subsystem prefix, which we ignore because
-        # we only care about the subsystem of the partitions' parent device.
-        _subsystem = uuid_fields[1]
-
-    if _subsystem == uuid or not _subsystem:
+    name = info.get("DM_NAME", None)
```
dwlehman: `None` is the default return value for `dict.get` when called with a missing key, so you can just use `info.get("DM_NAME")` there.

japokorn: Fixed.

blivet/udev.py
```diff
@@ -592,15 +597,12 @@ def device_get_lv_type(info):
...
-        # kpartx uses partN- as a subsystem prefix, which we ignore because
-        # we only care about the subsystem of the partitions' parent device.
-        _subsystem = uuid_fields[1]
-
-    if _subsystem == uuid or not _subsystem:
+    name = info.get("DM_NAME")
+    if name is None:
+        return False
+
+    _subsystem = blockdev.dm_get_subsystem_from_name(name)
```
vpodzime: This should be ``blockdev.dm.get_subtsystem_from_name()``. ``blockdev.dm`` is a so-called ``ErrorProxy`` making sure nice python exceptions are returned instead of the raw C ones.

japokorn: Fixed.

japokorn:
jenkins, test this please

japokorn:
Pushed.

PR #641, opened 2017-10-16
mkfs.ext4 no longer hangs on confirmation

japokorn:
Fixed confirmation hang when ext[234] partition was created on the whole
disk.

Resolves: rhbz#1461459

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -967,6 +967,9 @@ class Ext2FS(FS):
     _minSize = Size(0)
     _defaultMountOptions = ["defaults"]
     _defaultCheckOptions = ["-f", "-p", "-C", "0"]
+    # force option required to avoid confirmation question when mkfs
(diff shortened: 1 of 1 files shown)


blivet/formats/fs.py
```diff
@@ -1103,7 +1103,7 @@ def needsFSCheck(self):
 class Ext4FS(Ext3FS):
     """ ext4 filesystem. """
     _type = "ext4"
-    _defaultFormatOptions = ["-t", "ext4"]
+    _defaultFormatOptions = ["-F", "-t", "ext4"]
```
vpodzime: What about ext2 and ext3? Also, I think this option is worth adding a comment to the code.

japokorn: I have reproduced the issue on ext2 and ext3 as well. Added fixes for these too. I have also added comment to explain why force option is needed.

japokorn:
Pushed.

PR #644, opened 2017-11-24
Fix resolve_devspec to fully support raid devices

poncovka:
Return md name if it matches a device name for the given device
specification, so the device names are consistent when we work
with udev module and a devicetree.

(cherry-picked from a commit 76ec17a)

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -115,6 +115,7 @@ def resolve_devspec(devspec, sysname=False):
     from . import devices
 
     devname = devices.device_path_to_name(devspec)
+    get_name = device_get_name
 
     ret = None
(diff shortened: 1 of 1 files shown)


poncovka:
I am not sure, because `device_get_name` will not return md name if it finds dm name, so the check in `resolve_devspec` can fail. For example, sys name can be also returned by `device_get_name`, but there is a check for it in `resolve_devspec`. But I don't know if it is a relevant scenario.

The original bug for RHEL is here: https://bugzilla.redhat.com/show_bug.cgi?id=1327439

poncovka:
Ok, then I'm closing this pull request.

PR #646, opened 2017-11-30
Use libblockdev runtime dependency checks

vojtechtrefny:
By default libblockdev will fail to load a plugin during its
initialization if one (or more) of the plugin runtime dependencies
is missing. For example if the "swaplabel" utility is missing,
loading of the swap module will fail.
This commit disables dependency checks during init -- in the same
situation swap plugin will load and only calling "swap_set_label"
would fail (because the utility it uses is missing).
Specification of all required "technologies" and "modes" is added
to the "tasks" module so checking for external tools availability
still checks if libblockdev supports everything we need.

### DIFF
--- a/blivet/tasks/availability.py
+++ b/blivet/tasks/availability.py
@@ -26,8 +26,10 @@ from six import add_metaclass
 
 import gi
 gi.require_version("BlockDev", "2.0")
+gi.require_version("GLib", "2.0")
 
 from gi.repository import BlockDev as blockdev
+from gi.repository import GLib
 
 import logging
 log = logging.getLogger("blivet")
@@ -172,9 +174,43 @@ class VersionMethod(Method):
         return self._availability_errors[:]
 
 
+class BlockDevTechInfo(object):
+
+    def __init__(self, plugin_name, check_fn, technologies):
+        """ Initializer.
+
+            :param str plugin_name: the name of the libblockdev plugin
+            :param check_fn: function used to check for support availability
+            :param technologies: list of required technologies
+        """
+        self.plugin_name = plugin_name
+        self.check_fn = check_fn
+        self.technologies = technologies
+
+    def __str__(self):
+        return "blockdev-%s" % self.plugin_name
+
+
 class BlockDevMethod(Method):
 
     """ Methods for when application is actually a libblockdev plugin. """
+
+    def __init__(self, tech_info):
+        """ Initializer.
+
+            :param :class:`AppVersionInfo` version_info:
+        """
+        self._tech_info = tech_info
+        self._availability_errors = None
+
+    def _check_technologies(self):
+        errors = []
+        for tech, mode in self._tech_info.technologies.items():
+            try:
+                self._tech_info.check_fn(tech, mode)
+            except GLib.GError as e:
+                errors.append(str(e))
+        return errors
 
     def availability_errors(self, resource):
         """ Returns [] if the plugin is loaded.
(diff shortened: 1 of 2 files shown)


blivet/tasks/availability.py
```diff
@@ -246,15 +285,108 @@ def available_resource(name):
...
+
+# libblockdev loop plugin required technologies and modes
+BLOCKDEV_LOOP_ALL_MODES = (blockdev.LoopTechMode.CREATE |
+                           blockdev.LoopTechMode.CREATE |
+                           blockdev.LoopTechMode.DESTROY |
+                           blockdev.LoopTechMode.MODIFY |
+                           blockdev.LoopTechMode.QUERY)
+BLOCKDEV_LOOP = BlockDevTechInfo(plugin_name="loop",
+                                 check_fn=blockdev.loop_is_tech_avail,
+                                 technologies={blockdev.LoopTech.LOOP_TECH_LOOP: BLOCKDEV_LOOP_ALL_MODES})  # FIXME
```
vpodzime: What is it to be fixed here?

blivet/tasks/availability.py
```diff
@@ -246,15 +285,108 @@ def available_resource(name):
...
+BLOCKDEV_LVM = BlockDevTechInfo(plugin_name="lvm",
+                                check_fn=blockdev.lvm_is_tech_avail,
+                                technologies={blockdev.LVMTech.BASIC: BLOCKDEV_LVM_ALL_MODES,
+                                              blockdev.LVMTech.BASIC_SNAP: BLOCKDEV_LVM_ALL_MODES,
+                                              blockdev.LVMTech.THIN: BLOCKDEV_LVM_ALL_MODES,
+                                              blockdev.LVMTech.CACHE: BLOCKDEV_LVM_ALL_MODES,
+                                              blockdev.LVMTech.CALCS: blockdev.LVMTechMode.QUERY,
+                                              blockdev.LVMTech.THIN_CALCS: blockdev.LVMTechMode.QUERY,
+                                              blockdev.LVMTech.CACHE_CALCS: blockdev.LVMTechMode.QUERY,
+                                              blockdev.LVMTech.GLOB_CONF: blockdev.LVMTechMode.QUERY})
```
vpodzime: ``GLOB_CONF`` should be checked for both ``QUERY`` and ``MODIFY``.
(I know the checks for this in *libblockdev* are not implemented strictly right now)

blivet/tasks/availability.py
```diff
@@ -246,15 +285,108 @@ def available_resource(name):
...
+                            blockdev.MpathTechMode.QUERY)
+BLOCKDEV_MPATH = BlockDevTechInfo(plugin_name="mpath",
+                                  check_fn=blockdev.mpath_is_tech_avail,
+                                  technologies={blockdev.MpathTech.BASE: BLOCKDEV_MPATH_ALL_MODES})
+BLOCKDEV_MPATH_TECH = BlockDevMethod(BLOCKDEV_MPATH)
+
+# libblockdev swap plugin required technologies and modes
+BLOCKDEV_SWAP_ALL_MODES = (blockdev.SwapTechMode.CREATE |
+                           blockdev.SwapTechMode.ACTIVATE_DEACTIVATE |
+                           blockdev.SwapTechMode.QUERY)
```
vpodzime: These are not all modes of the swap tech.

blivet/tasks/availability.py
```diff
@@ -246,15 +285,108 @@ def available_resource(name):
...
+BLOCKDEV_SWAP = BlockDevTechInfo(plugin_name="swap",
+                                 check_fn=blockdev.swap_is_tech_avail,
+                                 technologies={blockdev.SwapTech.SWAP_TECH_SWAP: BLOCKDEV_SWAP_ALL_MODES})
+BLOCKDEV_SWAP_TECH = BlockDevMethod(BLOCKDEV_SWAP)
+
+# libblockdev plugins
+# we can't just check if the plugin is loaded, we also need to make sure
+# that all technologies required by us our supported (some may be missing
+# due to missing depenecies)
```
vpodzime: *dependencies*

blivet/tasks/availability.py
vojtechtrefny: That was a reminder for me to try to fix the constant name (`blockdev.LoopTech.LOOP_TECH_LOOP`) in libblockdev.

blivet/tasks/availability.py
vojtechtrefny: Yes, but these are all modes we need in blivet. (We don't yet support changing label on swap.)

vojtechtrefny:
Jenkins, test this please.

PR #649, opened 2017-12-15
Fixes for factory thin pool calculations

vpodzime:
See the particular commit messages for details.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -1320,6 +1320,12 @@ class LVMFactory(DeviceFactory):
         if self.vg and self.vg.exists:
             return space
 
+        # unset the thpool_reserve here so that the previously reserved space is
+        # considered free space (and thus swallowed) -> we will set it and
+        # calculate an updated reserve below
+        if self.vg:
+            self.vg.thpool_reserve = None
+
         if self.container_size == SIZE_POLICY_AUTO:
             # automatic container size management
             if self.vg:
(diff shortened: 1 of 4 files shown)


dwlehman:
I finally got some quiet time to review these, and I already have some questions based only on the commit messages.

We base the thin pool metadata padding on the size of the vg? That seems weird. Why not base it on the sum of the sizes of the _thin_ lvs (or the pool size)? Is this because that's what the lvm team recommended way back when?

I know this is (also) backing up a bit, but why do we get/store/update the metadata size as recommended by lvm tooling? Is that so we can make a better guess at the total vg space usage for the pool?

vojtechtrefny:
> We base the thin pool metadata padding on the size of the vg? That seems weird. Why not base it on the sum of the sizes of the thin lvs (or the pool size)? Is this because that's what the lvm team recommended way back when?

The free space reserve in the vg is based on the vg size. Ad yes, this is what lvm team recommended -- 20 % of the vg should be reserved for the future thinpool grow.

> I know this is (also) backing up a bit, but why do we get/store/update the metadata size as recommended by lvm tooling? Is that so we can make a better guess at the total vg space usage for the pool?

Yes, and we also manually specify metadata size when creating the pool. I guess this is the only way how to control it and not "guess" LVM defaults.

vojtechtrefny:
I've tested the fix with blivet 3.1 and it fixes the issue -- without this on 500 GiB drive we will create 20 MiB metadata (during manual partitioning), with this fix it will be 200 MiB (that is what `thin_metadata_size` recommends for ~400 GiB thinpool with 65 KiB chunks).

PR #650, opened 2018-01-04
1.20 fedora updates

dwlehman:
python2-blivet1 has fallen into disrepair on Fedora since at least F25 due to various changes to dependencies. This should resolve everything. It was tested manually on F27.


vojtechtrefny:
Jenkins, test this please.

PR #651, opened 2018-01-22
Remove stale lvm metadata found on new md arrays.

dwlehman:
Multiple subsequent installations with the same layout (lvm-on-raid1) with lazy cleanup in between can lead to a situation in which stale lvm metadata is found by udev/lvmetad/pvscan immediately after creation of a new md array. The same udev/lvmetad/pvscan machinery activates this lvm, which trips blivet since it is unaware of the newly-activated devices.

There will also be a version of this for newer branches once the fix is confirmed by reporters.

### DIFF
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -20,6 +20,7 @@
 #
 
 import os
+import time
 
 from ..devicelibs import lvm
 from ..devicelibs import mdraid, raid
@@ -27,7 +28,6 @@ from ..devicelibs import mdraid, raid
 from .. import errors
 from .. import util
 from ..flags import flags
-from ..formats import getFormat
 from ..storage_log import log_method_call
 from .. import udev
 from ..i18n import P_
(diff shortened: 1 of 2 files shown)


blivet/devices/md.py
```diff
@@ -554,6 +556,42 @@ def preCommitFixup(self):
         if getattr(self.format, "mountpoint", None) == "/boot/efi":
             self.metadataVersion = "1.0"
 
+    def _removeStaleLVM(self):
+        """ Remove any stale LVM metadata that pre-existed in a new array's on-disk footprint.
+
+            XXX This should only be run from _postCreate.
```
vpodzime: Why is this not a nested function in ``_postCreate()`` then?

blivet/devices/md.py
```diff
@@ -554,6 +556,42 @@ def preCommitFixup(self):
...
+        """
+        import time
```
vpodzime: Why a local import?

blivet/devices/md.py
```diff
@@ -554,6 +556,42 @@ def preCommitFixup(self):
...
+
+        vg_name = None
+        vg_uuid = None
+        try:
+            vg_name = udev.device_get_vg_name(pv_info)
+            vg_uuid = udev.device_get_vg_uuid(pv_info)
+        except KeyError:
+            return
+
+        if vg_name and vg_uuid:
```
vpodzime: Are really both of these required? I'd think ``vg_uuid`` could be enough. I can definitely see some race conditions where the VG uuid is available, but it's name is not (yet).

blivet/devices/md.py
dwlehman: It probably should be. Good call.

blivet/devices/md.py
dwlehman: No good reason. That's just how I added it initially.

blivet/devices/md.py
dwlehman: I went back and forth on that. I considered not requiring `vg_name`. I even considered skipping the piece where I check the formatting as reported by udev because the only thing that really matters is if lvm found lv metadata. I think, though, that lvm will only find metadata if blkid/udev report the array is a pv. Anyway, you could argue to drop many of the checks and only rely on lvm's reporting and a `vg_uuid` to remove potential barriers to effectiveness. As I type this I'm leaning back in the direction of removing all of the non-essential checks.

blivet/devices/md.py
```diff
@@ -605,7 +569,33 @@ def _postCreate(self):
...
+            if vg_uuid:
+                log.info("removing stale LVM metadata found on %s", self.name)
+                try:
+                    lvm.vgremove(None, vg_uuid=vg_uuid)
+                except errors.LVMError as e:
+                    log.error("Failed to remove stale volume group from newly-created md array %s: %s",
+                              self.path, str(e))
+                    raise
+
+        removeStaleLVM()
```
vpodzime: This looks weird now that I see it. Maybe it does't have to be a separate function at all in the end? Anyway, looks good (or equally bad) to me either way.

dwlehman: Yeah, it's gross either way. I just wanted to clearly separate this particular hack from the rest of the code.

PR #652, opened 2018-01-24
Find and remove stale LVM metadata immediately after creating md array.

dwlehman:
This is a port of #651 to active development branches.

### DIFF
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -21,6 +21,7 @@
 
 import os
 import six
+import time
 
 import gi
 gi.require_version("BlockDev", "2.0")
@@ -31,6 +32,7 @@ from ..devicelibs import mdraid, raid
 
 from .. import errors
 from .. import util
+from ..static_data import pvs_info
 from ..storage_log import log_method_call
 from .. import udev
 from ..size import Size
@@ -533,6 +535,30 @@ class MDRaidArrayDevice(ContainerDevice, RaidDevice):
         self.uuid = info.uuid
         for member in self.members:
             member.format.md_uuid = self.uuid
+
+        def remove_stale_lvm():
+            """ Remove any stale LVM metadata that pre-existed in a new array's on-disk footprint. """
+            log.debug("waiting 5s for activation of stale lvm on new md array %s", self.path)
+            time.sleep(5)
+            udev.settle()
+            pvs_info.drop_cache()
+            pv_info = pvs_info.cache.get(self.path)
+            if pv_info is None:
+                return
+
+            if pv_info.vg_uuid:
+                log.info("removing stale LVM metadata found on %s", self.name)
+                try:
+                    blockdev.lvm.vgremove(pv_info.vg_name, extra={"--select": "vg_uuid=%s" % pv_info.vg_uuid})
+                except blockdev.LVMError as e:
+                    log.error("Failed to remove stale volume group from newly-created md array %s: %s",
+                              self.path, str(e))
+                    raise
+
+            # lvm says it is a pv whether or not there is vg metadata, so wipe the pv signature
+            blockdev.lvm.pvremove(self.path)
+
+        remove_stale_lvm()
 
     def _create(self):
         """ Create the device. """
--- a/tests/devices_test/device_methods_test.py
+++ b/tests/devices_test/device_methods_test.py
@@ -382,6 +382,8 @@ class MDRaidArrayDeviceMethodsTestCase(StorageDeviceMethodsTestCase):
         self.patchers["md"] = patch("blivet.devices.md.blockdev.md")
         self.patchers["is_disk"] = patch.object(self.device_class, "is_disk",
                                                 new=PropertyMock(return_value=False))
+        self.patchers["pvs_info"] = patch("blivet.devices.md.pvs_info")
+        self.patchers["lvm"] = patch("blivet.devices.md.blockdev.lvm")
 
     @property
     def teardown_method_mock(self):


blivet/devices/md.py
```diff
@@ -521,6 +523,40 @@ def pre_commit_fixup(self, current_fmt=False):
...
+
+        vg_name = pv_info.vg_name
+        vg_uuid = pv_info.vg_uuid
+        if vg_name and vg_uuid:
+            log.info("removing stale LVM metadata found on %s", self.name)
+            try:
+                # As of libblockdev-2.15 the vg name is passed straight to
+                # com.redhat.lvmdbus1.Manager.LookUpByLvmId, which returns
+                # a DBus object path based on vg name or UUID without the
+                # caller having to specify which is being passed.
```
vojtechtrefny: This unfortunately doesn't work with the "non-dbus" plugin. I think the right way to do this should be to use the extra args and pass the `--select vg_uuid=XXXX` option (so something like this -- https://gist.github.com/vojtechtrefny/782b0160471e341635a3a24467791fe3). The problem is that this doesn't work with the dbus plugin because it doesn't expect name to be NULL/None.

We either need to change the dbus plugin to extract the uuid in the [remove method](https://github.com/tasleson/lvm-dubstep/blob/master/lvmdbus/vg.py#L189) from the extra arguments, or change the "old" plugin to accept uuid instead of name (without using the extra arguments).

blivet/devices/md.py
```diff
@@ -521,6 +523,40 @@ def pre_commit_fixup(self, current_fmt=False):
...
+                blockdev.lvm.vgremove(vg_uuid, extra=["--force"])
```
vojtechtrefny: This is not needed -- both plugins use `--force` by default.

blivet/devices/md.py
vpodzime: Just FYI, changing *libblockdev* to accept UUIDs in place of PV/VG/LV paths/names has been a goal for a long time. I just haven't gotten to it yet.

dwlehman:
I wanted to change the call from this:

```python
blockdev.lvm.vgremove(vg_uuid)
```

to

```python
blockdev.lvm.vgremove(vg_name, extra={'--select': "vg_uuid=%s" % vg_uuid})
```

But the second call doesn't seem to work and I get a failure in the subsequent `pvcreate` call due to the stale metadata still being there. The first version works as expected, although it will not work w/ the old lvm plugin. Any hints @vpodzime or @tasleson ?

tasleson:
@dwlehman Not sure... although I'm starting to wonder if the lvm dbus daemon should be using the select syntax as the default, to resolve any ambiguity if it exists.  The lvm dbus daemon actually has both the name and uuid at the time the remove is done, it however only uses the name.

Just to clarify, adding the `select` in libblockdev works with lvmdbusd plugin, but not with the libblockdev plugin that directly calls the command line?

dwlehman:
Passing `--select vg_uuid=<UUID>` to the lvmdbusd plugin does not work. Pass the UUID instead of the name does.

tasleson:
I'm wondering if we are passing the correct `uuid` for the vg?  It appears that if you try to delete a VG and you pass a non-existent VG UUID you get success, but the VG doesn't actually get removed, an example.  **NOTE THE INCORRECT UUID**

``` bash
[root@aperture ~]# vgcreate testing /dev/sdg
  Volume group "testing" successfully created
[root@aperture ~]# vgs -oNAME,UUID
  VG      VG UUID                               
  testing TW5Mz3-KWiu-S2Xi-Icmb-Q4m4-Wp3z-vLVrg0
[root@aperture ~]# vgremove -f testing --select uuid=uuid=MbCD6I-gsaS-oGfB-ycoK-hhKS-Bilr-2E9tRp
[root@aperture ~]# echo $?
0
[root@aperture ~]# vgs
  VG      #PV #LV #SN Attr   VSize  VFree 
  testing   1   0   0 wz--n- <8.00g <8.00g
```

tasleson:
For the actual vg remove the libblkid needs to locate the object path and then call the method 'Remove' on it.  The look up function on the manager can look up by uuid or name to find the object path.

tasleson:
Note: I wasn't aware that lvm could have duplicate vg names.  At the moment I think lvm dbusd would handle this case badly.

tasleson:
We may want to incorporate something like this upstream

``` diff
diff --git a/daemons/lvmdbusd/cmdhandler.py b/daemons/lvmdbusd/cmdhandler.py
index 4fb1670a9..6bf03939a 100644
--- a/daemons/lvmdbusd/cmdhandler.py
+++ b/daemons/lvmdbusd/cmdhandler.py
@@ -270,10 +270,10 @@ def vg_rename(vg, new_name, rename_options):
        return call(cmd)
 
 
-def vg_remove(vg_name, remove_options):
+def vg_remove(vg_name, vg_uuid, remove_options):
        cmd = ['vgremove']
        cmd.extend(options_to_cli_args(remove_options))
-       cmd.extend(['-f', vg_name])
+       cmd.extend(['-f', vg_name, '--select', 'uuid=%s' % vg_uuid])
        return call(cmd)
 
 
diff --git a/daemons/lvmdbusd/vg.py b/daemons/lvmdbusd/vg.py
index c00084d7d..759190134 100644
--- a/daemons/lvmdbusd/vg.py
+++ b/daemons/lvmdbusd/vg.py
@@ -192,7 +192,7 @@ class Vg(AutomatedProperties):
                # Make sure we have a dbus object representing it
                Vg.validate_dbus_object(uuid, vg_name)
                # Remove the VG, if successful then remove from the model
-               rc, out, err = cmdhandler.vg_remove(vg_name, remove_options)
+               rc, out, err = cmdhandler.vg_remove(vg_name, uuid, remove_options)
                Vg.handle_execute(rc, out, err)
                return '/'
```

dwlehman:
The problem when using the `lvmdbusd` plugin was that it explicitly requires the old pv signature to be removed before writing a new one. So I added the `pvremove` call and it now works as expected. Thanks to @tasleson for pointing that out.

blivet/devices/md.py
```diff
@@ -534,6 +536,30 @@ def _post_create(self):
...
+            udev.settle()
+            pvs_info.drop_cache()
+            pv_info = pvs_info.cache.get(self.path)
+            if pv_info is None:
+                return
+
+            if pv_info.vg_uuid:
+                log.info("removing stale LVM metadata found on %s", self.name)
+                try:
+                    blockdev.lvm.vgremove(pv_info.vg_uuid)
```
vpodzime: So is this somehow working when the CLI-based LVM plugin is used by *libblockdev*?

dwlehman: No, I apparently pushed the wrong code.

dwlehman:
Latest version should work w/ both lvm plugins; tested and verified it fixes the related bug.

PR #653, opened 2018-02-01
Skip dmraid activation when already active

japokorn:
Fix of 1476513
dmraid activation no longer returns error when set is already active


vojtechtrefny:
The bug is probably not fixed, see https://bugzilla.redhat.com/show_bug.cgi?id=1476513#c39, but it definitely fixes some issues (see comment 42) so it might be good idea to merge this anyway.

vojtechtrefny:
Obsoleted by #784

PR #654, opened 2018-02-02
Account for pmspare/mlog volumes after growing a thin pool.

dwlehman:
Growable non-autopart (kickstart) thin pools don't get vg space for pmspare/mlog volumes. I have verified the fix after reproducing the original issue.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -1800,7 +1800,7 @@ class VGChunk(Chunk):
                              "LVRequest"))
 
         if req.device.metaDataSize:
-            self.pool -= int(self.vg.align(req.device.metaDataSize, roundup=True) / self.vg.peSize)
+            self.pool -= 2 * int(self.vg.align(req.device.metaDataSize, roundup=True) / self.vg.peSize)
 
         if req.device.cached:
             # cached LV -> reserve space for the cache
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -625,7 +625,7 @@ class LVMLogicalVolumeDevice(DMDevice):
         if self.cached:
             cache_size = self.cache.size
         return (self.vg.align(self.size, roundup=True) * self.copies
-                + self.logSize + self.metaDataSize + cache_size)
+                + self.logSize + 2 * self.metaDataSize + cache_size)
 
     def _setFormat(self, fmt):
         super(LVMLogicalVolumeDevice, self)._setFormat(fmt)


blivet/partitioning.py
```diff
@@ -2332,8 +2332,11 @@ def growLVM(storage):
         # now that we have grown all thin pools (if any), let's calculate and
         # set their metadata size if not told otherwise
         for pool in vg.thinpools:
-            if not pool.exists and pool.metaDataSize == Size(0):
-                pool.autoset_md_size()
+            if not pool.exists:
+                if pool.metaDataSize == Size(0):
+                    pool.autoset_md_size()
+                elif pool.req_grow:
+                    pool.size -= 2 * pool.metaDataSize
```
vpodzime: So this is for the case where metadata size set in kickstart and the pool is growable? Isn't the metadata part reserved in the VG already in that case? That would mean the above should be ``pool.size -= pool.metaDataSize`` to reserve the space for pmspare. Also, the VG should report the missing space for the pmspare on its own, right? (by its ``freeSpace`` property having negative value).

dwlehman:
Latest version is more focused and still works.

blivet/partitioning.py
```diff
@@ -1800,7 +1800,10 @@ def addRequest(self, req):
                              "LVRequest"))
 
         if req.device.metaDataSize:
-            self.pool -= int(self.vg.align(req.device.metaDataSize, roundup=True) / self.vg.peSize)
+            factor = 1
+            if hasattr(req.device, "lvs") and hasattr(req.device, "chunkSize"):  # thin pool
+                factor = 2
+            self.pool -= factor * int(self.vg.align(req.device.metaDataSize, roundup=True) / self.vg.peSize)
```
vpodzime: Is metadata specification supported for some other device than a thin pool? This new ``if`` looks quite confusing to me. And once the thin pool grows and the extra space for metadata and pmspare is reserved, the VG reports ``freeSpace == 0``, right? Otherwise some LV grown later could potentially eat that reserved space for its own growth.

dwlehman: Metadata size is an attribute of `LVMLogicalVolumeDevice` on rhel7-branch. I don't know what other types of LV might allow it, but I only know to double the space requirement for thin pools -- hence the complexity. Yes, the VG does report being completely full after growing the pool with this change.

vpodzime: > Yes, the VG does report being completely full after growing the pool with this change.

Good.

> Metadata size is an attribute of LVMLogicalVolumeDevice on rhel7-branch.

Yes, but thin pools are instances of the ``LVMThinPoolDevice`` class so I think the check can be made simpler/more straightforward.

dwlehman: So you think that even though the attr (`metaDataSize`) is in `LVMLogicalVolumeDevice` it is only used for thin pools? I wasn't sure if it was also used for something else like raid or cache lvs. (Or did I miss your point?)

vpodzime: No, it's a very good idea to not limit the calculation to thin pools. I just think the check is weird -- why not use ``isinstance()``? And I would say it's better to always reserve the pmspare space whenever there's some metadata space required by any LV. The LVM RAID can use it for reshaping/restriping/...

dwlehman:
Is that what you were suggesting? It's a bit more forgiving for other LV types w/ user-specified metadata size. Tested successfully.

dwlehman:
@vpodzime Does the latest version look more like what you expected?

PR #655, opened 2018-02-05
Name validity check fix

japokorn:
Name validity is ommitted for existing devices
Changes backported from commit a950e95

Resolves: rhbz#1325697


dwlehman:
It would be good to note that these changes are backported from previous commit a950e95d.

japokorn:
Updated commit message.

PR #663, opened 2018-02-27
Dasd is a valid label type on s390x (#1538550)

poncovka:
Add the dasd label type to the list of valid label types for s390x.
This bug was introduced in the commit ebd43b6.

Resolves: rhbz#1538550


vojtechtrefny:
Jenkins, test this please.

poncovka:
It is not possible to install F28 on s390x without this fix.

PR #664, opened 2018-03-06
Fixed KS forcing zerombr onto RO disk (#1544425)

japokorn:
The exception caused by using anaconda kickstart with zerombr and --only-use read-only disk
now produces a warning

Added tests for this behavior

Resolves: rhbz#1544425

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -874,8 +874,11 @@ class Blivet(object):
                 self.recursiveRemove(disk)
 
             if zerombr or should_clear:
-                log.debug("clearpart: initializing %s", disk.name)
-                self.initializeDisk(disk)
+                if disk.protected:
+                    log.warning("cannot clear '%s': disk is protected or read only", disk.name)
+                else:
+                    log.debug("clearpart: initializing %s", disk.name)
+                    self.initializeDisk(disk)
 
         self.updateBootLoaderDiskList()
 
--- a/tests/clearpart_test.py
+++ b/tests/clearpart_test.py
@@ -196,6 +196,80 @@ class ClearPartTestCase(unittest.TestCase):
         """
         pass
 
+    @mock.patch.object(blivet.Blivet, "removeEmptyExtendedPartitions")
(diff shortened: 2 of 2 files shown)


blivet/__init__.py
```diff
@@ -873,7 +873,11 @@ def clearPartitions(self):
             if should_clear:
                 self.recursiveRemove(disk)
 
-            if zerombr or should_clear:
+            if disk.readonly:
+                raise ProtectedDiskError("cannot clear '%s': disk is read only" % disk.name)
```
dwlehman: We shouldn't raise an exception here in the general case. Consider the following kickstart:

```zerombr
clearpart --initlabel --all
autopart
```

If there are several disks and only one of them is read-only, we should just not use the read-only disk. We already do this for `clearpart`, but not `zerombr`. I think the most you should do in these cases is log a warning and continue without initializing the disk. Also, I'm pretty sure that `disk.readonly` implies `disk.protected`, so you wouldn't need to check both of them.

japokorn: Fixed. I have checked that KS configuration. It behaves like you described. Apparently #636 does not apply for this situation. I have replaced the exception with a warning. Also removed `disk.readonly` and updated the tests accordingly.

blivet/__init__.py
```diff
@@ -873,7 +873,9 @@ def clearPartitions(self):
             if should_clear:
                 self.recursiveRemove(disk)
 
-            if zerombr or should_clear:
+            if disk.protected:
+                log.warning("cannot clear '%s': disk is protected or read only", disk.name)
```
dwlehman: If you placed this check in the `elif zerombr or should_clear:` branch it would only appear in the logs when relevant, ie: not when the read-only drive was properly excluded via `ignoredisk`.

japokorn: Fixed.

dwlehman:
The patch looks good, but you'll have to close this and open a pull request against 2.1-devel (and one against anaconda' master branch).

PR #665, opened 2018-03-06
Remove dependency on pocketlint

vojtechtrefny:
It is not needed during build. This was already removed on
2.1-devel but not completely removed on 3.0-devel.

----

We had the dependency multiple times on 3.0-devel and I didn't notice that when merging #633


vojtechtrefny:
New commit added. I'm not sure if this is the best solution, but it currently isn't possible to do a build without this change -- `update-po` runs the `xgettext_werror.sh` script from translation canary and because it isn't in the tarball the build always fail.

PR #668, opened 2018-03-08
Add option to get udev device by device node path

dwlehman:
The idea is to add an alternative way to get a `pyudev.Device` without breaking the API.


jkonecny12:
Could this be merged soon please? It is blocker on Anaconda side.

PR #672, opened 2018-03-20
Conditionalize the Python 2 subpackage and don't build it on EL > 7 and Fedora > 28

hroncok:
- Use bcond for with python3, allow it on RHEL > 7
- This reverts commit 243b84bb9629a09e77dcb9f8fb87414f5b679c33
- See https://src.fedoraproject.org/rpms/python-blivet/pull-request/2

### DIFF
--- a/python-blivet.spec
+++ b/python-blivet.spec
@@ -1,10 +1,19 @@
-%define is_rhel67 0%{?rhel} != 0 && 0%{?rhel} <= 7
+%define is_rhel 0%{?rhel} != 0
 
-# python3 is not available on RHEL6/7
-%if %{is_rhel67}
-%define with_python3 0
+# python3 is not available on RHEL <=7
+%if %{is_rhel} && 0%{?rhel} <= 7
+# disable python3 by default
+%bcond_with python3
 %else
-%define with_python3 1
+%bcond_without python3
+%endif
+
+# python2 is not available on RHEL > 7 and not needed on Fedora > 28
+%if 0%{?rhel} > 7 || 0%{?fedora} > 28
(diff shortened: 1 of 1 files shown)


python-blivet.spec
```diff
@@ -76,13 +85,21 @@ Requires: systemd-udev
 Requires: %{realname}-data = %{epoch}:%{version}-%{release}
 
 Obsoletes: blivet-data < 1:2.0.0
+
+%if %{without python2}
+# the version here needs to be updated to actual version when this happened
+Obsoletes: python2-blivet < 1:2.0.2-2
```
hroncok: Note that in my original PR this was 3.0.0-0.6, here the version in the spec is 2.0.2 but that is nonsense.

hroncok: Can be changed to `< %{version}-%{release}` - that works but is hostile towards potential return of the python2 package (in a module, etc.). I'd do that if such scenario is unlikely, let me know.

vojtechtrefny: We have a really complicated system of branches -- the right version is only in spec on `3.0-release` and we merge `3.0-devel` to `3.0-release` before every upstream release. Thanks to this we can merge bugfixes from `2.1-devel` to `2.2-devel` and `3.0-devel` without opening new pull requests or creating separate patches. As I said, it's very complicated and you can just ignore the version written here.

tl;dr version -- the version is ok, thank you for the changes.

hroncok:
This a bit time sensitive for us. Is there anything I can do to speed things up? Thanks.

hroncok:
Thanks

PR #674, opened 2018-03-22
Add toplevel tuple of fstypes supported by kernel

### LINKED ISSUE #673: [RFE] top level "supported filesystem formats" property
There is a `default_fstypes` property in `blivet.formats`.  It would be helpful if I could also query for a list of what types are actually supported.

https://github.com/storaged-project/blivet/blob/2.1-devel/blivet/formats/__init__.py#L64

jcpunk:
This resolves #673 without too much complexity.  Anaconda has a few additional concerns so this is largely based on `get_default_filesystem_type`.


jcpunk:
Any chance for a merge?

PR #675, opened 2018-03-23
Fixed error message when slave is missing

japokorn:
Now DeviceTreeError is raised when adding slave devices and
udev.get_device fails/returns None

(Related: #1519578)


japokorn:
Created a substitute PR #706 for merge to 3.0-devel. Closing this one.

PR #683, opened 2018-04-11
Replace deprecated iscsi_firmware dracut option

rvykydal:
The supported option is rd.iscsi.firmware

Resolves: rhbz#1566133


vojtechtrefny:
Jenkins, ok to test.

PR #686, opened 2018-04-16
WIP: Don't crash in populate when blockdev plugins are missing.

dwlehman:
This was thrown together as quickly as possible on Friday to see what it would take. Now I need to decide how to organize things for a final/real version. Using the existing availability tasks from the device classes is somewhat heavy-handed since we only need query functionality for populating the devicetree, so I am considering creating some new functions in devicelibs that call `BlockDev.foo_is_tech_avail()` to determine which functionality is available at runtime.


vpodzime:
I'd suggest using the [``DependencyGuard``](https://github.com/storaged-project/blivet/blob/2.1-devel/blivet/util.py#L1073) instances as decorators of the populator methods. Similarly to how [``iSCSIDependencyGuard``](https://github.com/storaged-project/blivet/blob/2.1-devel/blivet/iscsi.py#L106) is used in the [iscsi code](https://github.com/storaged-project/blivet/blob/2.1-devel/blivet/iscsi.py#L181)

vpodzime:
Nice, I really like this!

vpodzime:
> One thing still needed is to make it clear that these guards are only for query/populate as opposed to being general-purpose.

Yeah. Maybe there could be a new [``EvalMode``](https://github.com/storaged-project/blivet/blob/2.1-devel/blivet/util.py#L1067) like ``in_populate`` checking something like ``self.__class__._is_populate_guard`` with just these special new classes having that set to ``True``?

vojtechtrefny:
Obsoleted by #1306

PR #687, opened 2018-04-20
Fixed create_device fail behavior

japokorn:
- when `create_device` raises an exception during the format creation, previously
created device is now cleaned from the devicetree; the exception is then reraised
 - this prevents having the device with the same name already in the devicetree
when retrying the `create_device` command
 - changed dependency related exception in `DeviceAction.__init__` from `ValueError`
to `DependencyError` and updated the rest of code and the tests accordingly

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -36,7 +36,7 @@ from .deviceaction import ActionDestroyFormat, ActionResizeDevice, ActionResizeF
 from .devicelibs.edd import get_edd_dict
 from .devicelibs.btrfs import MAIN_VOLUME_ID
 from .devicelibs.crypto import LUKS_METADATA_SIZE
-from .errors import StorageError
+from .errors import StorageError, DependencyError
 from .size import Size
 from .devicetree import DeviceTree
 from .formats import get_default_filesystem_type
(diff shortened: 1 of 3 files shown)


blivet/blivet.py
```diff
@@ -791,7 +791,12 @@ def create_device(self, device):
         is_snapshot = isinstance(device, LVMLogicalVolumeDevice) and device.is_snapshot_lv
 
         if device.format.type and not device.format_immutable and not is_snapshot:
-            self.devicetree.actions.add(ActionCreateFormat(device))
+            try:
+                self.devicetree.actions.add(ActionCreateFormat(device))
+            except:  # pylint: disable=broad-except
+                # revert devicetree changes no matter what happened
+                self.reset_device(device)
```
dwlehman: This will cancel all actions scheduled for the device, which may not be warranted. A slightly less heavy-handed approach would be to remove only the actions you create in this method on failure. You would need to do a few things to make it work nicely, though:

1. change `DeviceAction.__init__` to raise `blivet.errors.DependencyError` instead of `ValueError`, then catch only that exception in `create_device`
2. be prepared for an exception to be raised late in `remove`, which you can safely ignore

blivet/blivet.py
```diff
@@ -791,7 +791,12 @@ def create_device(self, device):
         is_snapshot = isinstance(device, LVMLogicalVolumeDevice) and device.is_snapshot_lv
 
         if device.format.type and not device.format_immutable and not is_snapshot:
-            self.devicetree.actions.add(ActionCreateFormat(device))
+            try:
+                self.devicetree.actions.add(ActionCreateFormat(device))
+            except DependencyError:  # pylint: disable=broad-except
+                # revert devicetree changes no matter what happened
+                self.reset_device(device)
```
dwlehman: I think we talked about the possibility of canceling only the actions you created/scheduled just before in `create_device` instead of reverting all the way back to the original device. Is that possible?

japokorn: Fixed.

blivet/blivet.py
```diff
@@ -786,12 +786,22 @@ def create_device(self, device):
...
 
         if device.format.type and not device.format_immutable and not is_snapshot:
-            self.devicetree.actions.add(ActionCreateFormat(device))
+            action_create_fmt = None
+            try:
+                action_create_fmt = ActionCreateFormat(device)
+            except (ValueError, DependencyError) as e:
+                # revert devicetree changes done so far
+                if action_create_fmt is not None:
+                    self.devicetree.actions.remove(action_create_fmt)
```
dwlehman: This `remove` call isn't necessary and will trigger an exception, I believe, since `action_create_fmt` has not been added at this point.

dwlehman: Removing the `ActionCreateFormat` won't remove the device from the tree. (Removing the `ActionCreateDevice`, later, will.) You didn't add `action_create_fmt` yet, so there is no need to remove it.

japokorn: Ah. Now I see. You are right. My bad. Fixed.

PR #688, opened 2018-04-23
devicetree.names is now a property

japokorn:
- removed devicetree.names list
- introduced property with the same name and function
- devicetree.names are now obtained directly from devices
- added a test to check the proper behavior

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -44,6 +44,7 @@ from . import util
 from .populator import PopulatorMixin
 from .storage_log import log_method_call, log_method_return
 from .threads import SynchronizedMeta
+from .static_data import lvs_info
 
 import logging
 log = logging.getLogger("blivet")
@@ -93,9 +94,6 @@ class DeviceTreeBase(object):
         self._actions = ActionList(addfunc=self._register_action,
                                    removefunc=self._cancel_action)
 
-        # a list of all device names we encounter
-        self.names = []
-
         self._hidden = []
 
         lvm.lvm_cc_resetFilter()
@@ -144,6 +142,22 @@ class DeviceTreeBase(object):
 
         return devices
 
+    @property
+    def names(self):
+        """ List of devices names """
+        lv_info = list(lvs_info.cache.keys())
+
+        names = []
+        for dev in self._devices + self._hidden:
+            # don't include "req%d" partition names
+            if (dev.type != "partition" or not dev.name.startswith("req")) and \
+               dev.type != "btrfs volume" and \
+               dev.name not in names:
+                names.append(dev.name)
+
+        names.extend(n for n in lv_info if n not in names)
+        return names
+
     def _add_device(self, newdev, new=True):
         """ Add a device to the tree.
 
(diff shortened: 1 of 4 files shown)


blivet/devicetree.py
```diff
@@ -144,6 +141,11 @@ def devices(self):
 
         return devices
 
+    @property
+    def names(self):
+        """ List of devices names currently in the tree """
+        return [dev.name for dev in self.devices]
```
dwlehman: I think this needs to be something like `[dev.name for dev in self._devices + self._hidden]` to account for hidden devices. I wonder if it also needs some special list of names of LVM logical volumes that are part of incomplete/unusable volume groups to make sure we do not create new ones with in-use names.

japokorn: Fixed.

blivet/devicetree.py
```diff
@@ -144,6 +142,14 @@ def devices(self):
 
         return devices
 
+    @property
+    def names(self):
+        """ List of devices names """
+        lv_info = list(lvs_info.cache.keys())
+        names = [dev.name for dev in self.devices] + self._hidden
```
dwlehman: This should use `self._devices` (private/hidden) instead of `self.devices` (public) so it includes any incomplete devices that are present.

japokorn: Fixed.

blivet/devicetree.py
```diff
@@ -144,6 +142,14 @@ def devices(self):
 
         return devices
 
+    @property
+    def names(self):
+        """ List of devices names """
+        lv_info = list(lvs_info.cache.keys())
+        names = [dev.name for dev in self._devices] + self._hidden
```
dwlehman: `self._hidden` is a list of devices just like `self._devices`. I think you want `+ self._hidden` to be inside the list comprehension rather than outside it.

dwlehman:
It seems like a good idea to write some basic tests that add devices, hide some of them, then verify that the names are still in `devicetree.names`.

japokorn:
Added the tests.

tests/devicetree_test.py
```diff
@@ -57,15 +58,39 @@ def test_resolve_device(self):
...
+        # * hide sda
+        # * hide and unhide again sdb
+        # * leave sdc unchanged
+        tree.hide(tree.get_device_by_name("sda"))
+        tree.hide(tree.get_device_by_name("sdb"))
+
+        tree.unhide(tree.get_device_by_name("sdb", hidden=True))
+
+        # some lvs names may be already present in the system
+        lv_info = list(lvs_info.cache.keys())
```
dwlehman: The tests will often be run as a non-root user, in which case the lvm cache would be empty. It would be better to use `mock` to fake the lvm data for your test.

japokorn: Fixed by mocking `lvs_info.cache` property.

vojtechtrefny:
Looks like all the comments from @dwlehman were resolved, can you please rebase this to current 3.1-devel? 
Do we need to change and/or test something in Anaconda before merging this?

japokorn:
jenkins, test this please

japokorn:
I have tested the change with anaconda. Look good.

blivet/devicetree.py
```diff
@@ -165,12 +171,6 @@ def _add_device(self, newdev, new=True):
         newdev.add_hook(new=new)
         self._devices.append(newdev)
 
-        # don't include "req%d" partition names
-        if ((newdev.type != "partition" or
-             not newdev.name.startswith("req")) and
-                newdev.type != "btrfs volume" and
-                newdev.name not in self.names):
-            self.names.append(newdev.name)
```
dwlehman: It looks like the property drops these special cases (unallocated partitions w/ names like 'req1', btrfs volumes w/ same name as first backing device). It should be fairly simple to preserve at least the partition one.

japokorn: Fixed. Instead of removing, the code was moved into property handling function.

PR #691, opened 2018-05-11
Fix of partitioning test fail

japokorn:
fixed issue when partitioning_test failed when run after
dependencies_test


japokorn:
Better way to fix this is by #692

PR #692, opened 2018-05-14
Fixups for missing dependency tests.

dwlehman:
These changes fix the partitioning test error and allow running (most of) these tests as non-root.

This is an alternative to #691.


japokorn:
jenkins, test this please

dwlehman:
Added a new commit that fixes the caching logic in `blivet.tasks.availability.ExternalResource`. All tests pass locally for me with this change.

PR #693, opened 2018-05-15
Disable sparse inode allocation for XFS /boot (#1575797)

AdamWill:
This attempts to disable sparse inode allocation by default for
XFS filesystems mounted at /boot, as a workaround for #1575797
until grub is fixed to be able to read such XFS filesystems.

Signed-off-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -1086,6 +1086,18 @@ class XFS(FS):
 
         super(XFS, self).write_uuid()
 
+    def create(self, **kwargs):
+        """Until https://bugzilla.redhat.com/show_bug.cgi?id=1575797
+        is resolved, this tries to ensure that sparse inodes are not
+        used if the mount point of the filesystem is /boot. We won't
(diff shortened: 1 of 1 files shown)


AdamWill:
I tested this just by doing an install with an updates.img containing the patch and creating /boot as xfs (using blivet-gui custom partitioning). It worked as intended: called mkfs.xfs with `-i sparse=0`, and that worked (the bootloader install succeeded and the installed system booted).

sandeen:
I've sent a [grub patch](https://lists.gnu.org/archive/html/grub-devel/2018-05/msg00057.html) to teach it to handle the new format.  Better to fix the damage there than to route around it here, IMHO.

dwlehman:
@sandeen great -- thanks!

AdamWill:
Agreed, that's much better, I was only working on this based on the suggestion that the upstream issue may not be fixed quickly. Let's close the PR.

PR #694, opened 2018-05-15
NVDIMM support

vojtechtrefny:
#681 for rhel7-branch

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -55,7 +55,7 @@ class DiskDevice(StorageDevice):
     def __init__(self, name, fmt=None,
                  size=None, major=None, minor=None, sysfsPath='',
                  parents=None, serial=None, vendor="", model="", bus="",
-                 exists=True):
+                 uuid=None, exists=True):
         """
             :param name: the device name (generally a device node's basename)
             :type name: str
@@ -86,7 +86,7 @@ class DiskDevice(StorageDevice):
                                major=major, minor=minor, exists=exists,
                                sysfsPath=sysfsPath, parents=parents,
                                serial=serial, model=model,
-                               vendor=vendor, bus=bus)
+                               vendor=vendor, bus=bus, uuid=uuid)
 
     def __repr__(self):
         s = StorageDevice.__repr__(self)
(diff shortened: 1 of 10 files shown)


vojtechtrefny:
Jenkins, test this please.

blivet/nvdimm.py
```diff
@@ -0,0 +1,181 @@
...
+                                          BlockDev.NVDIMMTechMode.QUERY |
+                                          BlockDev.NVDIMMTechMode.ACTIVATE_DEACTIVATE)
+        except GLib.GError:
+            return False
+        return True
+
+blockdev_nvdimm_required = NVDIMMDependencyGuard()
+
+
+log_bd_message = lambda level, msg: program_log.info(msg)
```
dwlehman: This was done on line 33 already.

blivet/nvdimm.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self._namespaces = dict((namespace.dev, namespace) for namespace in namespaces)
+
+    def get_namespace_info(self, device):
+        """ Get namespace information for a device
+            :param str device: device name (e.g. 'pmem0') or path
+        """
+        for info in self.namespaces.values():
+            if info.blockdev == device or \
+               (device.startswith("/dev/") and info.blockdev == device[5:]):
```
dwlehman: You could simplify it slightly by doing `if info.blockdev == device.split("/")[-1]` or `if info.blockdev == os.path.basename(device)`. Maybe not worthwhile if this code is already on 3.1-devel.

blivet/devices/disk.py
```diff
@@ -613,3 +613,49 @@ def dracutSetupArgs(self):
...
+            :type exists: bool
+            :keyword size: the device's size
+            :type size: :class:`~.size.Size`
+            :keyword parents: a list of parent devices
+            :type parents: list of :class:`StorageDevice`
+            :keyword format: this device's formatting
+            :type format: :class:`~.formats.DeviceFormat` or a subclass of it
+            :keyword mode: mode of the namespace
+            :type mode: str
+            :keyword devname: name of the namespace (e.g. 'namespace0.0')
```
dwlehman: How is `devname` different from `name`? Is the former more like a label?

blivet/nvdimm.py
vojtechtrefny: Fixed.

blivet/devices/disk.py
vojtechtrefny: Name is something like `pmem0` (and it is what udev see as name) and devname is `namespace0.0` (this is nvdimm/ndctl "internal" name). devname is a "unique" identifier for the nvdimm "device". For example name depends on the current mode -- after reconfiguring the namespace to sector mode, the name will be changed to `pmem0s`. `namespace0.0` can be used also for nvdimms not visible by udev (without a block device, e.g. in memory mode).

vojtechtrefny:
> It would be nice if the dependency guard backport was a separate commit.

Fixed.

> Is there anything we need to do for lorax?

I hope not. This adds dependency on newest libblockdev and it depends on ndctl(-devel), but I'll check with Radek to be 100 % sure.

PR #696, opened 2018-05-21
The key size of LUKS should be 0 by default

poncovka:
If the cipher is specified, then the key size is None by default
and liblockdev fails to create the format, because it expects 0 by
default. Because of that, Anaconda fails to do the partitioning
if the kickstart option --cipher is specified.

### DIFF
--- a/blivet/formats/luks.py
+++ b/blivet/formats/luks.py
@@ -96,7 +96,7 @@ class LUKS(DeviceFormat):
         self._resize = self._resize_class(self)
 
         self.cipher = kwargs.get("cipher")
-        self.key_size = kwargs.get("key_size")
+        self.key_size = kwargs.get("key_size") or 0
         self.map_name = kwargs.get("name")
 
         if not self.exists and not self.cipher:


blivet/formats/luks.py
```diff
@@ -105,6 +105,9 @@ def __init__(self, **kwargs):
                 # default to the max (512 bits) for aes-xts
                 self.key_size = 512
 
+        if not self.exists and self.key_size is None:
+            self.key_size = 0
+
```
dwlehman: Maybe we should instead change the `kwargs.get("key_size)` to `kwargs.get("key_size") or 0` if the default of None isn't useful.

vpodzime: You can do ``kwargs.get("key_size", 0)``

poncovka: Fixed.

dwlehman: Using `dict.get` alone will work as long as nobody passes `key_size=None`.

poncovka: Fixed.

vpodzime: > Using dict.get alone will work as long as nobody passes key_size=None.

Good point!

vojtechtrefny:
Jenkins, ok to test.

poncovka:
One test has failed, but I don't think that it is related to this fix:
```
FAIL: test_add_partition (partitioning_test.PartitioningTestCase)
_ped.PartitionException: Could not create partition /tmp/blivet.130dbhcn-addparttest-1

During handling of the above exception, another exception occurred:

  File "/home/jenkins/workspace/blivet-3.x-PR/tests/partitioning_test.py", line 226, in test_add_partition
    Size("10 MiB"))
AssertionError: "no extended partition" does not match "Could not create partition /tmp/blivet.130dbhcn-addparttest-1"

```

PR #697, opened 2018-05-23
LUKS2 support

vojtechtrefny:
This is first part of luks2 support for blivet. These changes solve problem with luks2+integrity -- blivet currently expects that the luks device is child of the partition with luks format and this is not true with integrity devices.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -33,6 +33,7 @@ from .devicelibs import btrfs
 from .devicelibs import mdraid
 from .devicelibs import lvm
 from .devicelibs import raid
+from .devicelibs import crypto
 from .partitioning import SameSizeSet
 from .partitioning import TotalSizeSet
 from .partitioning import do_partitioning
@@ -314,6 +315,8 @@ class DeviceFactory(object):
             :keyword min_luks_entropy: minimum entropy in bits required for
                                        LUKS format creation
             :type min_luks_entropy: int
+            :keyword luks_version: luks format version ("luks1" or "luks2")
+            :type luks_version: str
 
         """
         self.storage = storage  # a Blivet instance
@@ -328,6 +331,7 @@ class DeviceFactory(object):
 
         self.fstype = None  # not included in default_settings b/c of special handling below
         self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
+        self.luks_version = kwargs.get("luks_version") or crypto.DEFAULT_LUKS_VERSION
 
         # If a device was passed, update the defaults based on it.
         self.device = kwargs.get("device")
@@ -674,6 +678,7 @@ class DeviceFactory(object):
             fstype = "luks"
             mountpoint = None
             fmt_args["min_luks_entropy"] = self.min_luks_entropy
+            fmt_args["luks_version"] = self.luks_version
         else:
             fstype = self.fstype
             mountpoint = self.mountpoint
@@ -811,7 +816,8 @@ class DeviceFactory(object):
             orig_device = self.device
             leaf_format = self.device.format
             self.storage.format_device(self.device, get_format("luks",
-                                                               min_luks_entropy=self.min_luks_entropy))
+                                                               min_luks_entropy=self.min_luks_entropy,
+                                                               luks_version=self.luks_version))
             luks_device = LUKSDevice("luks-%s" % self.device.name,
                                      fmt=leaf_format,
                                      parents=self.device)
@@ -1146,7 +1152,8 @@ class PartitionSetFactory(PartitionFactory):
             if not member_encrypted and self.encrypted:
                 members.remove(member)
                 self.storage.format_device(member, get_format("luks",
-                                                              min_luks_entropy=self.min_luks_entropy))
+                                                              min_luks_entropy=self.min_luks_entropy,
+                                                              luks_version=self.luks_version))
                 luks_member = LUKSDevice("luks-%s" % member.name,
                                          parents=[member],
                                          fmt=get_format(self.fstype))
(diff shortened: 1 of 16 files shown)


vojtechtrefny:
Added support for creating luks2 format.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
What is the best place for "input" checking for the pbkdf extra arguments? There are two potential issues:

- `max_memory_kb` is not valid option/argument for *pbkdf2* (it makes sense only for *argon2i/argon2id*) -- if you specify `max_memory_kb` for *pbkdf2* libblockdev will ignore it -- shoud we ignore it too, return a ValueError or do the check only in Anaconda during kickstart validation (or do both?)
- same question for `iterations` and `time_ms` -- these are mutually exclusive and if you specify both, `time_ms` will be ignored by cryptsetup

dwlehman:
In blivet, I think it should be sufficient to log a warning in either case. I would expect anaconda to do at least as much -- probably raise an exception -- but that is up to them, of course.

vojtechtrefny:
Ok, I have added the warnings.
This should be ready for review now. It is possible that we'll need to change few things later (after anaconda starts to implement this) but I don't plan to add more things now.

vojtechtrefny:
Jenkins, test this please.

blivet/udev.py
```diff
@@ -616,6 +616,17 @@ def device_is_dm_luks(info):
     return is_crypt and _type.startswith("luks")
 
 
+def device_is_dm_integrity(info):
+    """ Return True if the device is a mapped integrity device. """
+    is_crypt = device_dm_subsystem_match(info, "crypt")
+    try:
```
vpodzime: Why not return `False` here early `if not is_crypt`?

blivet/formats/luks.py
```diff
@@ -135,6 +143,15 @@ def __init__(self, **kwargs):
             elif "discard" not in self.options:
                 self.options += ",discard"
 
+        self.pbkdf_args = kwargs.get("pbkdf_args", None)
```
vpodzime: `None` is not needed here, it's the default.

poncovka:
Hi, you need to update the `populate_ksdata` method. See: https://github.com/poncovka/blivet/commit/98470ceab88138e6677537a3dc9bfcccc5cd8e33

blivet/devices/luks.py
```diff
@@ -140,3 +140,34 @@ def populate_ksdata(self, data):
         self.slave.populate_ksdata(data)
         data.encrypted = True
         super(LUKSDevice, self).populate_ksdata(data)
+
+
+class IntegrityDevice(DMIntegrityDevice):
```
dwlehman: So it is possible (expected?) to have dm-integrity devices that are not associated with LUKS2 devices?

blivet/formats/luks.py
```diff
@@ -250,7 +256,8 @@ def _create(self, **kwargs):
                                     key_file=self._key_file,
                                     cipher=self.cipher,
                                     key_size=self.key_size,
-                                    min_entropy=self.min_luks_entropy)
+                                    min_entropy=self.min_luks_entropy,
+                                    luks_version=crypto.LUKS_VERSIONS[self.luks_version])
```
dwlehman: Ideally, `luks_version` would be a property with the check from `__init__` moved into the setter.

blivet/formats/luks.py
```diff
@@ -241,6 +241,12 @@ def _teardown(self, **kwargs):
         log.debug("unmapping %s", self.map_name)
         blockdev.crypto.luks_close(self.map_name)
 
+    def _pre_resize(self):
+        if self.luks_version == "luks2" and not self.__passphrase and not self.has_key:
```
dwlehman: `has_key` checks for both a passphrase and a key file.

blivet/devicefactory.py
```diff
@@ -328,6 +331,7 @@ def __init__(self, storage, **kwargs):
 
         self.fstype = None  # not included in default_settings b/c of special handling below
         self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
+        self.luks_version = kwargs.get("luks_version") or crypto.DEFAULT_LUKS_VERSION
```
dwlehman: Couldn't this just fall back to `None` and let the format supply the default?

blivet/formats/luks.py
```diff
@@ -135,6 +143,15 @@ def __init__(self, **kwargs):
             elif "discard" not in self.options:
                 self.options += ",discard"
 
+        self.pbkdf_args = kwargs.get("pbkdf_args")
+        if self.pbkdf_args:
+            if self.luks_version != "luks2":
+                raise ValueError("PBKDF arguments are valid only for LUKS version 2.")
+            if self.pbkdf_args.time_ms and self.pbkdf_args.iterations:
+                log.warning("Both iterations and specified for PBKDF, number of iterations will be ignored.")
```
dwlehman: "Both iterations and ___time_ms___ specified for ..."

blivet/devices/luks.py
dwlehman: I found out later that this is indeed possible.

vojtechtrefny:
> It seems like the max memory setting might be nicer to use if it were in the static crypto data singleton.

It probably isn't a "normal" use case but it is possible to create multiple luks devices with different pbkdf settings.

blivet/devicefactory.py
```diff
@@ -332,6 +334,7 @@ def __init__(self, storage, **kwargs):
         self.fstype = None  # not included in default_settings b/c of special handling below
         self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
         self.luks_version = kwargs.get("luks_version") or crypto.DEFAULT_LUKS_VERSION
+        self.pbkdf_args = kwargs.get("pbkdf_args", None)
```
dwlehman: `None` is the default fallback value for `dict.get`.

blivet/static_data/luks_data.py
```diff
@@ -64,6 +66,14 @@ def luks_devs(self):
...
         self.__luks_devs = value
 
+    @property
+    def pbkdf_args(self):
+        return self.__pbkdf_args
+
+    @pbkdf_args.setter
+    def pbkdf_args(self, value):
+        self.__pbkdf_args = value
+
```
dwlehman: I think a single underscore would suffice for this. I used two for passphrases because of the obfuscation required to access the attributes directly since they are passphrases. Maybe I'm wrong, but my sense is that the PBKDF args don't require such measures.

blivet/formats/luks.py
```diff
@@ -280,6 +281,16 @@ def _create(self, **kwargs):
                         type=self.type, status=self.status)
         super(LUKS, self)._create(**kwargs)  # set up the event sync
 
+        if not self.pbkdf_args:
```
poncovka: You should also check the version of LUKS, otherwise liblockdev might fail for LUKS1.

vojtechtrefny: Fixed.

PR #700, opened 2018-06-08
Use ID_WWN_WITH_EXTENSION for WWNs.

dwlehman:
Also strip off the '0x' prefix, which just takes up space.


dwlehman:
This is required for rhinstaller/anaconda#1503 (and replaces the previous blivet pull request related to it).

PR #701, opened 2018-06-13
Make sure WWNs are set for multipaths

dwlehman:
The fact that ID_WWN is not set for multipath devices was made apparent in rhinstaller/anaconda#1503.


dwlehman:
Merged into #700.

PR #705, opened 2018-06-25
Pylint fixes

vojtechtrefny:
New pylint, new errors to fix.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -606,7 +606,7 @@ def do_partitioning(storage, boot_disk=None):
     try:
         allocate_partitions(storage, disks, partitions, free, boot_disk=boot_disk)
         grow_partitions(disks, partitions, free, size_sets=storage.size_sets)
-    except Exception:
+    except Exception:  # pylint: disable=try-except-raise
         raise
     else:
         # Mark all growable requests as no longer growable.
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -410,7 +410,7 @@ class PopulatorMixin(object):
         parted.register_exn_handler(parted_exn_handler)
         try:
             self._populate()
-        except Exception:
+        except Exception:  # pylint: disable=try-except-raise
             raise
         finally:
             parted.clear_exn_handler()
--- a/blivet/devicelibs/edd.py
+++ b/blivet/devicelibs/edd.py
@@ -420,8 +420,8 @@ class EddMatcher(object):
 
         answers = []
         for path in pattern.glob():
-            emptyslash = util.Path("/", self.root)
-            path = util.Path(path, self.root)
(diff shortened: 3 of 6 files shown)


tests/pylint/runpylint.py
```diff
@@ -16,7 +16,8 @@ def __init__(self):
                                FalsePositive(r"Method 'do_task' is abstract in class 'UnimplementedTask' but is not overridden"),
```
vpodzime: *create* in the commit message.

blivet/populator/populator.py
```diff
@@ -410,7 +410,7 @@ def populate(self, cleanup_only=False):
         parted.register_exn_handler(parted_exn_handler)
         try:
             self._populate()
-        except Exception:
+        except Exception:  # pylint: disable=try-except-raise
```
vpodzime: This can be just dropped.

blivet/partitioning.py
```diff
@@ -606,7 +606,7 @@ def do_partitioning(storage, boot_disk=None):
     try:
         allocate_partitions(storage, disks, partitions, free, boot_disk=boot_disk)
         grow_partitions(disks, partitions, free, size_sets=storage.size_sets)
-    except Exception:
+    except Exception:  # pylint: disable=try-except-raise
```
vpodzime: Dropping the ``except`` section here and the ``else:`` keyword would keep the same behavior, AFAIK.

tests/pylint/runpylint.py
vojtechtrefny: Fixed.

blivet/partitioning.py
```diff
@@ -606,9 +606,7 @@ def do_partitioning(storage, boot_disk=None):
     try:
         allocate_partitions(storage, disks, partitions, free, boot_disk=boot_disk)
         grow_partitions(disks, partitions, free, size_sets=storage.size_sets)
-    except Exception:
-        raise
-    else:
+
```
dwlehman: Hmm, I prefer the original form. The intention is clearer. Can we just do a whitelist/disable comment instead?

tests/pylint/runpylint.py
dwlehman: I still see "create" in the commit message.

blivet/partitioning.py
vojtechtrefny: Yes, it is possible to disable the warning, fixed.

tests/pylint/runpylint.py
vojtechtrefny: I originally had *create**s*** there and I've changed it to *create*.

dwlehman: Sorry, for some reason I misunderstood it when I read it yesterday. It does indeed look good as it is.

PR #706, opened 2018-07-10
Fixed error message when slave is missing

japokorn:
Now DeviceTreeError is raised when adding slave devices and
udev.get_device fails/returns None

(Related: #1519578)


japokorn:
This is a replacement PR for approved #675. The only difference is in merging into 3.0-devel instead of 2.1-devel.

japokorn:
jenkins, test this please

PR #699, opened 2018-06-01
Rhel7 branch nvdimm devices handling in installer mode

rvykydal:
Depends on https://github.com/storaged-project/blivet/pull/694

PR #707, opened 2018-07-18
Remove NVDIMM support on non-x86_64

vojtechtrefny:
This basically ignores nvdimm devices on systems without libblockdev-nvdimm . In installer mode, all `pmem` devices will be ignored, outside of installer existing NVDIMM namespaces in sector mode will be treated as normal disks.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -2134,6 +2134,10 @@ class Blivet(object):
         :rtype: set(str)
         """
 
+        if not self.nvdimm.nvdimm_plugin_available:
+            log.debug("NVDIMM is not supported, all NVDIMM devices are ignored via a udev blacklist")
+            return set()
+
         ks_allowed_namespaces = set()
         ks_allowed_blockdevs = set()
         if nvdimm_ksdata:
--- a/blivet/nvdimm.py
+++ b/blivet/nvdimm.py
@@ -23,6 +23,7 @@ gi.require_version("GLib", "2.0")
 from gi.repository import BlockDev
 from gi.repository import GLib
 
+from . import arch
 from . import util
 
 import logging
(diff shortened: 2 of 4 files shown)


blivet/udev.py
```diff
@@ -28,19 +28,25 @@
...
 
 import logging
 log = logging.getLogger("blivet")
 
-INSTALLER_BLACKLIST = (r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram', r'ndblk')
+INSTALLER_BLACKLIST = [r'^mtd', r'^mmcblk.+boot', r'^mmcblk.+rpmb', r'^zram', r'ndblk']
 """ device name regexes to ignore when flags.installer_mode is True """
+# we don't have NVDIMM plugin, just ignore all pmem devices too
+if not nvdimm.nvdimm_plugin_available:
+    INSTALLER_BLACKLIST.append(r'pmem')
```
dwlehman: Will it be safer if we use `r'^pmem'` instead? (I think we should have done the same for 'ndblk' above as well, but we didn't.)

blivet/nvdimm.py
```diff
@@ -33,21 +34,24 @@
...
+    try:
+        # do not check for dependencies during libblockdev initializtion, do runtime
+        # checks instead
+        BlockDev.switch_init_checks(False)
+        succ_, avail_plugs = BlockDev.try_reinit(require_plugins=_REQUESTED_PLUGINS,
+                                                 reload=False,
+                                                 log_func=log_bd_message)
+    except GLib.GError as err:
+        log.error("Failed to intialize the libblockdev library: %s", err)
+        _HAVE_NVDIMM = False
```
dwlehman: This should be redundant since you initialized it above and (I think) this only runs once.

blivet/udev.py
```diff
@@ -28,19 +28,25 @@
...
+from .nvdimm import nvdimm
 
 import pyudev
 global_udev = pyudev.Context()
 
 import gi
 gi.require_version("BlockDev", "2.0")
+gi.require_version("GLib", "2.0")
 from gi.repository import BlockDev as blockdev
+from gi.repository import GLib
```
dwlehman: As jenkins pointed out this isn't being used (nor is the `require_version` for it, above)

blivet/udev.py
vojtechtrefny: Probably, the nvdimm devices names should look like `pmemX` or `pmemXs`. We originally had only `pmem` there (https://github.com/storaged-project/blivet/pull/699/files#diff-48651c0ada40d943223dea89af3e129aL38). I'm not sure if there was a reason for that.

blivet/__init__.py
```diff
@@ -2134,6 +2134,9 @@ def getIgnoredNvdimmBlockdevs(self, nvdimm_ksdata):
         :rtype: set(str)
         """
 
+        if not self.nvdimm.nvdimm_plugin_available:
```
rvykydal: I'd maybe add log.debug message saying we are returning empty set because the plugin is not available?

vojtechtrefny: Fixed.

rvykydal:
Thanks for the log message, still looks good to me.

blivet/udev.py
dwlehman: I think it was just an oversight. It should have been `r'^pmem'`.

dwlehman:
Looks good to me, better still if you tweak that blacklist string as discussed.

blivet/udev.py
vojtechtrefny: Ok, fixed.

PR #709, opened 2018-07-26
Do not ignore "Image out-of-sync" internal LVs

vojtechtrefny:
These are still "valid" existing internal RAID LVs just not synced
yet.

We are using internal lvs to calculate `free_space` property of vgs. If we ignore these we will report wrong free space, see https://github.com/storaged-project/blivet-gui/issues/107

There is also an issue in lvmdbus (https://bugzilla.redhat.com/show_bug.cgi?id=1608930) -- it doesn't update the Attr property of the LVs so we ignore all internal lvs for raid LVs (because the type is never changed from *(I)mage out-of-sync* to *(i)mage, mirror or raid*).


vojtechtrefny:
New PR for 3.0-devel: #710

PR #711, opened 2018-08-02
Fixed various issues preventing successful build

japokorn:
- now marked as pylint false positive: non-existing 'id' attribute of ObjectID and its children
- fixed sudden wild ksdata variable reappearance caused by merging


japokorn:
Moved PR to 3.1-devel

PR #716, opened 2018-08-16
Create a separate availability check for dmraid support

vojtechtrefny:
Libblockdev can now be build without dmraid support so we need
a separate check for dmraid and 'other' dm devices support.

Resolves: rhbz#1617958

### DIFF
--- a/blivet/tasks/availability.py
+++ b/blivet/tasks/availability.py
@@ -331,9 +331,13 @@ BLOCKDEV_DM_ALL_MODES = (blockdev.DMTechMode.CREATE_ACTIVATE |
                          blockdev.DMTechMode.QUERY)
 BLOCKDEV_DM = BlockDevTechInfo(plugin_name="dm",
                                check_fn=blockdev.dm_is_tech_avail,
-                               technologies={blockdev.DMTech.MAP: BLOCKDEV_DM_ALL_MODES,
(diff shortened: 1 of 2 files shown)


blivet/tasks/availability.py
```diff
@@ -399,6 +403,7 @@ def available_resource(name):
 BLOCKDEV_BTRFS_PLUGIN = blockdev_plugin("btrfs", BLOCKDEV_BTRFS_TECH)
 BLOCKDEV_CRYPTO_PLUGIN = blockdev_plugin("crypto", BLOCKDEV_CRYPTO_TECH)
 BLOCKDEV_DM_PLUGIN = blockdev_plugin("dm", BLOCKDEV_DM_TECH)
+BLOCKDEV_DM_PLUGIN_RAID = blockdev_plugin("dm", BLOCKDEV_DM_TECH_RAID)
```
dwlehman: It would be a bit clearer if you named it `BLOCKDEV_DM_RAID_PLUGIN` instead.

dwlehman: But I guess I understand if you don't want to imply that there is a separate plugin for dmraid, since there isn't.

vojtechtrefny: It's just a part of the dm plugin, so I think it makes more sense to name it `BLOCKDEV_DM_PLUGIN_RAID`

PR #721, opened 2018-09-05
 Deactivate partially hidden VGs during installation

vojtechtrefny:
Alternative solution by forcing `vgdeactivate` in `_pre_destroy` -- https://github.com/vojtechtrefny/blivet/commit/fedb62320a5169635733910276f373209676177f -- but I think this way is safer.

Also this currently doesn't work -- we need to run the `pvremove` with `--force --force` but this is currently broken in libblockdev.

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -862,7 +862,9 @@ class DeviceTreeBase(object):
         if not device.exists:
             return
 
-        self._remove_device(device, force=True, modparent=False)
+        # XXX set modparent=True when hiding a VG to unset vg_name from PVs
+        modparent = device.type == "lvmvg"
+        self._remove_device(device, force=True, modparent=modparent)
 
         self._hidden.append(device)
         lvm.lvm_cc_addFilterRejectRegexp(device.name)
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -390,6 +390,19 @@ class PopulatorMixin(object):
             loop_device = dm_device.parents[0]
             loop_device.teardown()
 
+    def _teardown_partially_hidden_vgs(self):
+        """ Teardown VGs that are not completely hidden (one of the PVs
+            is not hidden)
+        """
+
+        for device in self._hidden:
+            if device.type == "lvmvg":
+                if not all(pv in self._hidden for pv in device.pvs):
+                    log.info("Deactivating partially hidden VG '%s'", device.name)
+                    # use _teardown() to avoid _pre_teardown() because it prevents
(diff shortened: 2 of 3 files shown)


blivet/devicetree.py
```diff
@@ -862,7 +862,9 @@ def hide(self, device):
         if not device.exists:
             return
 
-        self._remove_device(device, force=True, modparent=False)
+        # XXX set modparent=True when hiding a VG to unset vg_name from PVs
+        modparent = device.type == "lvmvg"
+        self._remove_device(device, force=True, modparent=modparent)
```
dwlehman: This chunk makes me nervous, both because I fear it may break things when users repeatedly hide/unhide disks and because I don't understand why we need to unset vg_name in the pvs.

vojtechtrefny: > it may break things when users repeatedly hide/unhide disks

`add_hook` for `LVMVolumeGroupDevice` sets the `vg_name` on all PVs so it shouldn't be a problem. But it definitely can cause some problems in the future.

> I don't understand why we need to unset vg_name in the pvs

Becase PV format `status` is `True` when `vg_name` is set and we don't allow removing "active" formats (`DeviceFormat._pre_destroy`) raises an exception.

blivet/populator/populator.py
```diff
@@ -390,6 +390,17 @@ def teardown_disk_images(self):
...
+    def _teardown_partially_hidden_vgs(self):
+        """ Teardown VGs that are not completely hidden (one of the PVs
+            is not hidden)
+        """
+
+        for device in self._hidden:
+            if device.type == "lvmvg":
+                if not all(pv in self._hidden for pv in device.pvs):
+                    log.info("Deactivating partially hidden VG '%s'", device.name)
+                    device._teardown()
```
vpodzime: Why `_teardown()` and not `teardown()`?

vojtechtrefny: Because we can't run `StorageDevice._pre_teardown()` -- it checks VG `status` and we have "missing" PVs so it's False and because of this `_pre_teardown()` will return before actually running `_teardown()`.

vpodzime: That's worth a comment in the code, I believe.

vojtechtrefny: Ok, comment added.

vojtechtrefny:
Closing this. We've decided that this use case shouldn't be supported and selecting only one of the PVs in kickstart shouldn't be allowed -- so anaconda should behave the same way it does in the GUI (storage spoke doesn't allow to select only one of the PVs).

PR #722, opened 2018-09-12
Allow removing btrfs volumes without btrfs support

vojtechtrefny:
Btrfs volumes are removed using wipefs so we don't need to check
for device dependencies availability when removing the volume
(btrfs support depends on libblockdev btrfs plugin).

Resolves: rhbz#1605213

----

Fixing this in `DeviceAction` is definitely not ideal, but I don't think there is a better way -- this would require a quite radical rewrite of the availability code to be able to define different dependencies for adding, removing and modifying devices. And right now this problem exists only for Btrfs devices.

### DIFF
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -162,7 +162,8 @@ class DeviceAction(util.ObjectID):
 
         self.device = device
 
-        self._check_device_dependencies()
+        if self.is_device:
+            self._check_device_dependencies()
 
         self.container = getattr(self.device, "container", None)
         self._applied = False
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -160,14 +160,18 @@ class DeviceAction(util.ObjectID):
         if not isinstance(device, StorageDevice):
             raise ValueError("arg 1 must be a StorageDevice instance")
 
-        unavailable_dependencies = device.unavailable_dependencies
-        if unavailable_dependencies:
-            dependencies_str = ", ".join(str(d) for d in unavailable_dependencies)
-            raise DependencyError("device type %s requires unavailable_dependencies: %s" % (device.type, dependencies_str))
-
         self.device = device
+
+        self._check_device_dependencies()
+
         self.container = getattr(self.device, "container", None)
         self._applied = False
+
+    def _check_device_dependencies(self):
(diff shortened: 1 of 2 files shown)


tests/devices_test/dependencies_test.py
```diff
@@ -97,10 +97,6 @@ def test_availability_mdraidplugin(self):
             ActionCreateDevice(self.luks)
         with self.assertRaises(DependencyError):
             ActionDestroyDevice(self.dev)
-        with self.assertRaises(DependencyError):
-            ActionCreateFormat(self.dev)
-        with self.assertRaises(DependencyError):
-            ActionDestroyFormat(self.dev)
```
dwlehman: Why remove these if this test doesn't involve btrfs?

blivet/deviceaction.py
```diff
@@ -160,10 +160,19 @@ def __init__(self, device):
...
+            unavailable_dependencies = device.unavailable_dependencies
+            if unavailable_dependencies:
+                if device.type == "btrfs volume" and self.is_destroy:
+                    # XXX destroying a btrfs volume is a special case -- we don't destroy
+                    # the device, but use wipefs to destroy format on its parents so we
+                    # don't need btrfs plugin or btrfs-progs for this
+                    pass
+                else:
+                    dependencies_str = ", ".join(str(d) for d in unavailable_dependencies)
+                    raise DependencyError("device type %s requires unavailable_dependencies: %s" % (device.type, dependencies_str))
```
dwlehman: Maybe the dependency check should go into a private method and `ActionDestroyDevice` could run this new code and call the regular version of the checks via `super` for non-btrfs.

tests/devices_test/dependencies_test.py
vojtechtrefny: I've moved this change to a separate commit. This isn't really related to the btrfs changes, but I've noticed that we check the unavailable_dependencies for all actions but these dependencies are defined only for devices (formats use tasks and there are separate checks).
Currently it isn't possible to create (or remove) format on a device we don't support -- e.g. it isn't possible to format (opened) luks device without libblockdev crypto plugin. But the plugin isn't needed to run mkfs on a block device so we should allow it.

PR #723, opened 2018-09-17
Remove Anaconda flags

poncovka:
The flags `automated_install`, `live_install` and `image_install` can be
removed, because they are not used or not useful anymore.

Depends on: https://github.com/rhinstaller/anaconda/pull/1613

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -925,12 +925,7 @@ class Blivet(object):
         return tmp
 
     def _get_container_name_template(self, prefix=None):
-        template = prefix or ""
-
-        if flags.image_install:
-            template = "%s_image" % template
-
-        return template
+        return prefix or ""
 
     def suggest_container_name(self, prefix=""):
         """ Return a reasonable, unused device name.
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -426,7 +426,7 @@ class PopulatorMixin(object):
         self.drop_lvm_cache()
         mpath_members.drop_cache()
 
-        if flags.auto_dev_updates and not flags.image_install:
+        if flags.auto_dev_updates:
             blockdev.mpath.set_friendly_names(flags.multipath_friendly_names)
 
         self.setup_disk_images()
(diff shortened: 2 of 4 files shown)


blivet/blivet.py
```diff
@@ -925,12 +925,7 @@ def safe_device_name(self, name):
         return tmp
 
     def _get_container_name_template(self, prefix=None):
-        template = prefix or ""
-
-        if flags.image_install:
-            template = "%s_image" % template
-
-        return template
```
poncovka: Anaconda uses its own implementation of this method.

blivet/populator/populator.py
```diff
@@ -426,7 +426,7 @@ def _populate(self):
         self.drop_lvm_cache()
         mpath_members.drop_cache()
 
-        if flags.auto_dev_updates and not flags.image_install:
+        if flags.auto_dev_updates:
             blockdev.mpath.set_friendly_names(flags.multipath_friendly_names)
```
poncovka: This is the only place, where the flag `multipath_friendly_names` is used. So Anaconda might set `multipath_friendly_names` to False for the image installation.

poncovka:
I am not sure what to do with the `image_install` flag. Are the suggested changes ok or do you want to keep this flag?

blivet/populator/populator.py
dwlehman: I would expect that multipath is not at all relevant for image installations since they use the disk images as an exclusive set of available/visible disks, meaning you can probably do whatever you want about multipath for image installations and it won't matter at all.

tests/imagebackedtestcase.py
```diff
@@ -94,7 +94,7 @@ def set_up_storage(self):
 
     def setUp(self):
         """ Do any setup required prior to running a test. """
-        flags.image_install = True
+        flags.multipath_friendly_names = False
```
dwlehman: I think you can just remove the `image_install` lines without putting anything in their place.

poncovka:
Updated.

poncovka:
Hi, can I ask for a new review, please?

dwlehman:
Jenkins, ok to test.

PR #734, opened 2018-11-08
Use raid names when looking for slave devices

vojtechtrefny:
This fixes DeviceTreeError when adding LUKS device on a partition
on a raid -- the partition is added as "name1" so we can't use
the name from udev ("md127p1") here.

### DIFF
--- a/blivet/populator/helpers/partition.py
+++ b/blivet/populator/helpers/partition.py
@@ -23,10 +23,6 @@
 import copy
 import six
 
-import gi
-gi.require_version("BlockDev", "2.0")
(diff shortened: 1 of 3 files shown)


dwlehman:
I think the best way to go about this is to fix up `udev.device_get_name` to handle the md partition case better. I guess it could either call `name_from_node` or look through the udev symlinks for one that starts with '/dev/md/'.

vojtechtrefny:
> I think the best way to go about this is to fix up udev.device_get_name to handle the md partition case better.

Ok, changed.

vojtechtrefny:
Jenkins, test this please.

PR #738, opened 2018-11-21
Fix crash on reset on systems without nvdimm plugin

vojtechtrefny:
There is no easier way how to check if disk is an nvdimm
namespace so without nvdimm plugin udev.device_is_nvdimm_namespace
will return False and the device will be added as a normal disk.


vojtechtrefny:
Jenkins, test this please.

PR #747, opened 2018-12-17
Fixes for Offloaded FC

dwlehman:


blivet/populator/helpers/disklabel.py
```diff
@@ -44,7 +44,7 @@ def match(cls, data, device):
         return (bool(udev.device_get_disklabel_type(data)) and
                 not udev.device_is_biosraid_member(data) and
                 udev.device_get_format(data) != "iso9660" and
-                not (device.is_disk and mpath_members.is_mpath_member(device.path)))
+                not (device.is_disk and udev.device_get_format(data) == "mpath_member"))
```
vojtechtrefny: This is the only place where `mpath_members` is used in this file, so please remove the import too.

We probably should consider using the udev info everywhere, if it's better/more reliable than cached data from libblockdev. But this quite a big change and probably needs some further investigation, so I'm ok with changing it only here to fix this issue.

dwlehman: Removed in the latest revision.

PR #750, opened 2019-01-15
Do not crash if DM RAID activation fails (#1661712)

vojtechtrefny:
We don't want to crash during reset on systems with incomplete or
unsupported firmware RAIDs.

### DIFF
--- a/blivet/populator/helpers/dmraid.py
+++ b/blivet/populator/helpers/dmraid.py
@@ -66,7 +66,12 @@ class DMRaidFormatPopulator(FormatPopulator):
                 dm_array.parents.append(self.device)
             else:
                 # Activate the Raid set.
-                blockdev.dm.activate_raid_set(rs_name)
+                try:
+                    blockdev.dm.activate_raid_set(rs_name)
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
> Looks good to me. But I have no idea what's going to happen with the devices from the RAID array. Are they going to appear as disks? Or will they not be in the DeviceTree at all?

The disks should be already in the devicetree in this stage, just as normal disks with the *DMRaidMember* format.

dwlehman:
Since the format is dmraidmember they should be hidden and therefore it should be impossible to wipe the formatting via anaconda, which I think is best. Does it sound like what will happen, and do you agree it's best not to remove fwraid formatting from anaconda?

vojtechtrefny:
I think Anaconda needs some easy way to just wipe entire disk with degraded RAID, or some "old" firmware RAID etc. on the Storage Spoke, but that is a different problem. I think everything blivet should do is to add the disk to the devicetree with dmraidmember format and not crash.

dwlehman:
Sounds good to me.

PR #753, opened 2019-02-07
Mark exported VGs as protected

vojtechtrefny:
This needs some libblockdev changes -- https://github.com/storaged-project/libblockdev/pull/420 -- and it make take some time before this is merged and available, because it as a backwards incompatible ABI change, so we might need to wait to libblockdev 3.0. But this issue is not so common, so it shouldn't be a problem.

### DIFF
--- a/blivet/populator/helpers/lvm.py
+++ b/blivet/populator/helpers/lvm.py
@@ -36,7 +36,7 @@ from ...storage_log import log_method_call
 from .devicepopulator import DevicePopulator
 from .formatpopulator import FormatPopulator
 
-from ...static_data import lvs_info, pvs_info
+from ...static_data import lvs_info, pvs_info, vgs_info
 
 import logging
 log = logging.getLogger("blivet")
@@ -381,6 +381,17 @@ class LVMFormatPopulator(FormatPopulator):
                 log.warning("invalid data for %s: %s", self.device.name, e)
                 return
 
+            vg_info = vgs_info.cache.get(pv_info.vg_uuid)
+            if vg_info is None:
+                log.warning("Failed to get information about LVM volume group %s (%s)", vg_name, pv_info.vg_uuid)
+                return
+
+            if not hasattr(vg_info, "exported"):
+                log.info("Can't get exported information from VG info, assuming VG is not exported.")
+                exported = False
+            else:
+                exported = vg_info.exported
+
             vg_device = LVMVolumeGroupDevice(vg_name,
                                              parents=[self.device],
                                              uuid=vg_uuid,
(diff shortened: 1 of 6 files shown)


blivet/populator/helpers/lvm.py
```diff
@@ -381,6 +381,11 @@ def _add_vg_device(self):
                 log.warning("invalid data for %s: %s", self.device.name, e)
                 return
 
+            vg_info = vgs_info.cache.get(vg_name, None)
```
vpodzime: `get()` returns `None` by default.

blivet/populator/helpers/lvm.py
```diff
@@ -390,7 +395,8 @@ def _add_vg_device(self):
                                              pe_count=pe_count,
                                              pe_free=pe_free,
                                              pv_count=pv_count,
-                                             exists=True)
+                                             exists=True,
+                                             exported=vg_info.exported)
```
vpodzime: Would it make sense to do something like `getattr(vg_info, "exported", False)` so that this can be merged right away?

vojtechtrefny: I was actually thinking about doing that, but it might be confusing to have the property set to false for exported vgs just because we have old version of libblockdev.

vpodzime: Then it can be set to `None` for now which evaluates to `False` as a boolean anyway. But I see your point.

blivet/populator/helpers/lvm.py
vojtechtrefny: Fixed.

blivet/populator/helpers/lvm.py
vojtechtrefny: I've set the default value to `False` added a log message that should help with debugging in case there really is an exported vg.
Thoughts @dwlehman ?

vojtechtrefny:
> I'm a bit nervous about how this new vgs cache will behave when we have duplicate vgs, but otherwise I like it.

The `DuplicateVGError` is actually raised few lines above this change, so it shouldn't be a problem. Just to be 100 % sure, we have the right information in the cache, I've changed it to use `vg_uuid` as the key, so we should alway get information about the right VG, even with duplicate names.

blivet/static_data/lvm_info.py
```diff
@@ -115,7 +115,7 @@ def cache(self):
                 self._vgs_cache = dict()
                 return self._vgs_cache
 
-            self._vgs_cache = dict(("%s" % (vg.name), vg) for vg in vgs)
+            self._vgs_cache = dict(("%s" % (vg.uuid), vg) for vg in vgs)
```
vpodzime: This should go to the first commit.

vojtechtrefny: Good catch, fixed.

PR #756, opened 2019-02-14
Do not try to get format free space for non-existing formats

vojtechtrefny:
Resolves: rhbz#1676480

`lvmpv.free` raises exception for non-existing PVs. I'm not sure why, it is possible that simply returning "0" would be better way to solve this bug, but it is possible there is some reason behind this behaviour.
Also `fs_free` in `get_free_space` is defined as *space available by shrinking filesystems* so it probably doesn't make sense to include non-existing filesystems.


dwlehman:
I think there must have been some other change that caused this to suddenly be an issue.

PR #759, opened 2019-02-25
Remove unused API code

jkonecny12:
You have as part of your API `UnknownSourceDeviceError` exception but this error is not used anywhere in your code and Anaconda removed the import.

I know you can't just remove this now because it is your API but you should do that on the next opportunity to remove unused code and make your API more lightweight.

Anaconda removing this in PR: https://github.com/rhinstaller/anaconda/pull/1834/

### DIFF
--- a/blivet/errors.py
+++ b/blivet/errors.py
@@ -265,12 +265,6 @@ class UnrecognizedFSTabEntryError(StorageError):
 class FSTabTypeMismatchError(StorageError):
     pass
 
-# probing
-
-
-class UnknownSourceDeviceError(StorageError):
-    pass
-
 # factories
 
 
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
@dwlehman Can we merge this for 3.2? We already have some behaviour changes and changes in API but only small things like new exception being raised, not removed API.

dwlehman:
> @dwlehman Can we merge this for 3.2?

Yeah, it's an anaconda-only exception anyway.

jkonecny12:
Rebased on `3.2-devel`.

vojtechtrefny:
> Rebased on `3.2-devel`.

Thank you.

PR #762, opened 2019-02-27
Automatically adjust size of growable devices for new format

vojtechtrefny:
Without this kickstart 'part /home --size=1 --grow --encrypted'
will fail because min size for LUKS is 2 MiB.

Resolves: rhbz#1680013

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -721,7 +721,12 @@ class StorageDevice(Device):
             if fmt.max_size and fmt.max_size < self.size:
                 raise errors.DeviceError("device is too large for new format")
             elif fmt.min_size and fmt.min_size > self.size:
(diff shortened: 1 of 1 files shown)


blivet/devices/storage.py
```diff
@@ -721,7 +721,10 @@ def _set_format(self, fmt):
             if fmt.max_size and fmt.max_size < self.size:
                 raise errors.DeviceError("device is too large for new format")
             elif fmt.min_size and fmt.min_size > self.size:
-                raise errors.DeviceError("device is too small for new format")
+                if self.growable:
+                    self.size = fmt.min_size
```
vpodzime: I'd log an info message here.

vojtechtrefny: Fixed.

PR #763, opened 2019-03-07
Support legacy MBR (msdos) as part of UEFI to enable hybrid builds

nullr0ute:
The UEFI spec supports legacy MBR (msdos) in section 5.2 of the UEFI 2.7 spec [1].
There's a number of use cases where we might boot on a UEFI device but want to
support, but not by default, the creation of images using legacy MBR partitions.
One of these use cases is cloud images where we can produce a single image
tha will run on UEFI supported, including secure-boot, clouds but also
support AWS which only supports MBR partitioning schemes [2] and in the current
config anaconda fails the install where this is a perfectly valid and widely
supported deployment mechanism.

So enable the ability to use legacy msdos partitioning schemes, but don't use
it for default installs, so people may consume this mechanism using kickstarts
rather than having the install fail for a valid use case.

[1] https://uefi.org/sites/default/files/resources/UEFI%20Spec%202_7_A%20Sept%206.pdf
[2] https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html


vojtechtrefny:
Can you please create a new PR against `3.1-devel`, we have compilcated system of branches and the `release` ones are only for doing releases. I've tried to change the branch using GH, but it doesn't wort.

nullr0ute:
It would be useful if you'd actually document that somewhere, it would save everyone a bunch of time :)

nullr0ute:
https://github.com/storaged-project/blivet/pull/764

vojtechtrefny:
We have some [documentation for contributors](https://github.com/storaged-project/blivet/blob/3.1-devel/CONTRIBUTING) but it's a little bit outdated. Thank you for creating the new PR:

PR #764, opened 2019-03-08
Support legacy MBR (msdos) as part of UEFI to enable hybrid builds

nullr0ute:
The UEFI spec supports legacy MBR (msdos) in section 5.2 of the UEFI 2.7 spec [1].
There's a number of use cases where we might boot on a UEFI device but want to
support, but not by default, the creation of images using legacy MBR partitions.
One of these use cases is cloud images where we can produce a single image
tha will run on UEFI supported, including secure-boot, clouds but also
support AWS which only supports MBR partitioning schemes [2] and in the current
config anaconda fails the install where this is a perfectly valid and widely
supported deployment mechanism.

So enable the ability to use legacy msdos partitioning schemes, but don't use
it for default installs, so people may consume this mechanism using kickstarts
rather than having the install fail for a valid use case.

[1] https://uefi.org/sites/default/files/resources/UEFI%20Spec%202_7_A%20Sept%206.pdf
[2] https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -228,7 +228,7 @@ class DiskLabel(DeviceFormat):
         elif arch.is_efi() and arch.is_arm():
             label_types = ["msdos", "gpt"]
         elif arch.is_efi() and not arch.is_aarch64():
-            label_types = ["gpt"]
+            label_types = ["gpt", "msdos"]
         elif arch.is_s390():
             label_types = ["msdos", "dasd"]
 
--- a/tests/formats_test/disklabel_test.py
+++ b/tests/formats_test/disklabel_test.py
@@ -81,7 +81,7 @@ class DiskLabelTestCase(unittest.TestCase):
         arch.is_pmac.return_value = False
 
         arch.is_efi.return_value = True
-        self.assertEqual(disklabel_class.get_platform_label_types(), ["gpt"])
+        self.assertEqual(disklabel_class.get_platform_label_types(), ["gpt", "msdos"])
         arch.is_aarch64.return_value = True
         self.assertEqual(disklabel_class.get_platform_label_types(), ["gpt", "msdos"])
         arch.is_aarch64.return_value = False


nullr0ute:
Note we need this in at least F-29 -> rawhide builds (not sure about F-28) as it's affecting cloud builds in some situations

vojtechtrefny:
Jenkins, ok to test.

dustymabe:
> not sure about F-28

we don't need it for f28

blivet/formats/disklabel.py
```diff
@@ -228,7 +228,7 @@ def get_platform_label_types(cls):
         elif arch.is_efi() and arch.is_arm():
             label_types = ["msdos", "gpt"]
         elif arch.is_efi() and not arch.is_aarch64():
-            label_types = ["gpt"]
+            label_types = ["gpt", "msdos"]
```
vpodzime: The two `elif` branches for `is_efi()` could be merged into a single one now.

nullr0ute: The first one in the list is the default, in teh case of is_arm we need msdos as the default but x86 we obviously want gpt as the default. I had left the checks as they were due to arch defaults (this has been broken by others in the past).

vpodzime: Makes sense, thanks for the explanation!

AdamWill:
Has anyone tested the implications of this in other scenarios?

For instance, before this change I believe if you try a UEFI install to an MBR-labelled disk, even if you wipe all the partitions or otherwise provide enough space, anaconda will not let you do the install, it will require you to relabel it to GPT. I'm pretty sure this behaviour was *intentional*: we do not want regular people doing UEFI installs to MBR-labelled disks. Does it now accept this and go ahead with the install? If so, that might be a problem.

Has @vathpela been asked about this?

nullr0ute:
@AdamWill this has worked just fine on aarch64 with UEFI since F-25

AdamWill:
As I recall, we were worried about UEFI firmwares which won't cope with booting from MBR-labelled disks. In theory they ought to, and probably many do, but this is much like when we tried doing BIOS installs to GPT-labelled disks for one release - it turned out to be a disaster as we found out that several firmwares just don't handle this and so the system wouldn't boot.

That's why we wound up being quite strict about BIOS installs *only* using MBR-labelled disks (with the sole exception of >2TB disks) and UEFI installs *only* using GPT-labelled disks. Again, AIUI.

nullr0ute:
This doesn't change the defaults at all. It makes things an option. This is still UEFI complaint, see the commit message.

AdamWill:
I know it doesn't change the defaults, I'm familiar with the code. I know it doesn't violate the spec, I'm familiar with the spec. I gave an example of a scenario I'm concerned about, it does not involve either the defaults or the UEFI spec.

vathpela:
> I'm pretty sure this behaviour was _intentional_: we do not want regular people doing UEFI installs to MBR-labelled disks.

I'm sure the other way around was intentional—no BIOS on a GPT+PMBR—but I'm not really sure this way was.  I think it should work fine. Since this doesn't affect the installer defaults, I'm okay with it.

AdamWill:
If @vathpela is OK with it then I guess I'm fine to try it at least...

AdamWill:
Just for the record, after a substantial session of Git Archaeology, you can trace the origin of this decision to [this anaconda commit from 2008 by @vathpela](https://github.com/rhinstaller/anaconda/commit/00992d8cfb#diff-5b6842b333a3a8098f73d80417f72824R312) . That added a function used to decide if a partition was a valid EFI system partition, with part of the requirement being that it be on a GPT-labelled disk. Before that commit, this requirement doesn't seem to be expressed anywhere in anaconda (I'm not sure what the practical state of EFI install support was before, or immediately after, this commit).

There are several minor changes after that, then [this significant commit](https://github.com/rhinstaller/anaconda/commit/de1469011a56202d1b2fa73dca40e07da5212c34) which substantially changes the implementation, but effectively keeps the GPT requirement - it drops the whole `isEfiSystemPartition()` concept, but introduces a class attribute `target_device_disklabel_types` for bootloader classes, set to `["msdos", "gpt"]` for GRUB and `["gpt"]` for EFIGRUB. This is the basic design we still have now. The attribute got renamed, moved from the bootloader classes to the platform classes and renamed again, then moved to blivet (among other trivial changes), but the basic requirement 'EFI system partition must be on GPT-labelled disk' was kept the entire time.

The original commit from back in 2008 that introduced this requirement did not explicitly explain it, though, either in a comment or in the commit message. And it's significant that the requirement was introduced as part of what looks like basic EFI support engineering, not as a restriction specifically added in response to some sort of issue or anything. So that all supports the idea that it wasn't necessarily specifically thought through in terms of "we explicitly want to exclude MBR from this".

Anyway, figured I'd poke through it and write that down just in case we ever have to revisit this...

dustymabe:
thanks for the investigation and summary @AdamWill - I always enjoy your writeups :)

PR #769, opened 2019-04-01
Add tests to the source tarball

vojtechtrefny:


vojtechtrefny:
> Isn't this going to install tests to people installing from pip?

I'm not sure about pip, but using `setup.py install` won't install the tests, because it installs only "packages" included in `setup.py`.

vojtechtrefny:
Adding the tests using `recursive-include` breaks pip installation -- pip tries to go through all directories and ends in an infinite loop in the `edd_data` directory (there are some sysfs-like directory trees with symlinks). Adding the tests directory using `graft` (include all files under specified directory) and excluding some files we don't want to (.pyc files and pylint test results) seems to be working with pip.

vojtechtrefny:
#772 is a better solution for this.

PR #771, opened 2019-04-04
Fix mounting of the filesystem iso9660

poncovka:
It is not possible to mount the filesystem iso9660, if the kernel
module isofs (with alias iso9660) is not already loaded. Blivet will
fail with a message "mounting filesystem iso9660 is not supported",
because the mount type is not in kernel_filesystems, there is no file
/sbin/mount.iso9660 and there is no module iso9660.ko*. Therefore,
let's make sure that the kernel module is loaded.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -1242,6 +1242,7 @@ class Iso9660FS(FS):
 
     """ ISO9660 filesystem. """
     _type = "iso9660"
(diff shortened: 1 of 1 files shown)


poncovka:
I have tested the suggested solution with Anaconda and it seems to work. Another solution is to change the method `_has_driver` and call for example `modinfo` that is able to find a module by its alias.

PR #772, opened 2019-04-22
Tests archive

dwlehman:
This provides a make target to create an archive of the tests/ subdirectory and spec file logic to add it to the unpacked sources during rpm build. The remaining piece (that I know of) is to change all of the test dependencies to from `Requires` to `BuildRequires`, which I'm not super excited about.

Thoughts?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,10 @@ archive: po-pull
 	git checkout -- po/$(PKGNAME).pot
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
+tests-archive:
+	git archive --format=tar --prefix=$(PKGNAME)-$(VERSION)/ $(VERSION_TAG) tests/ | gzip -9 > $(PKGNAME)-$(VERSION)-tests.tar.gz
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -120,6 +120,10 @@ archive: po-pull
 	git checkout -- po/$(PKGNAME).pot
 	@echo "The archive is in $(PKGNAME)-$(VERSION).tar.gz"
 
+tests-archive:
+       git archive --format=tar --prefix=$(PKGNAME)-$(VERSION)/ $(VERSION_TAG) tests/ | gzip -9 > $(PKGNAME)-$(VERSION)-tests.tar.gz
```
vojtechtrefny: Spaces instead of tab, `make tests-archive` fails here with `Makefile:124: *** missing separator.  Stop.`

dwlehman: Sorry -- cut&paste error. Fixed in latest version (runs for me on local workstation).

vojtechtrefny:
This should be also dependency of the `release`, `srpm` and `rpm` targets to create the tests tarball automatically. The RPM builds tests are failing because `make srpm` uses `local` target which doesn't create the tests tarball.

PR #773, opened 2019-05-02
Only call mpath plugin when it is available. (#1697378)

dwlehman:
@atodorov let me know if you need assistance in making an updates image or similar for testing this.


atodorov:
@dwlehman , @jikortus is looking into this

jikortus:
@atodorov @dwlehman I tried to manually apply changes from this commit against the installed python3-blivet-3.1.2-4.fc29.noarch and it did the job (no traceback upon building composes that occurred otherwise), so I think the fix is fine.

PR #774, opened 2019-05-07
Use only sysfs for iscsi device attributes

rvykydal:
The changes are required by https://github.com/rhinstaller/anaconda/pull/1960

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -790,6 +790,16 @@ def device_get_iscsi_name(info):
     return "-".join(path_components[name_field:len(path_components) - 2])
 
 
+def device_get_iscsi_lun(info):
+    lun = 0
+    path_components = device_get_path(info).split("-")
+    idx = list(reversed(path_components)).index("lun")
+    if idx > 0:
+        lun = path_components[-idx]
+
+    return lun
+
+
 def device_get_iscsi_address(info):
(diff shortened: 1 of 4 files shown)


blivet/udev.py
```diff
@@ -790,6 +790,16 @@ def device_get_iscsi_name(info):
     return "-".join(path_components[name_field:len(path_components) - 2])
 
 
+def device_get_iscsi_lun(info):
+    lun = 0
+    path_components = device_get_path(info).split("-")
+    idx = list(reversed(path_components)).index("lun")
+    if idx > 0:
+        lun = path_components[-idx]
```
vpodzime: Would be nice to have an example string (path) for this magic.

rvykydal: The example is a few functions above (common for all the iscsi related functions).

rvykydal:
Updated the unit tests with the new attributes.

rvykydal:
@vpodzime Yes, I'd like to look at what could be moved from blivet/iscsi.py and the credentials are a good candidate. They would be stored in anaconda iscsi module (if we really need them, I have to check).

rvykydal:
It would be great to have this merged and built rather sooner as we have a dependent PR we'd like to merge soon in anaconda.

PR #776, opened 2019-05-13
Recommend DASD disk label for disks with an existing DASD label

vojtechtrefny:
It is possible to have a virtio disk backed by a DASD disk and
in this case we use disk type to determine best label type
because disk type is not DASD.

Resolves: rhbz#1677383

Submitted-by: Masahiro Matsuya <mmatsuya@redhat.com>


dwlehman:
Jenkins, test this please.

PR #777, opened 2019-05-22
Add flag for protecting cdrom devices during populate

rvykydal:
Related: rhbz#1671713

### DIFF
--- a/blivet/flags.py
+++ b/blivet/flags.py
@@ -77,6 +77,9 @@ class Flags(object):
         # (so far only for LUKS)
         self.discard_new = False
 
+        # whether cdroms should be protected
+        self.protect_cdroms = False
+
         self.boot_cmdline = {}
 
         self.update_from_boot_cmdline()
--- a/blivet/devices/optical.py
+++ b/blivet/devices/optical.py
@@ -24,6 +24,7 @@ import os
 from .. import errors
 from .. import util
 from ..storage_log import log_method_call
+from ..flags import flags
 
 import logging
 log = logging.getLogger("blivet")
@@ -81,3 +82,16 @@ class OpticalDevice(StorageDevice):
             util.run_program(["eject", self.name])
         except OSError as e:
             log.warning("error ejecting cdrom %s: %s", self.name, e)
+
+    @property
+    def protected(self):
+        protected = super(OpticalDevice, self).protected
+
+        if flags.protect_cdroms:
+            return True
+        else:
+            return protected
+
+    @protected.setter
+    def protected(self, value):
+        self._protected = value


rvykydal:
Please see the discussion in the related https://github.com/rhinstaller/anaconda/pull/1977

blivet/flags.py
```diff
@@ -77,6 +77,9 @@ def __init__(self):
         # (so far only for LUKS)
         self.discard_new = False
 
+        # wheteher cdroms should be made protected during devicetree population
```
vpodzime: *whether*

blivet/flags.py
```diff
@@ -77,6 +77,9 @@ def __init__(self):
...
+        self.protect_cdroms = True
```
vpodzime: Any reason why this should be configurable? Let's not introduce flags for absurd/artificial cases.

blivet/devicetree.py
```diff
@@ -936,6 +936,12 @@ def _hide_ignored_disks(self):
                 if ignored:
                     self.hide(disk)
 
+    def _protect_cdroms(self):
+        # protect cdrom devices
+        for device in self._devices:
+            if device.type == "cdrom":
+                log.debug("protecting cdrom %s", device.name)
+                device.protected = True
```
vpodzime: Shouldn't this be done when the CD-ROMs are being "populated-in"?

blivet/flags.py
dwlehman: This should default to False.

rvykydal: The reason is that to be safe - I wanted to apply the change only to image installations (or better - installations on running system in the future).
This approach is definitely quite ad-hoc solution.
See the discussion in related anaconda PR: https://github.com/rhinstaller/anaconda/pull/1977
Maybe we could set OpticalDevices as protected by default?
I was also thinking the devices could be protected as being read-only (if _mark_readonly_device was not limited to disks but was applied also for cdroms) but apparently cdrom device is not read-only.

blivet/devicetree.py
rvykydal: In the optical device populator run() ? Would it be OK to use the flag then?

blivet/flags.py
rvykydal: Ah, sure.

rvykydal:
After talking to Vojta I moved application of the flag to OpticalDevice "protected" property.
As for a solution without need for the special flag, I opened https://github.com/storaged-project/blivet/issues/779.

blivet/flags.py
rvykydal: @dwlehman I fixed the default value and changed the approach a bit (see below)

rvykydal:
Just updated rhbz reference to BZ clone for blivet componet.

vojtechtrefny:
@rvykydal Can we merge this or are we waiting for some changes in Anaconda?

dwlehman:
I still prefer another approach as indicated in discussion of #779. The `teardown_all` at the end of populate only runs when blivet is being driven by anaconda. If it turns out that removing that resolves the issue I would vastly prefer that approach to adding new logic and a flag for such a specific circumstance.

poncovka:
This change can be handled in Anaconda now. See https://github.com/storaged-project/blivet/issues/779#issuecomment-572522589.

rvykydal:
I think we can close this PR - it was added to python-blivet for rhel as downstream patch. I've created jira issue for porting of the BZ to Fedora, building on https://github.com/storaged-project/blivet/pull/780 which has been merged to 3.1 devel.

PR #784, opened 2019-06-26
Check dmraid status before trying to activate array during populate

dwlehman:
https://bugzilla.redhat.com/show_bug.cgi?id=1723979

### DIFF
--- a/blivet/populator/helpers/dmraid.py
+++ b/blivet/populator/helpers/dmraid.py
@@ -70,12 +70,13 @@ class DMRaidFormatPopulator(FormatPopulator):
                 # We add the new device.
                 dm_array.parents.append(self.device)
             else:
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
I could change libblockdev to do the check before trying to activate the array instead of returning error. The question is what is expected behaviour here. Similar functions sometimes return error, sometimes don't, it usually depends on the tool we are using -- `md_activate` fails, but `lvm_lvactivate` doesn't (because *lvchange -ay* on an active lv works).

dwlehman:
My sense is that `activate` and `deactivate` functions would ideally be idempotent.

vpodzime:
> My sense is that `activate` and `deactivate` functions would ideally be idempotent.

:+1:

PR #785, opened 2019-07-08
RFC: Avoid some tracebacks during populate

dwlehman:
Having fixed big issues with active LVM on partial VGs and with unusable dmraid arrays, the most-reported failure during blivet populate involves partitioned disks that have been cloned by users without adjusting the disklabel UUIDs to make them unique. In the past I have generally favored a "you must fix this problem before proceeding" approach, but now I'm thinking more about being robust and not crashing unless it's really necessary.

I'm mostly curious what people think of "Allow duplicate UUIDs...", "Unset partition UUIDs...", and "General protection against...". The first two of these are tested, both via unit tests and a "real" test using examples/list_devices.py in a vm. The third ("General protection...") probably needs more work.

### DIFF
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -40,6 +40,7 @@ from .devices.lib import Tags
 from . import formats
 from .devicelibs import lvm
 from .events.handler import EventHandlerMixin
+from .flags import flags
 from . import util
 from .populator import PopulatorMixin
 from .storage_log import log_method_call, log_method_return
@@ -137,8 +138,9 @@ class DeviceTreeBase(object):
                 continue
 
             if device.uuid and device.uuid in [d.uuid for d in devices] and \
+               not flags.allow_inconsistent_config and \
                not isinstance(device, NoDevice):
-                raise DeviceTreeError("duplicate uuids in device tree")
+                raise DuplicateUUIDError("duplicate uuids in device tree")
 
             devices.append(device)
 
@@ -159,7 +161,7 @@ class DeviceTreeBase(object):
             dev = self.get_device_by_uuid(newdev.uuid, incomplete=True, hidden=True)
             if dev.name == newdev.name:
                 raise DeviceTreeError("Trying to add already existing device.")
-            else:
+            elif not flags.allow_inconsistent_config:
                 raise DuplicateUUIDError("Duplicate UUID '%s' found for devices: "
                                          "'%s' and '%s'." % (newdev.uuid, newdev.name, dev.name))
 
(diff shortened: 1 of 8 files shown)


blivet/devicetree.py
```diff
@@ -160,6 +160,12 @@ def _add_device(self, newdev, new=True):
             dev = self.get_device_by_uuid(newdev.uuid, incomplete=True, hidden=True)
             if dev.name == newdev.name:
                 raise DeviceTreeError("Trying to add already existing device.")
+            elif newdev.type == "partition" and dev.type == "partition" and not flags.uevents:
+                # This is a duplicate partition UUID resulting from a cloned disk. Just
+                # pretend it has no UUID. The only negative effect is on event handling,
+                # which is disabled by default.
+                newdev.uuid = None
+                dev.uuid = None
```
vojtechtrefny: I don't like the idea of "removing" the UUIDs. We already have two ways how to make the reset "stable" in this situation -- special exception that can be caught and used to ignore one of the disks and the flag. And now we are adding a special case for partitions that just "silently" removes the UUID?. I think the flag should be enough and Anaconda should set it by default.

dwlehman: I agree that seems like a sketchy thing to do. I dropped that commit.

blivet/devicetree.py
```diff
@@ -510,7 +511,14 @@ def get_device_by_uuid(self, uuid, incomplete=False, hidden=False):
         result = None
         if uuid:
             devices = self._filter_devices(incomplete=incomplete, hidden=hidden)
-            result = six.next((d for d in devices if d.uuid == uuid or d.format.uuid == uuid), None)
+            matches = iter(d for d in devices if d.uuid == uuid or d.format.uuid == uuid)
+            result = six.next(matches, None)
+            extra = six.next(matches, None)
```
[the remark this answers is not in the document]
dwlehman: Agreed. Done.

blivet/populator/populator.py
```diff
@@ -262,7 +263,20 @@ def handle_device(self, info, update_orig_fmt=False):
             helper_class = self._get_device_helper(info)
 
         if helper_class is not None:
-            device = helper_class(self, info).run()
+            try:
+                device = helper_class(self, info).run()
+            except DeviceTreeError as e:
```
vpodzime: What could be really important here is that every `run()` method should make sure to gather as much information as possible before throwing the exception. Otherwise the information about the problematic device may be very incomplete.

vojtechtrefny:
Obsoleted by #1306

PR #786, opened 2019-07-15
Fix removing LUKS devices with integrity

vojtechtrefny:
We need to propagate the teardown call down to the underlying
device to call teardown on its LUKS format. Teardown for the
Integrity device is a no-op (same as for the LUKS device itself).


vojtechtrefny:
Added two more fixes. I'm not sure if the `slave` change is safe, from the code it looks like `slave` should be always the underlying block device (e.g. partition) which means one extra layer in case of LUKS+integrity and without this change `dracut_setup_args` provides invalid boot arguments.

PR #788, opened 2019-07-17
Check if disklabel supports partition names (#1723228)

vojtechtrefny:
pyparted now raises an exception when trying to access partition
name on a disklabel that doesn't support partition names, see
https://github.com/dcantrell/pyparted/issues/59

### DIFF
--- a/blivet/populator/helpers/boot.py
+++ b/blivet/populator/helpers/boot.py
@@ -59,6 +59,7 @@ class MacEFIFormatPopulator(BootFormatPopulator):
         fmt = formats.get_format(cls._type_specifier)
         try:
             return (super(MacEFIFormatPopulator, MacEFIFormatPopulator).match(data, device) and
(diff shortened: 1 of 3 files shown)


blivet/populator/helpers/boot.py
```diff
@@ -59,6 +59,7 @@ def match(cls, data, device):
         fmt = formats.get_format(cls._type_specifier)
         try:
             return (super(MacEFIFormatPopulator, MacEFIFormatPopulator).match(data, device) and
+                    device.disk.format.supports_names and
```
dwlehman: Is it possible this could be called with a non-partition device? If so I think it will crash. You could also use `device.disks[0]` which would work for any device type.

vojtechtrefny: `BootFormatPopulator.match` checks if the device is a partition so it should be ok.

PR #792, opened 2019-07-30
Always use luks_data.min_entropy as a default minimum entropy

poncovka:
Use `luks_data.min_entropy` if the minimum entropy is not set.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -332,7 +332,11 @@ class DeviceFactory(object):
             setattr(self, setting, value)
 
         self.fstype = None  # not included in default_settings b/c of special handling below
-        self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
+        self.min_luks_entropy = kwargs.get("min_luks_entropy")
+
+        if self.min_luks_entropy is None:
+            self.min_luks_entropy = luks_data.min_entropy
+
         self.luks_version = kwargs.get("luks_version") or crypto.DEFAULT_LUKS_VERSION
         self.pbkdf_args = kwargs.get("pbkdf_args", None)
 
--- a/blivet/formats/luks.py
+++ b/blivet/formats/luks.py
@@ -130,7 +130,10 @@ class LUKS(DeviceFormat):
         self._key_file = kwargs.get("key_file")
         self.escrow_cert = kwargs.get("escrow_cert")
         self.add_backup_passphrase = kwargs.get("add_backup_passphrase", False)
-        self.min_luks_entropy = kwargs.get("min_luks_entropy", 0)
+        self.min_luks_entropy = kwargs.get("min_luks_entropy")
+
+        if self.min_luks_entropy is None:
+            self.min_luks_entropy = luks_data.min_entropy
 
         if not self.map_name and self.exists and self.uuid:
             self.map_name = "luks-%s" % self.uuid
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -611,13 +611,10 @@ class ActionCreateFormat(DeviceAction):
             udev.settle()
 
         if isinstance(self.device.format, luks.LUKS):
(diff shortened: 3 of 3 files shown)


blivet/devicefactory.py
```diff
@@ -332,7 +332,7 @@ def __init__(self, storage, **kwargs):
             setattr(self, setting, value)
 
         self.fstype = None  # not included in default_settings b/c of special handling below
-        self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
+        self.min_luks_entropy = kwargs.get("min_luks_entropy", luks_data.min_entropy)
```
dwlehman: The old code here is actively protecting against using a passed-in `0` or `None`. Is that not desirable?

blivet/formats/luks.py
```diff
@@ -130,7 +130,7 @@ def __init__(self, **kwargs):
         self._key_file = kwargs.get("key_file")
         self.escrow_cert = kwargs.get("escrow_cert")
         self.add_backup_passphrase = kwargs.get("add_backup_passphrase", False)
-        self.min_luks_entropy = kwargs.get("min_luks_entropy", 0)
+        self.min_luks_entropy = kwargs.get("min_luks_entropy", luks_data.min_entropy)
```
dwlehman: If this is a real minimum maybe we should be treating as such by doing something like `min(kwargs.get('min_luks_entropy') or 0, luks_data.min_entropy)`.

dwlehman: Oops, had that backwards.

blivet/devicefactory.py
vpodzime: Yeah, `None` should be replaced  by a number, I think. `0` should, however, be respected.

blivet/formats/luks.py
vpodzime: It's not really an absolutely required minimum, it's more a recommended minimum.

poncovka:
Updated.

This pull request shouldn't change the current behaviour of Blivet or Anaconda. Blivet uses by default zero minimum entropy, Anaconda uses by default `MIN_CREATE_ENTROPY`. This is not going to change. We just want to simplify the code in Anaconda (see https://github.com/rhinstaller/anaconda/pull/2061).

blivet/devicefactory.py
```diff
@@ -332,7 +332,11 @@ def __init__(self, storage, **kwargs):
             setattr(self, setting, value)
 
         self.fstype = None  # not included in default_settings b/c of special handling below
-        self.min_luks_entropy = kwargs.get("min_luks_entropy") or luks_data.min_entropy
+        self.min_luks_entropy = kwargs.get("min_luks_entropy", None)
```
vpodzime: `None` is returned by default by `.get()` if the key doesn't exist.

blivet/formats/luks.py
```diff
@@ -130,7 +130,10 @@ def __init__(self, **kwargs):
         self._key_file = kwargs.get("key_file")
         self.escrow_cert = kwargs.get("escrow_cert")
         self.add_backup_passphrase = kwargs.get("add_backup_passphrase", False)
-        self.min_luks_entropy = kwargs.get("min_luks_entropy", 0)
+        self.min_luks_entropy = kwargs.get("min_luks_entropy", None)
```
vpodzime: This is a change in behavior in Blivet, isn't it? It was defaulting to `0` and now it defaults to `luks_data.min_entropy`.

poncovka: But `luks_data.min_entropy` is set to `0` by default in Blivet.

blivet/devicefactory.py
poncovka: Fixed.

blivet/formats/luks.py
vpodzime: Oh! That explains a lot :smile:

poncovka:
@dwlehman, is there anything else I should change?

PR #794, opened 2019-07-31
More consistent lvm errors (API break)

### LINKED ISSUE #528: Raise more specific exceptions in devices/lvm.py
The `lvm.py` raising the `ValueError` exception on everything. IMO two things should be changed here.

1. Use other exception than basic `ValueError`, could be a subclass of the `ValueError` exception.
2. Use more different types of an exceptions not just one.

The second point is because now you have:
`ValueError("not enough free space in volume group")`
and
`ValueError("new size must of type Size")`

both are `ValueError` but one of those can be just a bad settings and second is clearly an programming error. It would be nice to have some lead what is what. Now you can't easily decide how to handle errors.

japokorn:
- changed exception types in devices/lvm.py to
be more consistent (as per https://github.com/storaged-project/blivet/issues/528)
- these modifications change API

Fixes #528

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -303,7 +303,7 @@ class LVMVolumeGroupDevice(ContainerDevice):
     def _add_log_vol(self, lv):
         """ Add an LV to this VG. """
         if lv in self._lvs:
-            raise ValueError("lv is already part of this vg")
+            raise errors.DeviceError("lv is already part of this vg")
 
         # verify we have the space, then add it
         # do not verify for growing vg (because of ks)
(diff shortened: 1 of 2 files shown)


japokorn:
fixes issue #528

vojtechtrefny:
I'm not sure if we want to do this in 3.2.0 or wait for a major version to change the API (do we plan to do Blivet 4.0 soon?). What do you think @dwlehman?

vojtechtrefny:
@jkonecny12 Can you take a look at this? Youˇreported the original issue. Do we need some changes in Anaconda to adjust to the new exceptions?

jkonecny12:
@poncovka has a much better knowledge of these now. I'll leave the review on her.

vojtechtrefny:
I've moved this PR from 3.2-devel to 3.3-devel and I think this can be merged now.

PR #797, opened 2019-08-08
Move dependencies code from StorageDevice to Device

vojtechtrefny:
This fixes a new pylint warning about Device not having the
'type_external_dependencies' method. The 'external_dependencies'
checks dependencies of all ancestors and theoretically it is
possible that a StorageDevice has Device in its ancestors.

----

We could simply ignore this pylint warning, but I think this time pylint is right, technically we can have a `StorageDevice` with `Device` within its ancestors. And all the `ancestors` and `parents` methods/properties are defined in the `Device` class so it makes sense to have the dependencies functions there too.

### DIFF
--- a/blivet/devices/device.py
+++ b/blivet/devices/device.py
@@ -71,6 +71,7 @@ class Device(util.ObjectID):
 
     _type = "device"
     _packages = []
+    _external_dependencies = []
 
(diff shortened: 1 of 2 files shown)


japokorn:
Looks good to me.

PR #801, opened 2019-08-23
Do not try to normalize size for zero size device factories

vojtechtrefny:
Factories with zero size are special cases for adjusting container
size after removing a device for it. We don't want to change size
of the factory in this case.
The recent change of filesystem minimal size to 2 MiB resulted
in changing of size of these factories from 0 to 2 MiB which
caused the "adjusting factory" to create a new LV after removing
one from the container.

Resolves: rhbz#1743753

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -462,6 +462,10 @@ class DeviceFactory(object):
     def _normalize_size(self):
         if self.size is None:
             self._handle_no_size()
+        elif self.size == Size(0):
+            # zero size means we're adjusting the container after removing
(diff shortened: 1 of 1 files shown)


blivet/devicefactory.py
```diff
@@ -463,6 +463,11 @@ def _normalize_size(self):
         if self.size is None:
             self._handle_no_size()
 
+        if self.size == Size(0):
```
vpodzime: I'd use `elif` here.

vojtechtrefny: Ok, fixed.

PR #802, opened 2019-09-11
Fix util.detect_virt function

vojtechtrefny:
Fixed the systemd Manager object path, also get_property_sync
returns a tuple so we need to check its first element.

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -40,7 +40,7 @@ program_log_lock = Lock()
 
 
 SYSTEMD_SERVICE = "org.freedesktop.systemd1"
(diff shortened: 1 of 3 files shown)


blivet/util.py
```diff
@@ -1117,4 +1117,4 @@ def detect_virt():
     except (safe_dbus.DBusCallError, safe_dbus.DBusPropertyError):
         vm = None
 
-    return vm in ('qemu', 'kvm')
+    return vm[0] in ('qemu', 'kvm')
```
dwlehman: If we catch an exception above `vm` will be None and this will cause a traceback.

vojtechtrefny: Fixed.

PR #803, opened 2019-09-23
Fix bug #1750123 blivet only try to create one pv on one disk,which means users can't take use of the smaller freespace region even they want to

skycastlelily2:


blivetghoul:

PR #805, opened 2019-10-04
Check for PV sector size when creating new VG

vojtechtrefny:
LVM no longer allows creating VGs on top of disks with different logical sector sizes so we need to check this before running vgcreate (which fails).

There is a pull request for anaconda -- https://github.com/rhinstaller/anaconda/pull/2142 -- which adds a new StorageChecker check but we need to check this also for other users and we also need to provide API for getting sector size for disks without a disklabel.

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -356,6 +356,18 @@ class LVMVolumeGroupDevice(ContainerDevice):
     def _add_parent(self, parent):
         super(LVMVolumeGroupDevice, self)._add_parent(parent)
 
+        # we are creating new VG or adding a new PV to an existing (complete) one
+        if not self.exists or (self.exists and self._complete):
+            parent_sectors = set([p.sector_size for p in self.pvs] + [parent.sector_size])
+            if len(parent_sectors) != 1:
+                if not self.exists:
+                    msg = "The volume group %s cannot be created. Selected disks have " \
+                          "inconsistent sector sizes (%s)." % (self.name, parent_sectors)
+                else:
+                    msg = "Disk %s cannot be added to this volume group. LVM doesn't " \
+                          "allow using physical volumes with inconsistent (logical) sector sizes." % parent.name
+                raise ValueError(msg)
+
         if (self.exists and parent.format.exists and
                 len(self.parents) + 1 == self.pv_count):
             self._complete = True
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -19,9 +19,12 @@
 # Red Hat Author(s): David Lehman <dlehman@redhat.com>
 #
 
+import math
 import os
 import six
 import time
+
+from six.moves import reduce
 
 import gi
 gi.require_version("BlockDev", "2.0")
@@ -194,6 +197,14 @@ class MDRaidArrayDevice(ContainerDevice, RaidDevice):
             raise errors.DeviceError(e)
 
         self._level = level
+
+    @property
+    def sector_size(self):
+        if not self.exists:
+            # Least common multiple of parents' sector sizes
+            return reduce(lambda a, b: a * b // math.gcd(a, b), (int(p.sector_size) for p in self.parents))
+
+        return super(MDRaidArrayDevice, self).sector_size
 
     @property
     def chunk_size(self):
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -191,6 +191,21 @@ class StorageDevice(Device):
         return self
 
     @property
+    def sector_size(self):
+        """ Logical sector (block) size of this device """
+        if not self.exists:
+            if self.parents:
+                return self.parents[0].sector_size
+            else:
+                return LINUX_SECTOR_SIZE
+
+        block_size = util.get_sysfs_attr(self.sysfs_path, "queue/logical_block_size")
+        if block_size:
+            return int(block_size)
+        else:
+            return LINUX_SECTOR_SIZE
+
+    @property
     def controllable(self):
         return self._controllable and not flags.testing and not self.unavailable_type_dependencies()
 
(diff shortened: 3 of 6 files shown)


blivet/devices/lvm.py
```diff
@@ -356,6 +356,18 @@ def _remove_log_vol(self, lv):
     def _add_parent(self, parent):
         super(LVMVolumeGroupDevice, self)._add_parent(parent)
 
+        # we are creating new VG or adding a new PV to an existing one
+        if not self.exists or self._complete:
```
vpodzime: This `or self._complete` looks quite arbitrary here.

vojtechtrefny:
New version pushed, I've moved the `sector_size` code to `StorageDevice` with some special cases implemented for partitions and MD RAIDs.

blivet/devices/lvm.py
vojtechtrefny: That's because `_add_parent` is used also during populate for existing VGs and we want to skip this check for existing "broken" VGs but we also want to run this check when adding a brand new PV to an existing (complete) VG. I've changed the if and the comment a little, I hope it makes more sense now.

blivet/devices/storage.py
```diff
@@ -190,6 +190,18 @@ def raw_device(self):
         """ The device itself, or when encrypted, the backing device. """
         return self
 
+    @property
+    def sector_size(self):
+        """ Logical sector (block) size of this device """
+        if not self.exists:
+            return LINUX_SECTOR_SIZE
```
vpodzime: Shouldn't this generally be derived from the parent devices' sector sizes like it's done for MD arrays?

blivet/devices/md.py
```diff
@@ -195,6 +195,13 @@ def level(self, value):
 
         self._level = level
 
+    @property
+    def sector_size(self):
+        if not self.exists:
+            return max(p.sector_size for p in self.parents)
```
dwlehman: Is this enough, or does it need to be the least common multiple of the two? I imagine in most cases it will be the same, but will it always?

blivet/devices/lvm.py
```diff
@@ -356,6 +356,18 @@ def _remove_log_vol(self, lv):
...
+        # we are creating new VG or adding a new PV to an existing (complete) one
+        if not self.exists or (self.exists and self._complete):
+            parent_sectors = set([p.sector_size for p in self.pvs] + [parent.sector_size])
+            if len(parent_sectors) != 1:
+                if not self.exists:
+                    msg = "The volume group %s cannot be created. Selected disks have " \
+                          "inconsistent sector sizes (%s)." % (self.name, parent_sectors)
+                else:
+                    msg = "Disk %s cannot be added to this volume group. LVM doesn't " \
+                          "allow using physical volumes with incosistent (logical) sector sizes." % parent.name
```
dwlehman: _inconsistent_

dwlehman: (in the second error message)

tests/devices_test/lvm_test.py
```diff
@@ -352,6 +352,17 @@ def test_target_size(self):
...
         self.assertEqual(lv.size, orig_size)
 
+    def test_lvm_inconsistent_sector_size(self):
+        pv = StorageDevice("pv1", fmt=blivet.formats.get_format("lvmpv"),
+                           size=Size("1024 MiB"))
+        pv2 = StorageDevice("pv2", fmt=blivet.formats.get_format("lvmpv"),
+                            size=Size("1024 MiB"))
+
+        with patch("blivet.devices.StorageDevice.sector_size", new_callable=PropertyMock) as mock_property:
+            mock_property.__get__ = lambda _mock, pv, _class: 512 if pv.name == "pv1" else 4096
```
dwlehman: nice ;)

blivet/devices/md.py
vojtechtrefny: I'm not sure if you can have a device with sector size that is not multiple of 512, but it might be possible. I've changed the code to calculate LCM of the sector sizes.

blivet/devices/storage.py
vojtechtrefny: I've changed this to `parents[0].sector_size`. I think we can safely assume that for non-existing devices other than LVM and MD all parents must have same sector size so we can use only the first parent. (LVM now can't create such VG and MD is covered separately.) Thoughts?

blivet/devices/md.py
```diff
@@ -195,6 +198,14 @@ def level(self, value):
 
         self._level = level
 
+    @property
+    def sector_size(self):
+        if not self.exists:
+            # Least common multiple of parents' sector sizes
+            return reduce(lambda a, b: a * b // math.gcd(a, b), [p.sector_size for p in self.parents])
```
vpodzime: Could use a generator instead of list comprehension, right? :stuck_out_tongue:

blivet/devices/md.py
```diff
@@ -202,7 +202,7 @@ def level(self, value):
     def sector_size(self):
         if not self.exists:
             # Least common multiple of parents' sector sizes
-            return reduce(lambda a, b: a * b // math.gcd(a, b), [p.sector_size for p in self.parents])
+            return reduce(lambda a, b: a * b // math.gcd(a, b), [int(p.sector_size) for p in self.parents])
```
vpodzime: This should go to the previous commit, but no big deal, I think.

blivet/devices/storage.py
vpodzime: I think that's sufficient. The only other case coming to my mind are some non-LVM DM devices (think Docker, Stratis,...). But in those cases, it might very hard to tell what the "right" sector size would be.

blivet/devices/md.py
vojtechtrefny: Yes, fixed.

blivet/devices/md.py
vojtechtrefny: Fixed.

PR #808, opened 2019-10-31
Do not load module when creating an FS instance

vojtechtrefny:
The module is needed only for mounting the filesystem and mount
will autoload the modules so we don't need to load it when
creating the instance. Instead we can use modprobe with the
`--dry-run` option to check if the module is available to mark
the filesystem as (un)supported.

----
This was discovered by anaconda team -- they are not running their unit tests as root so `modprobe` will always. `modrpobe --dry-run` works even when running as a "normal" user because it actually doesn't try to load the module and we don't really need to load the module just to check if the filesystem is mountable/supported.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -55,10 +55,19 @@ from ..i18n import N_
 from .. import udev
 from ..mounts import mounts_cache
 
-from .fslib import kernel_filesystems, update_kernel_filesystems
+from .fslib import kernel_filesystems
 
 import logging
 log = logging.getLogger("blivet")
+
+import gi
+gi.require_version("GLib", "2.0")
+gi.require_version("BlockDev", "2.0")
+
+from gi.repository import GLib
+from gi.repository import BlockDev
+
+AVAILABLE_FILESYSTEMS = kernel_filesystems
 
 
 class FS(DeviceFormat):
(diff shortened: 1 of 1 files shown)


blivet/formats/fs.py
```diff
@@ -445,27 +448,28 @@ def do_check(self):
...
+    def check_module(self):
+        """Check if kernel module required to support this filesystem is available."""
+        if not self._modules or self.mount_type in AVAILABLE_FILESYSTEMS:
             return
 
         for module in self._modules:
+            # try modprobe --dry-run to check if the module is available
             try:
-                rc = util.run_program(["modprobe", module])
+                rc = util.run_program(["modprobe --dry-run", module])
```
vpodzime: This could use `blockdev.utils_have_kernel_module()`.

jkonecny12:
It would be great to fix this soon. Our tests now needs BTRFS module loaded manually to work, which is pretty annoying.

blivet/formats/fs.py
vojtechtrefny: Good point, I've changed the code to use libblockdev.

vojtechtrefny:
Jenkins, test this please.

jkonecny12:
Thanks a lot for your work and help!

PR #810, opened 2019-10-31
Add a simple script for running tests manually

vojtechtrefny:
We have a similar script in libblockdev and udisks. I plan to add more features later (e.g. running against installed version, some logging etc.) but for now the biggest feature is it allows running a simple test case using something like `sudo python3 tests/run_tests.py devices_test.device_properties_test.LVMLogicalVolumeDeviceTestCase`

### DIFF
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -17,7 +17,7 @@ class BlivetLintConfig(PocketLintConfig):
                                FalsePositive(r"No value for argument 'member_count' in unbound method call$"),
                                FalsePositive(r"No value for argument 'smallest_member_size' in unbound method call$"),
                                FalsePositive(r"Parameters differ from overridden 'do_task' method$"),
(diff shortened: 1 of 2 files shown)


tests/pylint/runpylint.py
```diff
@@ -17,7 +17,7 @@ def __init__(self):
                                FalsePositive(r"No value for argument 'member_count' in unbound method call$"),
                                FalsePositive(r"No value for argument 'smallest_member_size' in unbound method call$"),
                                FalsePositive(r"Parameters differ from overridden 'do_task' method$"),
-                               FalsePositive(r"Bad option value '(subprocess-popen-preexec-fn|try-except-raise)'"),
+                               FalsePositive(r"Bad option value '(subprocess-popen-preexec-fn|try-except-raise|environment-modify)'"),
```
vpodzime: Why is this needed on top of the `pylint: disable` comment?

vojtechtrefny: The `environment-modify` is quite new and this prevents older versions of pylint from complaining about unknown/unsupported option being disabled.

PR #811, opened 2019-11-01
Tell lvm to ignore skip-activation flag on lvs we are removing or otherwise modifying. (#1766498)

dwlehman:
The problem is that we try to activate an lv in order to remove the formatting before destroying the lv, but the activation fails because the user set the lv to ignore activation commands.

It is possible that we should be more cautious and possibly only pass `ignore_skip=True` when we're about to wipe the formatting. That would require setting up a generic mechanism for passing additional arguments, which seems quite disruptive given the problem.

Opinions?

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -640,6 +640,7 @@ class LVMLogicalVolumeBase(DMDevice, RaidDevice):
         self.uuid = uuid
         self.seg_type = seg_type or "linear"
         self._raid_level = None
+        self.ignore_skip_activation = 0
 
         self.req_grow = None
         self.req_max_size = Size(0)
@@ -1429,12 +1430,6 @@ class LVMSnapshotMixin(object):
         # the old snapshot cannot be setup and torn down
         pass
 
-    def _setup(self, orig=False):
-        """ Open, or set up, a device. """
-        log_method_call(self, self.name, orig=orig, status=self.status,
-                        controllable=self.controllable)
-        blockdev.lvm.lvactivate(self.vg.name, self._name, ignore_skip=True)
-
     @old_snapshot_specific
     def teardown(self, recursive=False):
         # the old snapshot cannot be setup and torn down
@@ -2043,12 +2038,12 @@ class LVMLogicalVolumeDevice(LVMLogicalVolumeBase, LVMInternalLogicalVolumeMixin
     def display_lv_name(self):
         return self.lvname
 
-    @type_specific
     def _setup(self, orig=False):
         """ Open, or set up, a device. """
         log_method_call(self, self.name, orig=orig, status=self.status,
                         controllable=self.controllable)
-        blockdev.lvm.lvactivate(self.vg.name, self._name)
+        ignore_skip_activation = self.is_snapshot_lv or self.ignore_skip_activation > 0
+        blockdev.lvm.lvactivate(self.vg.name, self._name, ignore_skip=ignore_skip_activation)
 
     @type_specific
     def _pre_create(self):
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -1012,6 +1012,10 @@ class LVMLogicalVolumeBase(DMDevice, RaidDevice):
         # set up the vg's pvs so lvm can remove the lv
         self.vg.setup_parents(orig=True)
 
+    def set_rw(self):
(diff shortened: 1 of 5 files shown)


prajnoha:
Yes, I think that suits the installation environment well.

The "skip activation" flag is supposed to be improving the situation with many LVs when the system is already set up/installed anyway. Here, during installation, it has its meaning to always activate everything so we can clean it up correctly.

vojtechtrefny:
> The problem is that we try to activate an lv in order to remove the formatting before destroying the lv, but the activation fails because the user set the lv to ignore activation commands.

Another solution would be to just ignore the failed activation and skip destroying the format if we are going to destroy the LV, but this would require a bigger change so I guess it's okay to just pass `ignore_skip=True` every time to fix the bug.

dwlehman:
I wonder if this should only be set by anaconda. Does it make sense for when users are managing storage outside the context of OS installation?

blivet/devices/lvm.py
```diff
@@ -2048,7 +2042,7 @@ def _setup(self, orig=False):
         """ Open, or set up, a device. """
         log_method_call(self, self.name, orig=orig, status=self.status,
                         controllable=self.controllable)
-        blockdev.lvm.lvactivate(self.vg.name, self._name)
+        blockdev.lvm.lvactivate(self.vg.name, self._name, ignore_skip=True)
```
vpodzime: You can drop the `@type_specific` decorator if this becomes the only version of the `_setup()` method.

dwlehman:
> I don't really like this change. I think that if people set this flag on their LVs, they have good reasons to do so. I myself have for example used it to prevent two devices with the same UUID from showing up in the system.

That's reasonable, I guess. I was concerned about use-cases we hadn't thought of.

> I think the proper way of handling the problematic case is to not set the device up when it's about to be destroyed.

The problem with that is the fs metadata will still be on the PV and may cause havoc the next time an LV gets created using the same starting extent.

I see a few options:
1. ignore the skip-activate always (the proposed patch)
2. ignore the skip-activate only when some new blivet flag is set
3. quietly handle the activation no-op and effectively skip format destruction (@vpodzime proposal AIUI)
4. put an lvm-specific block in `ActionDestroyFormat.execute` to override the skip only there

I wonder what other problems will arise from us trying to manage these lvs without any special logic.

dwlehman:
Another (very quick&dirty) option is to run `lvchange -k n <lv>` in `LVMLogicalVolumeDevice._pre_destroy` (or `_destroy`). The downside to this is it does not address the user who wants to use blivet to reformat an lv that is marked skip-activation.

dwlehman:
@vpodzime do you have opinions on the options I've presented?

vpodzime:
What if a new attribute `force_activation` is added to the LV objects and then when the device is scheduled for destroy, this attribute will be set to `True`?

vojtechtrefny:
> What if a new attribute force_activation is added to the LV objects and then when the device is scheduled for destroy, this attribute will be set to True?

Sounds good.

dwlehman:
> What if a new attribute `force_activation` is added to the LV objects and then when the device is scheduled for destroy, this attribute will be set to `True`?

I like this idea. What about when we schedule actions on the format? If we schedule format destroy, format resize, or label change, we probably need to do the same. I think this means we need to treat the new flag as a semaphore by incrementing it when we schedule such an action and decrementing it when/if we cancel them. When we run `lvactivate` we tell it to ignore skip-activate if the value of the flag is > 0. Does that make sense?

dwlehman:
I added another commit since the reporters of the bug apparently are also using `lvchange` to set the skip-activate LV as readonly. I added it to this pull request because it's needed to resolve the bug even though it's a separate topic. Let me know if you feel strongly that it should be a separate pull request and I'll open one.

vojtechtrefny:
Do we really want to set the LV to rw? If it's read-only it should be protected. I don't like the idea of silently changing the LV to read-write and deleting it.

vojtechtrefny:
> Do we really want to set the LV to rw? If it's read-only it should be protected. I don't like the idea of silently changing the LV to read-write and deleting it.

Read-only LV actually still can be deleted so I no longer have problem with changing it to rw to be able to wipe it.

dwlehman:
I have tested this pull request in a vm where I did an automatic partitioning install, manually set the root lv to be read-only and skip-activation, then reinstalled successfully with the patch.

PR #814, opened 2019-11-25
Add a new "id_path" attribute for iSCSI and FCoE disks

vojtechtrefny:
Anaconda needs the "ID_PATH" udev attribute so it's better to
add it to our API so they don't need to get it from "device_links".


poncovka:
@vojtechtrefny, I have realized that we are using the long identifier also for `NVDIMMNamespaceDevice`. Could you add the attribute `id_path` to this class, please? That should be all.

vojtechtrefny:
> @vojtechtrefny, I have realized that we are using the long identifier also for NVDIMMNamespaceDevice. Could you add the attribute id_path to this class, please? That should be all.

No problem, I've added it to nvdimm too.

PR #815, opened 2019-12-16
Align base partition sizes in PartitionFactory.

dwlehman:
Take the alignment of candidate disks into account when creating the partition request to ensure that the initial/base size is never too small to be allocatable on the specified disks. A simpler approach, like choosing a larger base size than `1 MiB`, is appealing but could be problematic for certain formatting (eg: biosboot).

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -409,7 +409,11 @@ def add_partition(disklabel, free, part_type, size, start=None, end=None):
         else:
             _size = size
 
-        alignment = disklabel.get_alignment(size=_size)
+        try:
+            alignment = disklabel.get_alignment(size=_size)
+        except AlignmentError:
+            alignment = disklabel.get_minimal_alignment()
+
         end_alignment = disklabel.get_end_alignment(alignment=alignment)
     else:
         alignment = parted.Alignment(grainSize=1, offset=0)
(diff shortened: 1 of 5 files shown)


dwlehman:
Alright, so I pushed a version that does what @vojtechtrefny suggested. Thanks -- it was a good idea. But as I was writing the commit message I was wondering why we can't just align the size up. I started poking around. It turns out that we're calling `DiskLabel.get_alignment(size=_part.req_size)` in `allocate_partitions` but we're not catching the exception it can raise if the size is smaller than the minimal I/O size advertised by the disk. That's the failure this pull request is addressing. So why can't we just catch the exception there, bump the size as needed, and move on? We align partition sizes up in other cases, just like we align down sometimes. So why not in this case? If that seems too controversial, how about just a `continue` to try allocating the partition from other disks?

I think the way to go here is to just bump the size if necessary to match the disk's minimum I/O size. What do you think? (Other than that this could go on forever.)

vojtechtrefny:
Moving the code to `allocate_partitions` also makes sense. Moving all the rounding and adjusting to one place is definitely an improvement.

dwlehman:
Totally rewritten. This time I'm just making minor adjustments to existing partition allocation code to account for the behavior of `DiskLabel.get_alignment` when it is passed a size smaller than the minimal grain size. What we do in the partition code is fall back to using the minimal alignment and adjusting the base size up to match it.

blivet/partitioning.py
```diff
@@ -409,7 +409,12 @@ def add_partition(disklabel, free, part_type, size, start=None, end=None):
         else:
             _size = size
 
-        alignment = disklabel.get_alignment(size=_size)
+        try:
+            alignment = disklabel.get_alignment(size=_size)
+        except AlignmentError:
+            alignment = disklabel.get_minimal_alignment()
+            _size = Size(alignment.grainSize)
```
vojtechtrefny: `_size` is used only in the `get_alignment` call above, we shouldn't need it here.

It might be useful to add a `log.debug` here just to be sure we know about the change to the minimal alignment.

dwlehman: We do log which alignment we've chosen just below. Is that sufficient?

dwlehman:
- removed the unnecessary line that updates/sets `_size` in `add_partition`
- rebased onto `3.2-devel`
- added API docs for the public alignment-related entry points in `DiskLabel` (including `get_alignment`, which should have been added long ago).

PR #817, opened 2019-12-25
Add support for LVMPV format resize

vojtechtrefny:


vojtechtrefny:
> This looks good to me, but I'm curious what the motivation was for this change. Is there some application that will be offering direct resize of PVs? Or perhaps some concept of a combined pv/vg resize?

This is actually an old RFE for blivet-gui, users will still need to resize an LV first to make some free space in the VG to be able to resize the PV.

PR #818, opened 2019-12-31
Allow running 'write_label' in dry run mode on non-existing devices

vojtechtrefny:
We should be able to schedule action to change the label on
a non-existing device so we can't check for the device/format
existence in the dry run mode (it's run when applying the action).

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -639,16 +639,16 @@ class FS(DeviceFormat):
             Raises a FSError if the label can not be set.
         """
 
-        if not self.exists:
-            raise FSError("filesystem has not been created")
-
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
> What was the catalyst/motivation for this change?

I was testing the support for swap relabel from #820 in blivet-gui which uses the configure action ( (`ActionConfigureFormat`) for label changes. I agree it's easier to change the label directly for non-existing devices but I didn't want to add a different way of changing the label for non-existing devices or force users to undo the create action and create the device/format again if they forgot to set the label.
As far as I know the `dry_run` option is used only by the configure action so this shouldn't cause any issues.

PR #825, opened 2020-01-17
Use SHA256 instead of MD5 for /proc/mounts hash calculation

vojtechtrefny:
MD5 is not available in the FIPS mode.

Resolves: rhbz#1792340

### DIFF
--- a/blivet/mounts.py
+++ b/blivet/mounts.py
@@ -165,13 +165,13 @@ class MountsCache(object):
                     self.mountpoints[(devspec, None)].append(mountpoint)
 
     def _cache_check(self):
-        """ Computes the MD5 hash on /proc/mounts and updates the cache on change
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
> Looks good to me with the assumption that SHA256 will always be available.

Both SHA1 and SHA256 are currently available in FIPS mode on RHEL 8:

```
>>> import hashlib
>>> hashlib.md5()
  File "<stdin>", line 1, in <module>
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
>>> hashlib.sha1()
<sha1 HASH object @ 0x7fb7abf409b8>
>>> hashlib.sha256()
<sha256 HASH object @ 0x7fb7abf40a80>
```

but SHA256 is probably "safer" for the future given recently discovered issues with SHA1.

PR #827, opened 2020-03-05
Fix name resolution for md member partitions. (#1798792)

vojtechtrefny:
Taken from #785

This is a good workaround for rhbz#1798792 which is currently a Fedora 32 blocker. Root of the problem is an mdadm udev rule which adds the `MD_DEVNAME` property (and other `MD_` properties) to the RAID member partitions which makes us think the partition is actually on the mdraid.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -202,7 +202,7 @@ def device_get_name(udev_info):
     """ Return the best name for a device based on the udev db data. """
     if "DM_NAME" in udev_info:
         name = udev_info["DM_NAME"]
-    elif "MD_DEVNAME" in udev_info:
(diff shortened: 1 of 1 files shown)


poojayaul:
@vojtechtrefny @dwlehman @vpodzime 
Hi, could you please help understand why the condition has "/md" added?
Because, the MDRAID Partitions need not have this "/md" in the sysfs path. For example, for md127p3, the sysfs path would be 
Only for the parent MD Device i.e. md127 the sysfs path will have "/md" ex: "/sys/devices/virtual/block/md127/md/"

Due to this condition, partitions are failed to be listed in Reclaim space tool( RHBZ#1862904). To fix this issue, i have modified the condition by removing string "/md" and it works for me. Created a pull request https://github.com/storaged-project/blivet/pull/885 for the same. Please help review it.

PR #831, opened 2020-03-20
Fix crash for devices without ID_PATH udev property (#1814920)

vojtechtrefny:
In some cases some NVDIMM namespaces don't have the ID_PATH set
and this makes the populate helper crash.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -406,7 +406,7 @@ def device_get_bus(udev_info):
 
 
 def device_get_path(info):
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
@poncovka can you take a look at this? I've checked the anaconda code and it looks like the only place where `disk.id_path` is used is the Advanced Storage spoke and empty string should be a valid value there.

poncovka:
Based on the current code, Anaconda should be able to deal with it. Thanks for info.

PR #833, opened 2020-03-25
Fix adding partition with automatic type and start sector (#1814805)

vojtechtrefny:
When adding a fourth partition blivet automatically creates an
extended partition and the new partition is created as logical
inside it. When manually specifying the start sector of the
partition this always fail, because after creating the extended
partition the start is not in the free space inside the extended
partition.
This change sets the desired start sector for the extended
partition and creates the logical partition inside it without
specifying the start sector.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -423,7 +423,10 @@ def add_partition(disklabel, free, part_type, size, start=None, end=None):
     sector_size = Size(disklabel.sector_size)
     if start is not None:
         if end is None:
-            end = start + size_to_sectors(size, sector_size) - 1
+            if size is None:
+                end = free.end
+            else:
+                end = start + size_to_sectors(size, sector_size) - 1
     else:
         start = free.start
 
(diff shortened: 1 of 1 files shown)


dwlehman:
My intention for this was that anyone adding a partition w/ start/end sectors would manually allocate the extended partition as needed. Is that an option?

vojtechtrefny:
> My intention for this was that anyone adding a partition w/ start/end sectors would manually allocate the extended partition as needed. Is that an option?

Manually creating the extended partition of course works. This happens when user doesn't specify `req_part_type` and this would be the fourth primary partition. I guess we could create primary partition in this case or just return an error forcing our users to always specify requested type when creating partitions with start sector.

dwlehman:
I am not really comfortable with the result having a start sector that differs from the one specifically requested. That's the whole point of allowing start/end sectors. We don't even align start/end sectors passed in.

vojtechtrefny:
I understand, but we need to do something. It's either creating the extended partition on given start sector and moving the logical partition or create the fourth partition as primary. Given these two options, I like the first one better.

dwlehman:
What if we don't allow the combination of start sector and implicit/automatic partition type?

PR #834, opened 2020-03-27
Correctly recognize EFI format on an MD RAID device (#1695913)

vojtechtrefny:
It's possible to have /boot/efi on an MD RAID device -- with the
older version of metadata and RAID1 it looks like a normal
partition and EFI can boot from it.

### DIFF
--- a/blivet/devices/md.py
+++ b/blivet/devices/md.py
@@ -486,6 +486,12 @@ class MDRaidArrayDevice(ContainerDevice, RaidDevice):
         """
         return (self.member_devices <= len(self.members)) or not self.exists
 
+    @property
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
> Doesn't anaconda's storage checker enforce reformatting of any mountpoints that begin with '/boot' ?

No, it's possible to reuse existing `/boot/efi` without formatting it and we want to allow it because it's possible to have a single `/boot/efi` for multiple systems.
I think this change should be safe, we'll just mark the filesystem as EFI boot instead of FAT32, user still need to choose it and mount it to /boot/efi without formatting, it's not going to happen automatically.

PR #837, opened 2020-04-16
Do not remove _netdev mount option specified manually by users

vojtechtrefny:
This option is used also for non-network devices like Clevis-Tang
protected LUKS devices so we can't remove it.

Resolves: rhbz#1722262

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -752,10 +752,11 @@ class StorageDevice(Device):
 
         netdev_option = "_netdev"
         option_list = self._format.options.split(",")
+        user_options = self._format._user_mountopts.split(",")
         is_netdev = any(isinstance(a, NetworkStorageDevice)
                         for a in self.ancestors)
         has_netdev_option = netdev_option in option_list
-        if not is_netdev and has_netdev_option:
+        if not is_netdev and has_netdev_option and netdev_option not in user_options:
             option_list.remove(netdev_option)
             self._format.options = ",".join(option_list)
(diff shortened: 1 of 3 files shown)


blivet/devices/storage.py
```diff
@@ -755,10 +755,7 @@ def _update_netdev_mount_option(self):
         is_netdev = any(isinstance(a, NetworkStorageDevice)
                         for a in self.ancestors)
         has_netdev_option = netdev_option in option_list
-        if not is_netdev and has_netdev_option:
-            option_list.remove(netdev_option)
-            self._format.options = ",".join(option_list)
-        elif is_netdev and not has_netdev_option:
+        if is_netdev and not has_netdev_option:
```
dwlehman, line 758: The block that can remove the option is so that, if the parent devices were changed such that network-connected disks were switched out for locally-connected disks, we would do the right thing in updating the options. Obviously we need to be able to honor this explicitly-passed option somehow, but I don't think we can remove that block unconditionally.

vojtechtrefny: I've been thinking about this and i wasn't able to come up with situation where the parents would change after creating the format instance. Anyway, I've changed the code to only remove the option if it wasn't originally specified by the user and I've also added a test to make sure the option is removed when added by us.

dwlehman: I don't know whether anyone uses this workflow, but it is possible to change the candidate disks and re-run `do_partitioning`, which could lead to the _netdev option being added or removed.

PR #839, opened 2020-04-24
Add "is_empty" property to StorageDevice

vojtechtrefny:
A simple check if the device is empty -- there is no filesystem
and partitioned devices don't have partitions (or only "magic"
partitions).

This is basically just a copy of the `_is_device_empty` function from Anaconda used for clearpart -- https://github.com/rhinstaller/anaconda/blob/master/pyanaconda/modules/storage/disk_initialization/configuration.py#L119 -- I don't think installer should deal with such "low level" stuff as magic partitions so this makes more sense in Blivet.

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -810,6 +810,13 @@ class StorageDevice(Device):
         return self._is_disk
 
     @property
+    def is_empty(self):
+        if not self.partitioned:
+            return self.format.type is None and len(self.children) == 0
+
+        return all(p.type == "partition" and p.is_magic for p in self.children)
+
+    @property
     def partitionable(self):
         return self._partitionable
 
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -237,6 +237,10 @@ class LVMVolumeGroupDevice(ContainerDevice):
 
         return True
 
+    @property
+    def is_empty(self):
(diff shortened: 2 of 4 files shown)


jkonecny12:
Thanks for making our life easier!

blivet/devices/storage.py
```diff
@@ -809,6 +809,13 @@ def direct(self):
     def is_disk(self):
         return self._is_disk
 
+    @property
+    def is_empty(self):
+        if not self.partitioned:
+            return self.format.type is None
```
dwlehman, line 815: I think you may need an exception for lvm vgs here, just in case someone accesses this property for one.

vojtechtrefny: Good point, I didn't realize there are devices without format that can be "not empty". Same applies to thinpools. I've added an extra check for number of children.

vojtechtrefny:
I've added some tests for both vgs and disks.

jkonecny12:
@poncovka I think we should create an issue to replace our solution by this when this is released. Or did you already created something like that?

poncovka:
@jkonecny12, see https://github.com/rhinstaller/anaconda/pull/2470. However, this is not released yet and I am not going to merge our PR before the Anaconda rebase.

jkonecny12:
Great!

PR #842, opened 2020-05-12
Basic LVM VDO support

vojtechtrefny:


blivet/devices/lvm.py
```diff
@@ -1789,8 +1789,132 @@ def populate_ksdata(self, data):
         data.pool_name = self.pool.lvname
 
 
+class LVMVDOPoolMixin(object):
```
dwlehman, line 1792: Is it not useful/possible to derive from `LVMThinPoolMixin` to save some of the boilerplate?

blivet/devices/lvm.py
```diff
@@ -1789,8 +1789,132 @@ def populate_ksdata(self, data):
...
+    def direct(self):
+        """ Is this device directly accessible? """
+        return False
+
+    def _create(self):
+        """ Create the device. """
+        raise NotImplementedError
+
+
+class LVMVDOLogicalVolumeMixin(object):
```
dwlehman, line 1843: How similar is this to `LVMThinLogicalVolumeMixin`? A little weird, but aren't they mostly the same?

blivet/devices/lvm.py
```diff
@@ -2066,6 +2196,11 @@ def readonly(self):
     def display_lv_name(self):
         return self.lvname
 
+    @property
+    @type_specific
+    def pool(self):
```
dwlehman, line 2201: Why does the lvm-vdo lv need this when the thinp lv doesn't?

blivet/devices/lvm.py
```diff
@@ -2167,6 +2302,19 @@ def resize(self):
...
+
+    @type_specific
+    def _remove_log_vol(self, lv):
+        pass
+
+    @property
+    @type_specific
+    def lvs(self):
+        return []
+
```
dwlehman, line 2317: Same here, more or less... why is this needed if it wasn't needed for thinp?

vojtechtrefny: Because before this thinpool was the only "type" that needed these properties so these could be part of the thinpool "interface". Now we have two types that need it, so these need to be part of the mixin interface.

blivet/devices/lvm.py
vojtechtrefny: The functions and properties are similar but usually differ a little. And the mixins are already very complicated and I'd like to avoid having `LVMVDOLogicalVolumeMixin` inheriting from `LVMThinLogicalVolumeMixin` and `LVMLogicalVolumeDevice` depending on both. This could go wrong very fast and complex dependencies in different versions of Python are huge mess.

I was thinking about adding another class, something like `LVMThinVolumeBase` (and `LVMPoolBase` for thinpool and vdopool) but I think this would make the code even less readable than now.

And also these two will differ more when we add support for creating VDO LVs/pools.

PR #844, opened 2020-05-13
Fix toggling container encryption in devicefactory (#1827254)

vojtechtrefny:
We need to remove the member (LUKSDevice) from the container's
parents before trying to remove the member itself. Removing it
from the container parents will also remove the container from
the member children allowing us to remove the member because
devices with children (non-leaf devices) can't be removed.


poncovka:
Thanks for fixing it :-)

PR #846, opened 2020-05-14
WIP: fstab management

dwlehman:
This is a PoC using libmount, python bindings for which are not always available. The extra code to enable functioning without libmount is reasonably straightforward, but having to do that is sad.

This initial/rough version includes some functions which I ended up abandoning in favor of a class, which sits just below the functions. Maybe someone has an opinion on which approach is better.

### DIFF
--- a/blivet/fstab.py
+++ b/blivet/fstab.py
@@ -1,0 +1,165 @@
+import logging
+import os
+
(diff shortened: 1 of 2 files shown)


dwlehman:
This can be resurrected later if needed.

PR #848, opened 2020-05-18
Allow setting size for non-existing LUKS devices

vojtechtrefny:
The LUKS device "inherits" its size from the parent so we are
just setting the parent's size here. This allows automatic size
adjustments for format minimum size for growable devices with
sizes smaller than format minimum size (e.g. using --grow in
kickstart for PVs).

Resolves: rhbz#1836269

-----

I'm not sure about this one. There are two issues here. First, the size setter currently doesn't work for `LUKSDevice` at all. I think it's python problem with mixing the new style properties (using the `@property` decorator) and the old style properties and the setter from `StorageDevice` is currently not being inherited which causes the installation crash with the `can't set attribute` error. Second issue is about setting size for the `LUKSDevice` during `_set_format` -- the device factory first creates a 1 MiB partition, formats to PV format and updates its size to 8 MiB (we are talking about `--size=1 --grow` in the kickstart) and after this we add the encryption by creating LUKSDevice which is now 6 MiB (size of the partition minus LUKS metadata) and try to format it to PV updating its size to 8 MiB which now fails.
I've added a new test case to cover this new LUKS size getter/setter functionality (setter is covered with `test_size_setter` inherited from `StorageDeviceSizeTest`) but I'm not sure if this is the best approach and this might affect other things not covered by our test suite.

### DIFF
--- a/blivet/devices/luks.py
+++ b/blivet/devices/luks.py
@@ -25,6 +25,7 @@ from ..devicelibs import crypto
 from ..storage_log import log_method_call
 from ..size import Size
 from ..tasks import availability
+from ..errors import DeviceError
 
 import logging
 log = logging.getLogger("blivet")
@@ -73,8 +74,7 @@ class LUKSDevice(DMCryptDevice):
             return self.parents[0].parents[0]
         return self.parents[0]
 
-    @property
-    def size(self):
+    def _get_size(self):
         if not self.exists:
             size = self.slave.size - crypto.LUKS_METADATA_SIZE
         elif self.resizable and self.target_size != Size(0):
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
Jenkins, test this please.

dwlehman:
I never added a setter to `LUKSDevice` because I figured we could adjust the slave/backing device size directly. Your patch looks reasonable to me, and I don't think there is much risk given that it has never been possible to set `size` directly for `LUKSDevice`.

It looks like the traceback from the referenced bug is a regression caused by caec289d8, which is amazing given its commit date of Feb 2019.

vojtechtrefny:
> t looks like the traceback from the referenced bug is a regression caused by [caec289](https://github.com/storaged-project/blivet/commit/caec289d8220fc9a8d8b3d6e99271394f4ef83fe), which is amazing given its commit date of Feb 2019.

I think something also changed in Python and the setter was inherited from StorageDevice before.

PR #851, opened 2020-06-03
Do not limit swap to 128 GiB

vojtechtrefny:
The limit was part of change to limit suggested swap size in
kickstart which doesn't use the SwapSpace._max_size so there is no
reason to limit this for manual installations.

Resolves: rhbz#1444232

rhel7-branch version of #850


vojtechtrefny:
Closing, we've decided to not do this for RHEL 7 branch.

PR #854, opened 2020-06-15
Add basic support for LVM writecache devices

vojtechtrefny:
Only "read" support for now, we need libblockdev support to able
to create a writecache device. With this we at least won't crash
on systems with existing writecache.

### DIFF
--- a/blivet/devices/cache.py
+++ b/blivet/devices/cache.py
@@ -54,12 +54,6 @@ class Cache(object):
         """
 
     @abc.abstractproperty
-    def mode(self):
-        """Mode of the cache (writeback/writethrough...)
-        :rtype: str
-        """
-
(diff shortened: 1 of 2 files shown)


blivet/devices/cache.py
```diff
@@ -53,12 +53,6 @@ def stats(self):
         :rtype: :class:`CacheStats`
         """
 
-    @abc.abstractproperty
-    def mode(self):
-        """Mode of the cache (writeback/writethrough...)
-        :rtype: str
-        """
-
```
dwlehman, line 61: Why remove this from the base class? There's not any reasonable value in this new class for it?

vojtechtrefny: No, dm-writecache is very simple, there is no mode or metadata.

PR #855, opened 2020-06-15
Standalone DM Integrity devices support improvement

vojtechtrefny:
Just few small change so we are able to remove these devices.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -457,6 +457,20 @@ def device_get_slaves(info):
     return slaves
 
 
+def device_get_holders(info):
+    """ Return a list of udev device objects representing this device's holders. """
+    holders_dir = device_get_sysfs_path(info) + "/holders/"
+    names = list()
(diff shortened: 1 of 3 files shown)


blivet/udev.py
```diff
@@ -457,6 +457,20 @@ def device_get_slaves(info):
...
+def device_get_holders(info):
+    """ Return a list of udev device objects representing this device's holders. """
+    holders_dir = device_get_sysfs_path(info) + "/holders/"
+    names = list()
+    if os.path.isdir(holders_dir):
+        names = os.listdir(holders_dir)
+
+    holders = list()
+    for name in names:
+        holders.append(get_device(device_node="/dev/" + name))
```
vpodzime, line 469: Come on, this looks like old-grandma's Python from '90s :wink:

vojtechtrefny: That's a copy-paste from `device_get_slaves` so it probably is from the 90s :-)

PR #856, opened 2020-06-16
Fix resizable property for partitions

vojtechtrefny:
This fixes isses in anaconda unit tests where some non-existing
partitions can have disk set to None. We don't allow resizing of
non-existing devices so it is safe to return False here.


jkonecny12:
Great, thanks for this!

PR #863, opened 2020-06-30
Terminology cleanups

dwlehman:


blivet/devices/luks.py
```diff
@@ -66,26 +66,22 @@ def __init__(self, name, fmt=None, size=None, uuid=None,
 
     @property
     def raw_device(self):
-        return self.slave
-
-    @property
-    def slave(self):
```
vojtechtrefny, line 72: Both blivet-gui and anaconda use this property. I think we should deprecate this now and remove it later in 3.3 to give our users some time to adjust.

blivet/udev.py
```diff
@@ -444,18 +444,18 @@ def device_get_devname(info):
     return info.get('DEVNAME')
 
 
-def device_get_slaves(info):
-    """ Return a list of udev device objects representing this device's slaves. """
-    slaves_dir = device_get_sysfs_path(info) + "/slaves/"
+def device_get_parents(info):
+    """ Return a list of udev device objects representing this device's parents. """
+    parents_dir = device_get_sysfs_path(info) + "/slaves/"
```
vojtechtrefny, line 449: I wonder if the sysfs `slaves` directory will change in the future. That would be a huge API break.

blivet/devices/luks.py
dwlehman: Sounds reasonable. I'll adjust it.

blivet/udev.py
```diff
@@ -39,7 +39,7 @@
 global_udev = pyudev.Context()
 log = logging.getLogger("blivet")
 
-device_name_blacklist = []
+ignored_device_names = []
```
poncovka, line 42: Hi, Anaconda is using this variable, so we should coordinate the releases. See: https://github.com/rhinstaller/anaconda/pull/2698

vojtechtrefny: Oh, I didn't notice this one. We should probably also only deprecate this one and remove it in 3.3.

dwlehman: I agree, but I don't think our decorator will work on variables. Any ideas on how to indicate that it is deprecated? The only idea I have is to add the to-be-removed attributes to the API docs (they aren't there now AFAIK) with a note about their deprecation and impending removal.

dwlehman: Maybe we should add a note to docs for deprecated items. That will both make the deprecations more visible and provide a way to indicate deprecation of a variable.

vojtechtrefny: > but I don't think our decorator will work on variables

We could do something really crazy like create a subclass of `list` and decorate its `insert` method, but that I think adding a comment a mentioning the deprecation in documentation should be good enough.

dwlehman: It turns out none of the names I changed in this pull request appear in the API docs. I'm not sure if I want to add this to the API just so we can deprecate/remove it. Maybe we should fall back on the original suggestion to simply coordinate the releases with anaconda.

vojtechtrefny: In that case I guess we can just do this without the deprecation. I've already created 3.3-devel branch so let's just move this to 3.3.

vojtechtrefny:
Closing, new version in #889

Issue #864, opened 2020-07-03
lsblk: unknown column: FSAVAIL,FSUSE%,MOUNTPOINT'


yizhanglinux:
Found this error log on /tmp/blivet.log when I testing storage role on RHEL8

2020-07-03 01:49:49,345 INFO program/MainThread: Running [2] e2fsck -V ...
2020-07-03 01:49:49,351 INFO program/MainThread: stdout[2]:
2020-07-03 01:49:49,351 INFO program/MainThread: stderr[2]: e2fsck 1.45.6 (20-Mar-2020)
        Using EXT2FS Library version 1.45.6, 20-Mar-2020

2020-07-03 01:49:49,351 INFO program/MainThread: ...done [2] (exit code: 0)
2020-07-03 01:49:49,352 DEBUG blivet/MainThread: get_format('ext4') returning Ext4FS instance with object id 0
2020-07-03 01:49:49,355 DEBUG blivet/MainThread:                Ext4FS.supported: supported: True ;
2020-07-03 01:49:49,355 DEBUG blivet/MainThread: trying to set new default fstype to 'ext4'
2020-07-03 01:49:49,359 DEBUG blivet/MainThread:                      Ext4FS.supported: supported: True ;
2020-07-03 01:49:49,359 DEBUG blivet/MainThread: get_format('ext4') returning Ext4FS instance with object id 1
2020-07-03 01:49:49,362 DEBUG blivet/MainThread:                   Ext4FS.supported: supported: True ;
2020-07-03 01:49:49,362 INFO program/MainThread: Running... lsblk --bytes -a -o NAME,SIZE,OWNER,GROUP,MODE,FSTYPE,LABEL,UUID,PARTUUID,FSAVAIL,FSUSE%,MOUNTPOINT
2020-07-03 01:49:49,369 INFO program/MainThread: stderr:
2020-07-03 01:49:49,369 INFO program/MainThread: b'lsblk: unknown column: FSAVAIL,FSUSE%,MOUNTPOINT'
2020-07-03 01:49:49,370 DEBUG program/MainThread: Return code: 1
2020-07-03 01:49:49,370 DEBUG blivet/MainThread: lsblk output:

2020-07-03 01:49:49,371 INFO blivet/MainThread: resetting Blivet (version 3.2.2) instance <blivet.blivet.Blivet object at 0x7f5517c0ec50>

PR #865, opened 2020-07-03
Do not use FSAVAIL and FSUSE% options when running lsblk

vojtechtrefny:
These options were added in util-linux 2.33 which is not available
on older systems so we should not use these.

Fixes: #864


vojtechtrefny:
Btw. we are not alone, anaconda has the same problem: https://github.com/rhinstaller/anaconda/pull/2701

PR #866, opened 2020-07-09
set allowed disk labels for s390x as standard ones (msdos + gpt) plus…

sharkcz:
… dasd

This will solve issues when a SCSI or NVMe disk with GPT partition table
is used with a s390x machine (rhbz#1827066, rhbz#1854110).


sharkcz:
successfully tested with 3TB zFCP/SCSI storage
without the fix Anaconda reports "No usable disk selected", with the fix the 3TB LUN is accepted as target device

vojtechtrefny:
Jenkins, ok to test.

PR #868, opened 2020-07-10
Do not propagate ped exception from add_partition

vojtechtrefny:
Our users shouldn't need to catch parted exception, we should use
our PartitioningError where possible.

Related: rhbz#1855386


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

Issue #859, opened 2020-06-18
RFE: Add support for XFS grow


dwlehman:
It should be fairly straightforward to implement and users can reasonably expect blivet to have this capability. FWIW this came up in discussions about the [storage role](https://github.com/linux-system-roles/storage).

PR #872, opened 2020-07-14
XFS grow support

vojtechtrefny:
Fixes: #859

### DIFF
--- a/blivet/tasks/fsresize.py
+++ b/blivet/tasks/fsresize.py
@@ -20,7 +20,10 @@
 # Red Hat Author(s): Anne Mulhern <amulhern@redhat.com>
 
 import abc
+import os
+import tempfile
 
+from contextlib import contextmanager
 from six import add_metaclass
 
 from ..errors import FSError
@@ -31,6 +34,9 @@ from . import availability
 from . import task
 from . import fstask
 from . import dfresize
+
+import logging
+log = logging.getLogger("blivet")
 
 
 @add_metaclass(abc.ABCMeta)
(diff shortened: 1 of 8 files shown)


vojtechtrefny:
So I finally figured out why the partitioning tests are failing. The problem is in the libblockdev FS plugin which uses libparted for FAT filesystem resize and registers its own libparted exception handler. Unfortunately this overwrites the exception handler registered by pyparted which means pyparted now throws generic `_ped` exceptions instead of the custom ones.
I'm now trying to find a way to either fix this in libblockdev (e.g. not overwrite existing handlers) or to force pyparted to re-register its handler after libblockdev rewrites it, but it looks like we can't use libblockdev FS plugin for now.

blivet/tasks/fsresize.py
```diff
@@ -115,6 +121,54 @@ def args(self):
...
+        return str(self.fs.target_size.convert_to(self.unit) / self._get_block_size())
+
+    @property
+    def args(self):
+        return [self.fs.system_mountpoint, "-D", self.size_spec()]
+
+    def do_task(self):
+        """ Resizes the XFS format. """
+
+        with self._do_temp_mount():
```
vojtechtrefny, line 168: XFS must be mounted for resize. I've decided to mount the device if it isn't mounted, but it might be safer to let user decide and mount the device manually. We could simply add check to `_pre_resize`. Thoughts?

vojtechtrefny:
I've decided to rewrite the XFS resize to not use libblockdev. I plan to remove the libparted dependency from the FS plugin in 3.0 so we can change this (together with other filesystem operations in blivet) to use libblockdev later.

blivet/tasks/fsresize.py
dwlehman: The way you have it now seems reasonable to me.

dwlehman:
I discussed this briefly w/ @sandeen and invited him to take a look if he's inclined to.

sandeen:
I see no obvious problem with temporarily mounting the filesystem if a resize is requested for an unmounted device.  Is it unmounted again afterwards to return it to the prior state?

PR #874, opened 2020-07-15
Ignore zRAM devices in VMBackedTestCase

vojtechtrefny:
zRAM is now enabled by default on latest Fedora and because it
looks like a disk we try to use it during the tests.

### DIFF
--- a/tests/vmtests/vmbackedtestcase.py
+++ b/tests/vmtests/vmbackedtestcase.py
@@ -2,7 +2,7 @@
 import os
 import unittest
 
-from blivet import Blivet
+from blivet import Blivet, udev
 
 
 @unittest.skipUnless(os.environ.get("VM_ENVIRONMENT"), "vm only test")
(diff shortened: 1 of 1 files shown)


tests/vmtests/vmbackedtestcase.py
```diff
@@ -49,6 +49,9 @@ def set_up_storage(self):
             This will write the configuration to whatever disk images are
             defined in set_up_disks.
         """
+
+        udev.device_name_blacklist = [r'^zram']
```
vojtechtrefny, line 53: We'll to change this in #863

vojtechtrefny:
This is just a quick fix for now, we should probably add a special device type for zram devices in the future (it is `DiskDevice` right now).

Note: Anaconda also blacklists zram devices so it isn't a problem during installation.

tests/vmtests/vmbackedtestcase.py
dwlehman: Should we add this to the main list in `blivet.udev`?

vojtechtrefny: It is a block device and it can be used for other things than swap -- [documentation suggests](https://www.kernel.org/doc/Documentation/blockdev/zram.txt) using it for `/tmp` or `/var` so people might want to use blivet-gui to format it for example.

PR #877, opened 2020-07-16
Add exFAT to filesystems we recognize

vojtechtrefny:
exFAT support is slowly coming to Fedora so we should be able to
recognize it instead of marking is as an unknown "DeviceFormat".


vojtechtrefny:
The exfatprogs package for Fedora is still being reviewed -- https://bugzilla.redhat.com/show_bug.cgi?id=1824156 -- but we should consider adding "full" support when it's available.

PR #878, opened 2020-07-20
Make safe_device_name device type specific

vojtechtrefny:
This allows us to have a different safe device names for different
device types and for example allow characters like "@" and "/" for
btrfs only.


vojtechtrefny:
> Looks good to me. I don't know if I love using the devicefactory type constants outside the factory flow, but they are there and serve the purpose well enough for now.

I was thinking about using strings, but that would be even worse. We should think about adding some `constants` file for these purposes.

PR #879, opened 2020-07-21
Ignore devices marked as hidden in sysfs (#1856974)

vojtechtrefny:
NVMe multipath devices have a special internal hidden devices that
are visible in /sys/block but not directly usable so we should
ignore these as well as all devices with the hidden attribute.


vojtechtrefny:
lsblk also ignores these devices, see https://github.com/karelzak/util-linux/commit/c8487d854ba5cf5bfcae78d8e5af5587e7622351

If this looks like a too big/dangerous change we could also fix the bug in anaconda by blacklisting the `nvme*c*` devices.

PR #881, opened 2020-07-22
Do not ignore unknown/unsupported device mapper devices

vojtechtrefny:
For these devices we should add a "plain" DMDevice to the tree.

### DIFF
--- a/blivet/populator/helpers/dm.py
+++ b/blivet/populator/helpers/dm.py
@@ -22,7 +22,6 @@
 
 from ... import udev
 from ...devices import DMDevice
-from ...devicelibs import lvm
 from ...storage_log import log_method_call
 from .devicepopulator import DevicePopulator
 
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
The change is very simple, we already do this for LiveCD DM devices so it shouldn't cause problems with other "unknown" DM devices. I've tested this with stratis and dm-crypt (other than LUKS) devices and everything seems to be working fine.

vojtechtrefny:
Works also with the standalone VDO devices:
```
DMDevice instance (0x7f4a4f67df40) --
  name = bd-test-vdo  status = True  id = 37
  children = []
  parents = ['existing 8 GiB disk sda (27)']
  uuid = None  size = 24 GiB
  format = existing None
  major = 0  minor = 0  exists = True  protected = True
  sysfs path = /sys/devices/virtual/block/dm-0
  target size = 24 GiB  path = /dev/mapper/bd-test-vdo
  format args = []  original_format = None  target = None  dm_uuid = VDO-8729f5af-b8f7-4c10-9b46-94bce91901b6
```

PR #882, opened 2020-07-30
Fix reading hidden sysfs attribute

vojtechtrefny:
util.get_sysfs_attr returns a string so bool(hidden) is always
True and we hide all devices.


vojtechtrefny:
This was a really stupid mistake, don't know how could I miss it. I should have waited and not merge it when Jenkins was down :-)

PR #887, opened 2020-08-08
Fix Failed to instantiate MDRAID Paritions in Reclaim Space Tool RHBZ#1862904

poojayaul:
Sysfs path for MDRAID Partitions need not have "/md" existing. For example, the sysfs path for the md127p3 partition would be

Due to the below condition in device_get_name() function it returns SYS NAME instead of MD_DEVNAME for MDRAID Partitions

```
elif "MD_DEVNAME" in udev_info and os.path.exists(device_get_sysfs_path(udev_info) + "/md"):
       mdname = udev_info["MD_DEVNAME"]
else:
        name = udev_info["SYS_NAME"]
```

This causes existing MDRAID partitions fail to be instantiated because, in function getPartitionByPath(self.path) self.path will be of format SYS NAME: /dev/md/md127pX and self.partitions will be a list of partitions in MD_DEVNAME format i.e. /dev/md/127pX, and the "partition.path == path" condition fails resulting in returning None object and thereby exceptions.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -201,11 +201,14 @@ def __is_blacklisted_blockdev(dev_name):
 
 def device_get_name(udev_info):
     """ Return the best name for a device based on the udev db data. """
+    dev_is_partition = device_is_partition(udev_info)
+    dev_sysfs_path = device_get_sysfs_path(udev_info)
     if "DM_NAME" in udev_info:
         name = udev_info["DM_NAME"]
-    elif "MD_DEVNAME" in udev_info and os.path.exists(device_get_sysfs_path(udev_info) + "/md"):
+    elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md")
+                                        or (dev_is_partition and os.path.exists(dev_sysfs_path))):
         mdname = udev_info["MD_DEVNAME"]
-        if device_is_partition(udev_info):
+        if dev_is_partition:
             # for partitions on named RAID we want to use the raid name, not
             # the node, e.g. "raid1" instead of "md127p1"
             partnum = udev_info["ID_PART_ENTRY_NUMBER"]


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
Thank you for the patch. It looks good to me in general, but I'll need to do some manual testing first.

MD in UDev is unfortunately a huge mess. We did the original change because RAID memeber disks and sometimes even member partitions have the `MD_DEVNAME` property and because of this we were confusing disks with MD devices.
Partitions on MD are even more complicated, especially with "named" RAIDs and sometimes device paths and other UDev properties are different with newly crated arrays (autostarted with `mdadm --create --run`), arrays assembled with `mdadm --assemble --scan` and arrays assembled during boot. And of course there are small differences with different versions of mdadm and UDev...
We definitely need a better test coverage for MD RAID. Currently it looks like every fix we do breaks something else :-)

vojtechtrefny:
PEP8 check failed with:
```
blivet/udev.py:209:41: W503 line break before binary operator
```

poojayaul:
@vojtechtrefny we followed W504 style of line break, but it's failing

poojayaul:
@vojtechtrefny if "/md" checking is mandatory for differentiating MD devices and other disks then this is the new patch  https://github.com/poojayaul/blivet/commit/39998bdceb8517fbc62fbc7faec770dd4ee1e782

vojtechtrefny:
Your patch unfortunately doesn't work with all setups. This condition:

```
    elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md")
                                        or (dev_is_partition and os.path.exists(dev_sysfs_path)))
```

will be true for member partitions. These have `MD_DEVNAME` set in UDev but we need to use `SYS_NAME` for these.

I've created a new patch in #888 that should (I hope) work for all combinations of MD devices, members and partitions. It also includes tests for these combinations.

poojayaul:
Hi @vojtechtrefny ,

In my new commit I have mentioned earlier(https://github.com/poojayaul/blivet/commit/39998bdceb8517fbc62fbc7faec770dd4ee1e782, for which I have not created pull request, but kept open for reviewing with you), it does name the Member paritions with SYS_NAME. 
See below explanation of the patch.

FOR MDRAID ARRAYS, below condition will be true and hence names MD_DEVNAME
`elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md")`

For partitions on RAID,  below condition will be true and hence names MD_DEVNAME
`elif "MD_DEVNAME" in udev_info and (dev_is_partition and os.path.exists(os.path.split(dev_sysfs_path)[0] + "/md"))):`

For MD RAID Members(Disks/Partitions), this condition fails and hence uses SYS_NAME, as it falls to else section
```
elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md") or
                                       (dev_is_partition and os.path.exists(os.path.split(dev_sysfs_path)[0] + "/md"))):

else: 
name = udev_info["SYS_NAME"]
```

Can you please help me understand if this patch also doesn't work or how it is different from the Pull request #888 https://github.com/storaged-project/blivet/pull/888/commits/4bf7ef1f6b63d025dd047fe5f1c6dea4ef6c3b40

poojayaul:
> @vojtechtrefny if "/md" checking is mandatory for differentiating MD devices and other disks then this is the new patch [poojayaul@39998bd](https://github.com/poojayaul/blivet/commit/39998bdceb8517fbc62fbc7faec770dd4ee1e782)

Hi @vojtechtrefny ,
In my new commit I have mentioned earlier(poojayaul@39998bd, for which I have not created pull request, but kept open for reviewing with you), it does name the Member paritions with SYS_NAME.
See below explanation of the patch.

  FOR MDRAID ARRAYS, below condition will be true and hence names MD_DEVNAME
   ` elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md")`

  For partitions on RAID, below condition will be true and hence names MD_DEVNAME
   ` elif "MD_DEVNAME" in udev_info and (dev_is_partition and os.path.exists(os.path.split(dev_sysfs_path)[0] + "/md"))):`

  For MD RAID Members(Disks/Partitions), this condition fails and hence uses SYS_NAME, as it falls to else section
   ```
 elif "MD_DEVNAME" in udev_info and (os.path.exists(dev_sysfs_path + "/md") or
                                       (dev_is_partition and os.path.exists(os.path.split(dev_sysfs_path)[0] + "/md"))):
    else: 
                  name = udev_info["SYS_NAME"]
```
Can you please help me understand if this patch also doesn't work or how it is different from the Pull request #888 4bf7ef1

vojtechtrefny:
>> @vojtechtrefny if "/md" checking is mandatory for differentiating MD devices and other disks then this is the new patch poojayaul@39998bd

> Can you please help me understand if this patch also doesn't work or how it is different from the Pull request #888 4bf7ef1

Sorry I didn't realize this was a separate commit, I thought you've updated this PR (you can do that by force pushing to the branch). Your new commit looks good, but I'd still prefer using #888 if it also works for your use case -- it's a bit easier to understand thanks to the usage of `device_is_md`.

vojtechtrefny:
> @vojtechtrefny if "/md" checking is mandatory for differentiating MD devices and other disks then this is the new patch poojayaul/blivet@39998bd

What do you think about this and #888 @dwlehman @japokorn ?

dwlehman:
The commit attached to this pull request doesn't fix the issue completely. The referenced commit (poojayaul/blivet@39998bd) does fix the issue, but is just a bit less clean and clear than the patch from #888. My main complain is that your patch, while perhaps a bit more economical, relies on explicit checks for a 'md/' subdir while #888 uses the existing logic for that in `device_is_md`.

vojtechtrefny:
I've merged #888 so I'm closing this now.

PR #888, opened 2020-08-12
Fix name resolution for MD devices and partitions on them 

vojtechtrefny:
I hope the tests now cover all the combinations and `device_get_name` now finally forks properly.

There are two potential issues:
- Blivet intentionally ignores partitions on RAIDs on top of partitions. I guess this is ok, I didn't realize this when gathering the MD UDev data so we have test cover for this too so even if `populator` ignores these partitions, `udev.device_get_name` works correctly.
- This won't work with MD RAID on top of another MD RAID. I really hope we don't want to support this. Especially MD member partition on an array sounds really crazy and I can't image code that would work on setup like this.

This should also replace #887 

I plan to do some more manual testing, but so far it looks like this really covers everything.


Yurri-huang:
This code change seems to be reopened about DiskDevice get members crashed bug
image

I don't understand why the path check for md is removed here:
image

This is the data I observed:
image

The program crashed in storage.reset()

PR #890, opened 2020-08-19
Account for pmspare grow when adjusting thinpool metadata size

vojtechtrefny:
When making space for the bigger metadata we also need to make
extra space for pmspare because its size depends on the metadata
size.

Resolves: rhbz#1868623


vpodzime (approved):
ACK

PR #895, opened 2020-09-04
Ignore new pylint warning W0707 "raise-missing-from"

vojtechtrefny:
This warns about not using the "raise ... from" syntax which is
not available in Python 2 so we can't use it.


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #897, opened 2020-09-09
Avoid using unnecessary udev.settle calls

vojtechtrefny:
Apparently it is completely normal to have hundreds of btrfs snapshots and blivet is currently not very good in handling this situation. `reset` is relatively fast, but `teardown_all` (which anaconda runs multiple times during installation) can take more than 10 minutes on systems with 500 snapshots ([rhbz#1876162](https://bugzilla.redhat.com/show_bug.cgi?id=1876162)).

I was able to find two issues in our code:
- `MountsCache` uses `udev.resolve_device` for every mountpoint query. `resolve_device` itself uses `udev.settle` and also iterates over all devices in the UDev database.
- `StorageDevice._pre_teardown` always runs `udev.settle` even when the `teardown` is no-op. This means we run `udev.settle` 3 times for every snapshot during `teardown_all`.

### DIFF
--- a/blivet/devices/storage.py
+++ b/blivet/devices/storage.py
@@ -425,7 +425,6 @@ class StorageDevice(Device):
             self.original_format.teardown()
         if self.format.exists:
             self.format.teardown()
-        udev.settle()
         return True
 
     def _teardown(self, recursive=None):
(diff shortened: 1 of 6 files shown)


Conan-Kudo (approved):

vpodzime (approved):
*Try to not use~d~* in the first commit message. Looks good to me otherwise.

dwlehman (approved):
You could also set a flag based on the initial status of the format and then conditionally call `settle` only when the flag indicates that the format was initially active. It would change less, but I'm not sure if it's actually better. It's probably a good thing to explicitly call settle every time we tear down a format, whether that's done in `DeviceFormat.teardown` or in the subclasses as you've done.

PR #901, opened 2020-09-18
edd: Fix UnboundLocalError when trying to close fd in collect_mbrs

vojtechtrefny:
The "fd" variable is not defined when the "os.open" fails (e.g.
when the device we try to open doesn't exist).

Follow up for #899


vojtechtrefny:
@nijinashok Can you please take a look at this to be sure it doesn't break your fix? (It shouldn't, just to be safe.)

nijinashok:
I just tested this and it works great! It's not holding devices.

PR #902, opened 2020-09-18
Let parted fix fixable issues with partition table

vojtechtrefny:
This will automatically fix issues like GPT partition table not
covering whole device after disk size change.

Resolves: rhbz#1846869

-----

I'd like to know your opinion about this change. I'm not sure if this is the best way to fix the problem (blivet crashing when trying to add partition to such disk). There is no way to tell what issue is being fixed (other than exception message) and this also means we'll do partition table changes during populate without asking. On the other hand, there is no way to detect this problem, so we can't mark the disk as "broken" and avoid using it in autopart.

### DIFF
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -64,6 +64,9 @@ def parted_exn_handler(exn_type, exn_options, exn_msg):
     if exn_type == parted.EXCEPTION_TYPE_ERROR and \
        exn_options == parted.EXCEPTION_OPT_YES_NO:
         ret = parted.EXCEPTION_RESOLVE_YES
+    elif exn_type == parted.EXCEPTION_TYPE_WARNING and \
+            exn_options & parted.EXCEPTION_RESOLVE_FIX:
+        ret = parted.EXCEPTION_RESOLVE_FIX
     return ret
 
 


vpodzime:
> There is no way to tell what issue is being fixed
The exception handler can set some flag, right? I think it would be nice to inform the user about the change. And it's not committed to the disk during `populate()`, I hope.?

vojtechtrefny:
> And it's not committed to the disk during `populate()`, I hope.?

It is. The "fix" action immediately changes the GPT table.

vpodzime:
> It is. The "fix" action immediately changes the GPT table.

Oh, I though this was only happening in the model. :thinking:

blivet/populator/populator.py
```diff
@@ -64,6 +64,9 @@ def parted_exn_handler(exn_type, exn_options, exn_msg):
     if exn_type == parted.EXCEPTION_TYPE_ERROR and \
        exn_options == parted.EXCEPTION_OPT_YES_NO:
         ret = parted.EXCEPTION_RESOLVE_YES
+    elif exn_type == parted.EXCEPTION_TYPE_WARNING and \
+            exn_options & parted.EXCEPTION_RESOLVE_FIX:
+        ret = parted.EXCEPTION_RESOLVE_FIX
```
vpodzime, line 69: I think this should set some flag and the installer should inform the user.

vpodzime (commented):
Looks good to me otherwise. Lesser evil...

blivet/populator/populator.py
vojtechtrefny: I don't think Anaconda has a way to inform user about this change. We are already logging the exception so that should be enough for debugging.

It might be useful to add some callback that applications like blivet-gui could use to actually ask user, but that is something we can do in the future.

vpodzime: Works for me :+1:

vpodzime (approved):
ACK

vojtechtrefny:
I'd like to merge this soon so last chance to review this. @dwlehman @japokorn

dwlehman (approved):
Looks good to me. This basically follows what we already do with yes/no questions to resolve similar issues as needed.

PR #903, opened 2020-10-05
Start the iscsi-init service (#1880673)

poncovka:
The startup method of the iSCSI class should start the iscsi-init service that
generates the /etc/iscsi/initiatorname.iscsi file. Otherwise, the file might
not exist and the getter of the initiator property might raise an exception.

Resolves: rhbz#1880673

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -364,6 +364,9 @@ class iSCSI(object):
         if not has_iscsi():
             return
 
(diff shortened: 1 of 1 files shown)


blivetghoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
Looks good to me.

jkonecny12 (commented):
LGTM! ;)

PR #904, opened 2020-10-13
Enable packit RPM builds on pull requests

vojtechtrefny:


.packit.yaml
```diff
@@ -0,0 +1,12 @@
+actions:
+  get-current-version:
+    - "python3 ./setup.py --version"
```
jkonecny12, line 3: You should be able to replace this by https://packit.dev/docs/configuration/#upstream_tag_template.

```
upstream_tag_template: blivet-{version}-1
```

jkonecny12 (commented):
Just a small improvement.

.packit.yaml
vojtechtrefny: That's probably not going to work for us with the correct version only on the `-release` branch.

dwlehman (approved):
Nice! Looks good to me.

.packit.yaml
jkonecny12: OK ;)

PR #905, opened 2020-10-13
Fix blivet dependencies on Debian in README

vojtechtrefny:
libblockdev-part2 was missing from the list of libblockdev plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ To install these dependencies use following commands:
        # dnf install python3-blockdev libblockdev-plugins-all python3-bytesize libbytesize python3-pyparted parted libselinux-python3
  * On Debian and Ubuntu based distributions:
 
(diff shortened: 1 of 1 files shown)


dwlehman (approved):
The '2' at the end of each plugin name is unfamiliar, but your patch looks okay to me since that's clearly an established pattern.

vojtechtrefny:
> The '2' at the end of each plugin name is unfamiliar, but your patch looks okay to me since that's clearly an established pattern.

Libraries packaged in Debian always have the soversion in the package name: https://packages.debian.org/search?keywords=libblockdev-part2

vojtechtrefny:
Blivet actually doesn't need the part plugin. I noticed blivet-gui tests are failing when working on https://github.com/storaged-project/blivet-gui/pull/243 and immediately it's a missing blivet dependency, but I'm using libblockdev to create some preexisting storage configuration in blivet-gui test suite so it's a blivet-gui test dependency, not dependency of blivet.

PR #909, opened 2020-11-02
LVM VDO create support

vojtechtrefny:


vojtechtrefny:
I added a special file with LVM VDO documentation so together with the example, this should be the final piece for now. I didn't know what would be the best place to add the documentation so it is in `doc/lvmvdo.rst` for now, feel free to recommend a better location.

dwlehman:
Is it possible to add a test to verify the behavior if the lvm/vdo tools/plugins are absent?

blivet/blivet.py
```diff
@@ -637,10 +653,10 @@ def new_lv(self, *args, **kwargs):
         if "%s-%s" % (vg.name, name) in self.names:
             raise ValueError("name already in use")
 
-        if thin_pool or thin_volume:
+        if thin_pool or thin_volume or vdo_pool:
```
dwlehman, line 656: Does this need to also check `vdo_lv`?

blivet/devicefactory.py
```diff
@@ -69,6 +70,9 @@ def is_supported_device_type(device_type):
         :returns: True if this device type is supported
         :rtype: bool
     """
+    if device_type == DEVICE_TYPE_LVM_VDO:
+        return not any(e for e in LVMVDOPoolMixin._external_dependencies if not e.available)
```
dwlehman, line 74: Why not just set `devices` to `LVMVDOPoolMixin` (and maybe also `LVMLogicalVolumeDevice`) below?

blivet/devicefactory.py
```diff
@@ -1752,6 +1759,92 @@ def _get_new_device(self, *args, **kwargs):
...
 
 
+class LVMVDOFactory(LVMFactory):
+
+    """ Factory for creating LVM VDO volumes.
+    """
+
+    def __init__(self, storage, **kwargs):
+        self.pool_name = kwargs.pop("pool_name", None)
+        self.virtual_size = kwargs.pop("virtual_size", None)
```
dwlehman, line 1769: Might be nice to (re)document here what happens if `pool_name` and/or `virtual_size` are omitted.

blivet/devicefactory.py
```diff
@@ -1752,6 +1759,92 @@ def _get_new_device(self, *args, **kwargs):
...
+    def _set_name(self):
+        super(LVMVDOFactory, self)._set_name()
+        if self.pool_name:
+            self._set_pool_name()
+
+    def _reconfigure_device(self):
+        super(LVMVDOFactory, self)._reconfigure_device()
+
+        self.device.pool.compression = self.compression
+        self.device.pool.deduplication = self.deduplication
```
dwlehman, line 1817: No `virtual_size` reconfiguration?

blivet/formats/fs.py
```diff
@@ -263,6 +264,14 @@ def label_format_ok(self, label):
     label = property(lambda s: s._get_label(), lambda s, l: s._set_label(l),
                      doc="this filesystem's label")
 
+    def can_nodiscard(self):
+        """Returns True if this filesystem supports nodiscrd option during
```
dwlehman, line 268: suspected typo: nodiscrd->nodiscard

dwlehman (approved):
Looks good to me. I had a few questions, but none of them is a blocker to approval.

blivet/formats/fs.py
vojtechtrefny: Fixed.

blivet/blivet.py
vojtechtrefny: Good catch, fixed.

blivet/devicefactory.py
vojtechtrefny: I've tried to do that and didn't manage to make it work. With the `Mixin` classes not being subclasses of `Device` and `LVMLogicalVolumeDevice` inheriting from all the `Mixin` classes, it made Python really unhappy. I'm not saying it's impossible, but I gave up and did this instead :-)

blivet/devicefactory.py
vojtechtrefny: I added the LVM VDO specific keyword arguments to the docstring.

blivet/devicefactory.py
vojtechtrefny: That's covered in `_set_device_size` which is called from the generic `_reconfigure_device`.

blivet/devices/lvm.py
```diff
@@ -1794,7 +1794,14 @@ def populate_ksdata(self, data):
 
 
 class LVMVDOPoolMixin(object):
```
dwlehman, line 1796: Does this class need a special `min_size`?

vojtechtrefny: I knew I forgot something important. I added a new commit setting `min_size` for vdo pools.

blivet/devicefactory.py
```diff
@@ -277,6 +277,7 @@ class DeviceFactory(object):
                          "container_size": SIZE_POLICY_AUTO,
                          "container_raid_level": None,
                          "container_encrypted": None}
+    _device_min_size = Size(0)  # no limit by default, limited only by filesystem size
```
vojtechtrefny, line 280: I added this for tests. I'm not sure if we want to make this "public" for all devices/factories, Anaconda might need a new API for getting device size limits (only limits we have now are based on filesystem min size).

doc/lvmvdo.rst
```diff
@@ -0,0 +1,86 @@
...
+        LVMLogicalVolumeDevice instance (0x7f14d17422b0) --
+            name = data-00  status = False  id = 528
+            children = []
+            parents = ['non-existent 5 GiB lvmvdopool data-myvdopool (519)']
+            ...
+
+``size`` in this case sets the pool (physical) size, the VDO logical volume size can be specified
+with ``virtual_size`` (if not specified it will be same as the pool size). Name for the VDO volume
+can be specified using the ``name`` keyword argument. ``pool_name`` argument is optional and
+a unique name will be generated if omitted. Both ``compression`` and ``deduplication`` default to
```
rhawalsh, line 82: Should compression and/or dedupe be false by default?  What would be the point of using LVMVDO if not for using one or both of those features?

vojtechtrefny: What is the common use case with VDO? Having both deduplication and compression enabled or just one of these? In the latter case it makes more sense to set these to `False` (disabled) by default. But if the most users use both deduplication and compression I'll change this to enabled by default.

rhawalsh: Typically both Compression and Deduplication are enabled by default and they are only disabled when there is a reason to do so.  Sometimes that can be resource-driven, other times that can be performance-driven.  Generally speaking, they should both be enabled by default.

vojtechtrefny: Ok, I've changed the default to enabled for both. Thanks.

vojtechtrefny:
> Is it possible to add a test to verify the behavior if the lvm/vdo tools/plugins are absent?

I've changed the availability/dependencies code for the LVM VDO devices and added new test case  for that.

dwlehman (approved):
Looks good.

PR #910, opened 2020-11-30
Use device ID instead of name for LUKS devices in device factory

vojtechtrefny:
Names cause issues with partitions -- PartitionFactory calls
do_partitioning after every factory (re)configure which can change
name of all non-existing partitions. When adding tow partitions
we can end up with two LUKS devices with same name because the
name of the underlying partition changed after adding the second
one but the LUKS device name didn't change. The LUKS devuce name
is never used, it's just our internal name so we can set it to
anything and device ID is unique and won't change.

Resolves: rhbz#1883557

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -745,7 +745,7 @@ class DeviceFactory(object):
             fmt = get_format(self.fstype,
                              mountpoint=self.mountpoint,
                              **fmt_args)
-            luks_device = LUKSDevice("luks-" + device.name,
+            luks_device = LUKSDevice("luks-%d" % device.id,
                                      parents=[device], fmt=fmt)
             self.storage.create_device(luks_device)
             ret = luks_device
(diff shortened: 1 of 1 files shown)


VladimirSlavik (approved):
Looks good to me. Thank you!

I wonder what other bugs will this uncover, from a bz-scoping perspective the LUKS stuff seemed pretty leaky so far.

dwlehman:
Just a thought: you could also update the children, if LUKS, in `PartitionDevice.update_name`. I thought of this when wondering if there is a risk that the id will change as well, but I don't know of any way that could happen off the top of my head. It's technically possible, though.

dwlehman:
I expected to find a bunch of other places we set the names of LUKS devices in a similar way. Are those all in anaconda now? I guess the only specific example I have in mind is autopart.

vojtechtrefny:
> Just a thought: you could also update the children, if LUKS, in `PartitionDevice.update_name`.

That's way better approach. And definitely safer. I've created new PR #917 and I'm closing this one.

PR #911, opened 2020-12-02
Make sure LV names are unique in device factory

vojtechtrefny:
This fixes issue that happens when trying to add a new LV to an existing VG. This is not possible to do directly, new VG is always created for the new LV so user needs to move it to the existing VG after that. The problem here is we recommend a name that is unique in the new VG (for example `root` for `/`) but it might not be unique in the existing VG so user would need to rename the LV before moving it. Technically this is correct behaviour -- Anaconda tells us to use name that is not unique (that we recommended for a different VG) but it's not really user friendly to recommend a name and tell user it isn't unique in the next step.
I also tried to clean the suggest name functions a little and reuse the code that adds `00` to the name to make it unique.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -1471,6 +1471,10 @@ class LVMFactory(DeviceFactory):
 
     def _get_new_device(self, *args, **kwargs):
         """ Create and return the factory device as a StorageDevice. """
+
+        name = kwargs.pop("name", "")
+        kwargs["name"] = self.storage.unique_device_name(name, parent=self.vg, name_set=True)
+
         return self.storage.new_lv(*args, **kwargs)
 
     def _set_name(self):
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -925,8 +925,23 @@ class Blivet(object):
 
         return tmp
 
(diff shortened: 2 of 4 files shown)


blivet/devicefactory.py
```diff
@@ -1471,6 +1471,10 @@ def _get_child_factory_kwargs(self):
 
     def _get_new_device(self, *args, **kwargs):
         """ Create and return the factory device as a StorageDevice. """
+
+        name = kwargs.pop("name", "")
+        kwargs["name"] = self.storage.unique_device_name(name, parent=self.vg, name_set=True)
+
```
dwlehman, line 1477: No change to `LVMFactory._set_name`?

dwlehman (approved):
Looks okay to me, with one question inline.

blivet/devicefactory.py
vojtechtrefny: The original name in `LVMFactory._set_name` is unique so it simply refuses to rename the device to the new non-unique name, which makes sense. This is a special case where the "original" name is not unique and there is no new name so I decided to make it unique only in this one special case as a exception.

PR #912, opened 2020-12-04
TFT is still broken so let's avoid failures by just doing a build

jkonecny12:
Testing Farm Team cluster does not work reliably. It should be fine during
January but until then let's avoid these constant tests failures.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -5,7 +5,7 @@ actions:
     - "make local"
     - 'bash -c "ls *.tar*"'
 jobs:
-- job: tests
+- job: copr_build
   metadata:
     targets:
     - fedora-all


vojtechtrefny (approved):
I must say I'm less and less enthusiastic about Packit. It's a great tool, but we already have mock based solution that works and we are replacing it with something that has less features and it's broken all the time*. But if this makes it better, I'm OK with that.

*) Based on past experience. And I don't really care if its TFT problem, from my pow packit just doesn't work correctly.

jkonecny12:
For enabling ELN tests back this have to be resolved first: https://pagure.io/fedora-ci/general/issue/185

jkonecny12:
> I must say I'm less and less enthusiastic about Packit. It's a great tool, but we already have mock based solution that works and we are replacing it with something that has less features and it's broken all the time*. But if this makes it better, I'm OK with that.
> 
> *) Based on past experience. And I don't really care if its TFT problem, from my pow packit just doesn't work correctly.

This is exactly what I'm trying to fight. The point is that Packit have a plenty of useful features which work reliably but everything is somehow dragged to the same sack with tests which are the only thing which does not work right now. I was able to successfully move all our COPR nightly builds to Packit without a problem. That seems to me like a nice improvement, I don't need a VM with a cron/jenkins job anymore.

PR #915, opened 2020-12-09
WIP: Initial Stratis support

vojtechtrefny:
This adds just a basic support for Stratis devices. Information about stratis blokdevs, pools and filesystems is gathered using the Stratis DBus API. I've added two new devices (`StratisFilesystemDevice` and `StratisPoolDevice`) and a new format (`StratisBlockdev`) with just few stratis specific properties for now. This is definitely not finished, I'm posting it here just to get some feedback about the "design" decisions, naming etc.

Stratis devices in devicetree

```
existing 1024 MiB disk sdc (304) with existing stratis
  existing 2 GiB stratis_pool stratis_howto (313)
    existing 1024 GiB stratis_filesystem stratis_howto/fs_howto2 (317) with existing xfs filesystem
    existing 1024 GiB stratis_filesystem stratis_howto/fs_howto (327) with existing xfs filesystem
existing 1024 MiB disk sdb (338) with existing stratis
  existing 2 GiB stratis_pool stratis_howto (313)
    ...
```

Stratis filesystem device

```
StratisFilesystemDevice instance (0x7f6d3c2fe8b0) --
  name = stratis_howto/fs_howto  status = True  id = 327
  children = []
  parents = ['existing 2 GiB stratis_pool stratis_howto (313)']
  uuid = 2d3dd8a9c7fe4217a792a0ca0bae5d24  size = 1024 GiB
  format = existing xfs filesystem
  major = 0  minor = 0  exists = True  protected = False
  sysfs path = /sys/devices/virtual/block/dm-12
  target size = 1024 GiB  path = /stratis/stratis_howto/fs_howto
  format args = []  original_format = xfs
```

Stratis pool device

```
StratisPoolDevice instance (0x7f6d3c2fe6a0) --
  name = stratis_howto  status = True  id = 313
  children = ['existing 1024 GiB stratis_filesystem stratis_howto/fs_howto2 (317) with '
 'existing xfs filesystem',
 'existing 1024 GiB stratis_filesystem stratis_howto/fs_howto (327) with '
 'existing xfs filesystem']
  parents = ['existing 1024 MiB disk sdc (304) with existing stratis',
 'existing 1024 MiB disk sdb (338) with existing stratis']
  uuid = f9eb3e81d0c04f9fb28db24365e4f754  size = 2 GiB
  format = existing None
  major = 0  minor = 0  exists = True  protected = False
  sysfs path = 
  target size = 2 GiB  path = /stratis/stratis_howto
  format args = []  original_format = None
```

### DIFF
--- a/blivet/devicelibs/stratis.py
+++ b/blivet/devicelibs/stratis.py
@@ -1,0 +1,26 @@
+#
+# stratis.py
+# stratis functions
+#
+# Copyright (C) 2020  Red Hat, Inc.  All rights reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# Author(s): Vojtech Trefny <vtrefny@redhat.com>
+#
+
+from ..size import Size
+
+
+STRATIS_FS_SIZE = Size("1 TiB")
--- a/blivet/devicelibs/stratis.py
+++ b/blivet/devicelibs/stratis.py
@@ -20,7 +20,77 @@
 # Author(s): Vojtech Trefny <vtrefny@redhat.com>
 #
 
+import gi
+gi.require_version("GLib", "2.0")
+
+from gi.repository import GLib
+
+from ..errors import StratisError
 from ..size import Size
+from ..static_data import stratis_info
+from .. import safe_dbus
+
+
+STRATIS_SERVICE = "org.storage.stratis2"
+STRATIS_PATH = "/org/storage/stratis2"
+STRATIS_POOL_INTF = STRATIS_SERVICE + ".pool"
+STRATIS_FILESYSTEM_INTF = STRATIS_SERVICE + ".filesystem"
+STRATIS_BLOCKDEV_INTF = STRATIS_SERVICE + ".blockdev"
+STRATIS_PROPS_INTF = STRATIS_SERVICE + ".FetchProperties"
+STRATIS_MANAGER_INTF = STRATIS_SERVICE + ".Manager"
 
 
 STRATIS_FS_SIZE = Size("1 TiB")
+
+
+def remove_pool(pool_uuid):
+    if not safe_dbus.check_object_available(STRATIS_SERVICE, STRATIS_PATH):
+        raise StratisError("Stratis DBus service not available")
+
+    # repopulate the stratis info cache just to be sure all values are still valid
+    stratis_info.drop_cache()
+
+    if pool_uuid not in stratis_info.pools.keys():
+        raise StratisError("Stratis pool with UUID %s not found" % pool_uuid)
+
+    pool_info = stratis_info.pools[pool_uuid]
+
+    try:
+        (succ, _uuid), rc, err = safe_dbus.call_sync(STRATIS_SERVICE,
+                                                     STRATIS_PATH,
+                                                     STRATIS_MANAGER_INTF,
+                                                     "DestroyPool",
+                                                     GLib.Variant("(o)", (pool_info.object_path,)))
+    except safe_dbus.DBusCallError as e:
+        raise StratisError("Failed to remove stratis pool: %s" % str(e))
+    else:
+        if not succ:
+            raise StratisError("Failed to remove stratis pool: %s (%d)" % (err, rc))
+
+
+def remove_filesystem(pool_uuid, fs_uuid):
+    if not safe_dbus.check_object_available(STRATIS_SERVICE, STRATIS_PATH):
+        raise StratisError("Stratis DBus service not available")
+
+    # repopulate the stratis info cache just to be sure all values are still valid
+    stratis_info.drop_cache()
+
+    if pool_uuid not in stratis_info.pools.keys():
+        raise StratisError("Stratis pool with UUID %s not found" % pool_uuid)
+    if fs_uuid not in stratis_info.filesystems.keys():
+        raise StratisError("Stratis filesystem with UUID %s not found" % fs_uuid)
+
+    pool_info = stratis_info.pools[pool_uuid]
+    fs_info = stratis_info.filesystems[fs_uuid]
+
+    try:
+        (succ, _uuid), rc, err = safe_dbus.call_sync(STRATIS_SERVICE,
+                                                     pool_info.object_path,
+                                                     STRATIS_POOL_INTF,
+                                                     "DestroyFilesystems",
+                                                     GLib.Variant("(ao)", ([fs_info.object_path],)))
+    except safe_dbus.DBusCallError as e:
+        raise StratisError("Failed to remove stratis filesystem: %s" % str(e))
+    else:
+        if not succ:
+            raise StratisError("Failed to remove stratis filesystem: %s (%d)" % (err, rc))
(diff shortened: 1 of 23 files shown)


blivet/devices/stratis.py
```diff
@@ -0,0 +1,87 @@
...
+
+    _type = "stratis_filesystem"
+    _resizable = False
+    _packages = ["stratisd"]
+    _dev_dir = "/stratis"
+
+    def _get_name(self):
+        """ This device's name. """
+        if self.pool is not None:
+            return "%s/%s" % (self.pool.name, self._name)
```
vojtechtrefny, line 63: `name` is a unique identifier in blivet so we can't use just the filesystem name here. I decided to use `<pool>/<filesystem>` to make it similar to LVM where we use `<vg>-<lv>`. We could use the DM name which is unique but `name` is often displayed in the installer UI so we'd need to tell Anaconda to make exception for stratis and use `fsname` when displaying stratis devices.

blivet/devicelibs/stratis.py
```diff
@@ -0,0 +1,96 @@
...
+STRATIS_SERVICE = "org.storage.stratis2"
+STRATIS_PATH = "/org/storage/stratis2"
+STRATIS_POOL_INTF = STRATIS_SERVICE + ".pool"
+STRATIS_FILESYSTEM_INTF = STRATIS_SERVICE + ".filesystem"
+STRATIS_BLOCKDEV_INTF = STRATIS_SERVICE + ".blockdev"
+STRATIS_PROPS_INTF = STRATIS_SERVICE + ".FetchProperties"
+STRATIS_MANAGER_INTF = STRATIS_SERVICE + ".Manager"
+
+
+STRATIS_FS_SIZE = Size("1 TiB")
```
vojtechtrefny, line 43: Can we expect this will be always 1 TiB? We need the size just for the initial device scan, the size is later updated from sysfs.

blivet/devices/stratis.py
```diff
@@ -0,0 +1,87 @@
...
+from .. import devicelibs
+
+
+class StratisPoolDevice(StorageDevice):
+    """ A stratis pool device """
+
+    _type = "stratis_pool"
+    _resizable = False
+    _packages = ["stratisd"]
+    _dev_dir = "/stratis"
```
vojtechtrefny, line 38: I developed and tested this on F33, this needs to be changed to `/dev/stratis` for F34.

blivet/devicelibs/stratis.py
```diff
@@ -0,0 +1,96 @@
...
+from .. import safe_dbus
+
+
+STRATIS_SERVICE = "org.storage.stratis2"
+STRATIS_PATH = "/org/storage/stratis2"
+STRATIS_POOL_INTF = STRATIS_SERVICE + ".pool"
+STRATIS_FILESYSTEM_INTF = STRATIS_SERVICE + ".filesystem"
+STRATIS_BLOCKDEV_INTF = STRATIS_SERVICE + ".blockdev"
+STRATIS_PROPS_INTF = STRATIS_SERVICE + ".FetchProperties"
+STRATIS_MANAGER_INTF = STRATIS_SERVICE + ".Manager"
```
jbaublitz, line 40: A few notes about the interface design that we have. Because we try to be as backwards compatible as possible, these interface names should still be available in Stratis v2.x, but just a note that these are the oldest interfaces available. These will probably need to be updated as new Fedora releases include later versions of stratisd.

jbaublitz:
I'll do a more thorough look through after I've discussed with the team, but this overall looks like a really good starting point!

blivet/devicelibs/stratis.py
jbaublitz: Currently, you can expect this to always start as 1 TiB. It will grow if the physical storage available is greater than 1 TiB and the filesystem fills up. Stratis automatically extends the filesystem size when it fills up. Currently, resizing will double the filesystem size.

blivet/devices/stratis.py
jbaublitz: I think it makes a lot of sense to use the simpler format that we use for our symlinks as you suggest here as that's what's going to be display to a user at the command line if they run `stratis filesystem list`.

blivet/devices/stratis.py
```diff
@@ -0,0 +1,87 @@
...
+from ..static_data import stratis_info
+from ..size import Size
+from ..storage_log import log_method_call
+from .. import devicelibs
+
+
+class StratisPoolDevice(StorageDevice):
+    """ A stratis pool device """
+
+    _type = "stratis_pool"
```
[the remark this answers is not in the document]
jbaublitz: I have no objections to this. @mulkieran Do you have any thoughts here?

blivet/devicelibs/stratis.py
mulkieran: jbaublitz is correct. You'll generally want the most recent revision for the version of stratisd that is available.

blivet/formats/stratis.py
```diff
@@ -0,0 +1,81 @@
...
+class StratisBlockdev(DeviceFormat):
+    """ A Stratis block device """
+
+    _type = "stratis"
+    _name = N_("Stratis block device")
+    _udev_types = ["stratis"]
+    _formattable = False                 # can be formatted
+    _supported = True                    # is supported
+    _linux_native = True                 # for clearpart
+    _min_size = Size("1 MiB")            # FIXME
```
mulkieran, line 41: We require 1 GiB for our devices.

blivet/devices/stratis.py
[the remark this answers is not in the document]
mulkieran: It is fine w/ me.

mulkieran (commented):
Looks fine!

blivet/formats/stratis.py
vojtechtrefny: Thanks, I've updated the `_min_size` value.

blivet/devicelibs/stratis.py
vojtechtrefny: I've used these interfaces mostly because I didn't need any properties from the newer revisions, but I'll try to use the newest interfaces in the future.

blivet/devices/stratis.py
```diff
@@ -48,6 +48,10 @@ def _destroy(self):
         log_method_call(self, self.name, status=self.status)
         devicelibs.stratis.remove_pool(self.uuid)
 
+    def dracut_setup_args(self):
+        return set(["stratis.rootfs.pool_uuid=%s" % self.uuid] +
+                   ["stratis.rootfs.uuids_paths=/dev/disk/by-uuid/%s" % p.uuid for p in self.parents])
```
jbaublitz, line 53: One note here is that the dracut command line argument requires a UUID format with dashes (this is how the symlinks are generated by udev) but our D-Bus API exports UUIDs without dashes, only hex characters. We already have an issue out about this: 

stratis-storage/stratis-cli#698

For now, you may want to parse the UUID format returned from the D-Bus and ensure that the output is returned with dashes if that's possible.

cc @mulkieran

vojtechtrefny:
I've pushed an updated version which includes a special class for the XFS filesystem on Stratis -- it's basically XFS with `_create` and `_destroy` functions overridden to be no-op. I think it's way better solution that having the workaround to not remove/create filesystem on Stratis FS devices, it makes the low level API a little bit confusing but we are already doing something similar for Btrfs and no one should create the devices manually anyway, we have helper functions for that. The new `stratis_xfs` filesystem type looks weird, but it shouldn't cause any issues. Thoughts @dwlehman

vojtechtrefny:
I've pushed (I hope) last version of the encrypted pools support and this should be now ready for review. The only missing part (I know about) is the size limits/free space calculation for pools. Stratis team is working on a tool that will allow us to calculate these limits for non-existing pools (see https://github.com/stratis-storage/stratisd/pull/2513).

With the encrypted pools support, I've tried to hide as much of the internal structure as possible. For both locked and unlocked pools, the LUKS format and device are hidden and internally we pretend the LUKS device is actually a Stratis block device

Unlocked pool:
```
existing 10 GiB disk vdb (38) with existing msdos disklabel
  existing 10 GiB partition vdb1 (48) with existing stratis
    existing 10 GiB stratis_pool blivet (57)
```

Locked pool:
```
existing 10 GiB disk vdb (38) with existing msdos disklabel
  existing 10 GiB partition vdb1 (48) with existing stratis
```

This makes populator helpers for stratis and LUKS a little bit messy but I think it's worth it and makes the stratis encrypted devices easier to present to users.

I'm also working on blivet-gui support for Stratis (https://github.com/storaged-project/blivet-gui/pull/277) to test these changes and also to make sure the API makes sense from the user point of view.

PR #917, opened 2020-12-15
Update LUKS device name after parent partition name change

vojtechtrefny:
When adding two partitions we can end up with two LUKS devices
with the same name because the name of the underlying partition
changed after adding the second one but the LUKS device name
didn't change.

Resolves: rhbz#188355


dwlehman (approved):
Looks good to me.

PR #918, opened 2020-12-21
Replace pocketlint by a custom script

vojtechtrefny:
This is based on similar change Anaconda did recently, see
rhinstaller/anaconda#2289

Using pylint directly is much faster than pocketlint. And pocketlint
doesn't really add that much functionality, we currently only use
the option to skip false positive reports based on a regex and this
can be done with pylint too.

pocketlint:
real    1m55,183s
user    9m4,957s
sys     0m23,387s

pylint:
real    0m39,922s
user    4m38,094s
sys     0m4,393s

### DIFF
--- a/tests/pylint/runpylint.py
+++ b/tests/pylint/runpylint.py
@@ -1,52 +1,122 @@
 #!/usr/bin/python3
 
+import atexit
+import shutil
 import sys
+import tempfile
+import os
 
-from pocketlint import FalsePositive, PocketLintConfig, PocketLinter
+from os import path
+
+from censorship import CensorshipConfig, CensorshipLinter, FalsePositive
 
 
-class BlivetLintConfig(PocketLintConfig):
-
+class BlivetLintConfig(CensorshipConfig):
     def __init__(self):
-        PocketLintConfig.__init__(self)
+        super().__init__()
 
-        self.falsePositives = [FalsePositive(r"Catching an exception which doesn't inherit from (BaseException|Exception): (BlockDev|DM|Crypto|Swap|LVM|Btrfs|MDRaid|Utils|G)Error$"),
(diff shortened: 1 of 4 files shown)


tests/pylint/runpylint.py
```diff
@@ -1,52 +1,122 @@
...
-                               FalsePositive(r"Method 'do_task' is abstract in class 'Task' but is not overridden"),
-                               FalsePositive(r"Method 'do_task' is abstract in class 'UnimplementedTask' but is not overridden"),
-                               FalsePositive(r"No value for argument 'member_count' in unbound method call$"),
-                               FalsePositive(r"No value for argument 'smallest_member_size' in unbound method call$"),
-                               FalsePositive(r"Parameters differ from overridden 'do_task' method$"),
-                               FalsePositive(r"Bad option value '(subprocess-popen-preexec-fn|try-except-raise|environment-modify)'"),
-                               FalsePositive(r"Instance of '(Action.*Device|Action.*Format|Action.*Member|Device|DeviceAction|DeviceFormat|Event|ObjectID|PartitionDevice|StorageDevice|BTRFS.*Device|LoopDevice)' has no 'id' member$"),
-                               FalsePositive(r"Instance of 'GError' has no 'message' member")  # overriding currently broken local pylint disable
-                               ]
+        super().__init__()
```
dwlehman, line 16: I think you could just omit `__init__` if all it does it call `super().__init()`.

dwlehman (approved):
Looks good to me. I'm a little bit curious why this is so much faster, but I like it.

vojtechtrefny:
> I'm a little bit curious why this is so much faster, but I like it.

I guess pylint is just better in multiprocessing now. Pocketlint just spawns multiple pylint processes using subprocess which is slower. (But pylint used to be worse few years ago, that's why we created pocketlint.)

PR #921, opened 2021-01-05
Fix possible UnicodeDecodeError when reading model from sysfs

vojtechtrefny:
I unfortunately fixed only one place where we read device model in #861, we also do it from `DiskDevicePopulator`. Lets just ignore (replace) all weird unicode characters we get when reading from sysfs.

Resolves: rhbz#1903793
Related: rhbz#1849326


VladimirSlavik (approved):
Looks good to me, as far as I can tell...

dwlehman (approved):
I like it.

PR #922, opened 2021-01-12
Fix "suggest_container_name" for Anaconda

vojtechtrefny:
We removed the "_get_container_name_template" method in
https://github.com/storaged-project/blivet/pull/911 but Anaconda
has its own version of it so we need to keep it in our code so
Anaconda can override it with their specific version in
InstallerStorage.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -943,6 +943,9 @@ class Blivet(object):
 
         raise RuntimeError("unable to find suitable device name")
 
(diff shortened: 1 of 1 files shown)


jkonecny12 (approved):
Thanks for this fix!

poncovka (commented):
Looks good to me. Thanks!

dwlehman (approved):

PR #923, opened 2021-01-18
Make sure we use size >= LVM VDO min size in test_lv_unique_name

vojtechtrefny:
The test was added to 3.3-devel before merging the LVM VDO code so we need to adjust it to work with LVM VDO.

### DIFF
--- a/tests/devicefactory_test.py
+++ b/tests/devicefactory_test.py
@@ -563,7 +563,7 @@ class LVMFactoryTestCase(DeviceFactoryTestCase):
     def test_lv_unique_name(self, *args):  # pylint: disable=unused-argument,arguments-differ
         device_type = self.device_type
         kwargs = {"disks": self.b.disks,
-                  "size": Size("400 MiB"),
+                  "size": max(Size("500 MiB"), self.factory_class._device_min_size),
                   "fstype": 'ext4',
                   "mountpoint": "/factorytest",
                   "device_name": "name"}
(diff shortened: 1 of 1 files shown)


tests/devicefactory_test.py
```diff
@@ -746,6 +746,16 @@ def test_get_free_disk_space(self, *args):
...
 
+    @patch("blivet.formats.lvmpv.LVMPhysicalVolume.formattable", return_value=True)
+    @patch("blivet.formats.lvmpv.LVMPhysicalVolume.destroyable", return_value=True)
+    @patch("blivet.static_data.lvm_info.blockdev.lvm.lvs", return_value=[])
+    @patch("blivet.devices.lvm.LVMVolumeGroupDevice.type_external_dependencies", return_value=set())
+    @patch("blivet.devices.lvm.LVMLogicalVolumeBase.type_external_dependencies", return_value=set())
+    @patch("blivet.devices.lvm.LVMVDOPoolMixin.type_external_dependencies", return_value=set())
+    @patch("blivet.devices.lvm.LVMVDOLogicalVolumeMixin.type_external_dependencies", return_value=set())
+    def test_lv_unique_name(self, *args):  # pylint: disable=unused-argument,arguments-differ
+        super(LVMVDOFactoryTestCase, self).test_normalize_size()
```
vpodzime, line 757: I guess this should be `super(...).test_lv_unique_name()`, right?

vpodzime (approved):
Looks good to me otherwise.

tests/devicefactory_test.py
vojtechtrefny: Oops, fixed.

PR #925, opened 2021-01-28
Use real paths to Python site packages

poncovka:
Anaconda is not going to support Python site packages in `/run/install/updates`.

**Related to:** https://github.com/rhinstaller/anaconda/pull/3127 (but it doesn't depend on it)

### DIFF
--- a/scripts/makeupdates
+++ b/scripts/makeupdates
@@ -85,7 +85,7 @@ def doGitDiff(tag, args=None):
 
     return out.split('\n')
 
-def copyUpdatedFiles(tag, updates):
+def copyUpdatedFiles(tag, updates, python_version):
     def install_to_dir(fname, relpath):
         sys.stdout.write("Including %s\n" % fname)
         outdir = os.path.join(updates, relpath)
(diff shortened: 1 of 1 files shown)


blivetghoul:

scripts/makeupdates
```diff
@@ -27,6 +27,9 @@ import six
 import subprocess
 import sys
 
+# The Python site-packages path for blivet.
+PYTHON_LIB_PATH = "./usr/lib/python3.9/site-packages/"
```
poncovka, line 31: I am not sure how to handle different versions of Python in your case. Anaconda will change the version per branch.

vojtechtrefny, line 31: I don't like this. This is forcing us to update this path for every Fedora release (don't forget we don't tie our releases to Fedora releases) and also makes the updates images version specific -- and not Fedora version specific but Python version specific which changes **during** rawhide development -- most extreme example could be updates image working with latest rawhide release and not the one from day before.

vojtechtrefny: If we really must have "version specific" updates images, it would make more sense (for us) to always build RPMs here for a specific Fedora version in mock and create updates image from them.

vojtechtrefny (approved):
We discussed this with @poncovka yesterday and agreed on a temporary solution -- we'll need to manually specify python version when creating updates images using our `makeupdates` script for now and in the future installer team will provide a project-independent updates image script that we'll use instead our "custom" one and it will automatically detect python version based on target Fedora/RHEL version.

I still don't like this change, but I understand why installer team wants to do this and it shouldn't affect our workflow much.

PR #929, opened 2021-02-01
Allow running blivet without Python SELinux module

vojtechtrefny:


blivet/util.py
```diff
@@ -450,6 +456,8 @@ def get_cow_sysfs_path(dev_path, dev_sysfsPath):
 
 def match_path_context(path):
```
vojtechtrefny, line 457: All calls to these function are "protected" by checking the SELinux flag first, so we should never raise the `RuntimeError` below, but these are still "public" functions so it is probably good idea to add the check just to be safe.

blivet/flags.py
```diff
@@ -40,7 +39,13 @@ def __init__(self):
         #
         # enable/disable functionality
         #
-        self.selinux = selinux.is_selinux_enabled()
+        try:
+            import selinux
+        except ImportError:
+            self.selinux = False
```
vojtechtrefny, line 45: We might want to log this for situations where SELinux is enabled but `python3-libselinux` is not installed. Thoughts?

japokorn (approved):
Looks good to me.

dwlehman (commented):
Looks good to me, but I wonder if it would be more consistent to use `tasks`/`availability` for that part of it.

PR #931, opened 2021-02-09
[btrfs] add new `FSMount` subclass for BTRFS (#1926892)

michel-slm:
We need to use this to allow overriding the filesystem mount options
from the default. Ideally this can be overridden in configuration,
but short of having such a mechanism we can hardcode it for now.

Context: https://fedoraproject.org/wiki/Changes/BtrfsTransparentCompression

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

---
This change is [Reviewable](https://reviewable.io/reviews/storaged-project/blivet/931)


blivetghoul:

michel-slm:
Not doing this in blivet (see https://github.com/storaged-project/blivet/pull/930)

PR #933, opened 2021-02-19
Set both boot and ESP flag for EFIFS partitions (#1930486)

vojtechtrefny:
Parted automatically sets the ESP flag together with the boot flag
on GPT but not on MBR which means a wrong partition ID will be set
on MBR (FAT specific ID 0x6 instead of 0xef for EFI system
partition) because partition IDs are based on flags with parted.
To make sure the correct ID is set we need to set both flags
manually on both GPT and MBR.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -28,7 +28,13 @@ import tempfile
 import uuid as uuid_mod
 import random
 
-from parted import fileSystemType, PARTITION_BOOT
+from parted import fileSystemType
+
+try:
+    from parted import PARTITION_ESP
+except ImportError:
+    # this flag is sometimes not available in pyparted, see https://github.com/dcantrell/pyparted/issues/80
+    PARTITION_ESP = 18
 
 from ..tasks import fsck
 from ..tasks import fsinfo
@@ -941,7 +947,7 @@ class EFIFS(FATFS):
     _min_size = Size("50 MiB")
     _check = True
     _mount_class = fsmount.EFIFSMount
-    parted_flag = PARTITION_BOOT
+    parted_flag = PARTITION_ESP
 
     @property
     def supported(self):
--- a/python-blivet.spec
+++ b/python-blivet.spec
@@ -33,7 +33,7 @@ Source1: http://github.com/storaged-project/blivet/archive/%{realname}-%{realver
 
 # Versions of required components (done so we make sure the buildrequires
 # match the requires versions of things).
-%global partedver 1.8.1
+%global partedver 3.2
 %global pypartedver 3.10.4
 %global utillinuxver 2.15.1
 %global libblockdevver 2.24


bcl:
You could just switch it to use PARTITION_ESP, it  works the same on MBR and GPT and doesn't require converting everything to a list.

vojtechtrefny:
> You could just switch it to use PARTITION_ESP, it works the same on MBR and GPT and doesn't require converting everything to a list.

So we don't need the boot flag at all? If I use only `PARTITION_ESP` the boot flag is not set on MBR.

bcl:
> So we don't need the boot flag at all? If I use only `PARTITION_ESP` the boot flag is not set on MBR.

According to the UEFI spec only the partition type should matter, the boot flag isn't used by UEFI firmware. Table 18 of the spec says

"0x80 indicates that this is the bootable legacy partition. Other values indicate that this is not a bootable legacy partition. This field shall not be used by UEFI firmware."

If we set both it may actually be possible to confuse things if the system expects to have both a bootable partition (for bios) and an ESP partition. So it's probably better to err on the side of the UEFI spec an only set the ESP flag.

FWIW I've never seen a UEFI system using MBR, I'm not sure who would want to do that.

vojtechtrefny:
> > So we don't need the boot flag at all? If I use only `PARTITION_ESP` the boot flag is not set on MBR.
> 
> According to the UEFI spec only the partition type should matter, the boot flag isn't used by UEFI firmware. Table 18 of the spec says
> 
> "0x80 indicates that this is the bootable legacy partition. Other values indicate that this is not a bootable legacy partition. This field shall not be used by UEFI firmware."
> 
> If we set both it may actually be possible to confuse things if the system expects to have both a bootable partition (for bios) and an ESP partition. So it's probably better to err on the side of the UEFI spec an only set the ESP flag.
> 

I've updated the PR, thanks for help!

> FWIW I've never seen a UEFI system using MBR, I'm not sure who would want to do that.

I was also surprised to see EFI on MBR but it's apparently a valid use case.

blivet/formats/fs.py
```diff
@@ -941,7 +947,7 @@ class EFIFS(FATFS):
     _min_size = Size("50 MiB")
     _check = True
     _mount_class = fsmount.EFIFSMount
-    parted_flag = PARTITION_BOOT
+    parted_flags = PARTITION_ESP
```
japokorn, line 950: Is there a reason to change name of the variable from `parted_flag` to `parted_flags`?

blivet/formats/fs.py
```diff
@@ -30,6 +30,12 @@
 
 from parted import fileSystemType, PARTITION_BOOT
```
japokorn, line 31: It looks like `PARTITION_BOOT` is no longer used after this change.

blivet/formats/fs.py
vojtechtrefny: Fixed. That's from the first iteration where I tried to set both `PARTITION_BOOT` and `PARTITION_ESP`.

dwlehman (approved):
Looks good to me. Hopefully there won't be any fallout.

PR #935, opened 2021-02-25
Adapt to dosfstools 4.2 FAT label changes

vojtechtrefny:
The default "NO NAME" label is now not printed by the "dosfslabel"
tool and a new option "--reset" must be used to set empty label.

### DIFF
--- a/blivet/tasks/fswritelabel.py
+++ b/blivet/tasks/fswritelabel.py
@@ -66,7 +66,13 @@ class DosFSWriteLabel(FSWriteLabel):
 
     @property
     def args(self):
-        return [self.fs.device, self.fs.label]
+        if availability.MKDOSFS_NEW_APP.available:
+            if self.fs.label:
+                return [self.fs.device, self.fs.label]
+            else:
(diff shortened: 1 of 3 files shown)


blivet/tasks/fswritelabel.py
```diff
@@ -66,7 +66,13 @@ class DosFSWriteLabel(FSWriteLabel):
...
     @property
     def args(self):
-        return [self.fs.device, self.fs.label]
+        if availability.MKDOSFS_NEW_APP.available:
+            if self.fs.label:
+                return [self.fs.device, self.fs.label]
+            else:
+                return [self.fs.device, "--reset"]
+        else:
+            return [self.fs.device, self.fs.label]
```
dwlehman, line 75: This code is a bit redundant, but it's also clear. It might be better IMO to only have one `return` and maybe just set the second argument in the conditional, defining the first argument either at the top or in the `return`. Your call.

dwlehman (approved):
Looks good to me. One minor nitpick about brevity, but it's just personal preference.

PR #936, opened 2021-03-01
Try to get Btrfs volume UUID using libblockdev if UDev lookup fails

vojtechtrefny:
Resolves: rhbz#1860602

-----

This isn't really a fix, just a workaround. @dwlehman mentioned similar issue happening in the [store role](https://github.com/linux-system-roles/storage) too, we should consider reading the UUID and similar properties we gather in `_post_create` from other sources than UDev. We run `udev.settle()` but it doesn't help.


dwlehman (approved):
Good idea.

PR #937, opened 2021-03-02
Allow removing LVM VDO devices without VDO support

vojtechtrefny:
We don't need VDO to be able to remove these.

Resolves: rhbz#1933611


dwlehman (approved):
Would be nice if this was handled with finer-grained deps, but this looks good as a simpler approach.

PR #938, opened 2021-03-02
apply directory's SELinux context to freshly created mount points

rmetrich:
When installing a system with non-standard mount points such as `/apps`, the mount point is created with `etc_runtime_t` label instead of `default_t`.

This is due to using the **0** parameter as mode in `matchpatchcon()` instead of explicitly specifying it's a directory.
Indeed, in the SELinux database, there are 2 rules for `/apps` and related:
~~~
/.*                                                all files          system_u:object_r:default_t:s0 
/[^/]+                                             regular file       system_u:object_r:etc_runtime_t:s0 
~~~

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -27,6 +27,7 @@ import os
 import tempfile
 import uuid as uuid_mod
 import random
+import stat
 
 from parted import fileSystemType, PARTITION_BOOT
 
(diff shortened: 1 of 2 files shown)


blivetghoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
Looks good to me, thank you.

PR #940, opened 2021-03-10
Add support for renaming devices using ActionConfigureDevice

vojtechtrefny:
Currently only volume groups and logical volumes can be renamed.

----

The internal implementation for LVs is a little messier because of the difference between `name` and `lvname`, but I think it makes more sense making the API for users simpler (changing `name` for all devices) even thought the implementation is a little bit uglier.

### DIFF
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -1139,7 +1139,10 @@ class ActionConfigureDevice(DeviceAction):
         self.old_value = getattr(self.device, self.attr)
 
         if self._execute:
-            self._execute(dry_run=True)
+            kwargs = {"old_%s" % self.attr: self.old_value,
+                      "new_%s" % self.attr: self.new_value,
+                      "dry_run": True}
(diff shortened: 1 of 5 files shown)


dwlehman (approved):
Nice. Looks good to me.

The only thing I might question is whether we should just quietly move along instead of raising an exception when the old and new names match.

vojtechtrefny:
> The only thing I might question is whether we should just quietly move along instead of raising an exception when the old and new names match.

I also wasn't sure about this. Both `lvrename` and `vgrename` will fail if you try to rename the device to the same name, but we can simply not call it in this case. I think it might be a good "warning" for the user to tell him he is doing something wrong, like renaming a wrong LV. But I'm not against changing it to not raise the exception if you think it makes more sense.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #941, opened 2021-03-10
Fix usage of assert_called_with in lvm_test

vojtechtrefny:
Found when working on tests for #940. mock.called_with is always
true, because it's just another custom mocked function, correct
function is assert_called_with.


dwlehman (approved):
Looks good. Thanks.

PR #942, opened 2021-03-10
spec: Add 'make' to BuildRequires

vojtechtrefny:
make is no longer present in the Fedora build environment by
default.

----

We already have this in dist-git and on 3.4-release (see https://github.com/storaged-project/blivet/commit/bb0dc95a59b8e69f83dbf07d0267b015154c68f6) but we also need make in spec here on the devel branch for packit builds.


dwlehman (approved):
Looks good.

PR #943, opened 2021-03-11
Do not log entire exception when trying to get ISCSI initiator name

vojtechtrefny:
We do this every time and in most cases the call will fail
because there is no ISCSI firmware. Having the entire exception
logged is not useful for us (it will always be the same backtrace
for a failed DBus call) and it confuses users when reporting bugs
because it's the first error message in the log.


dwlehman (approved):
I might use `log.info` instead, but otherwise a big YES. Thank you!

vojtechtrefny:
> I might use `log.info` instead, but otherwise a big YES. Thank you!

I picked `log.debug` because `log_exception_info` logs everything as debug, but you are right, `log.info` makes more sense here.

Issue #945, opened 2021-03-15
Outdated dependencies in Makefile


VladimirSlavik:
When running `make install-requires`, I can't get the following packages:
* `zanata-python-client`
* `python3-pep8`

PR #947, opened 2021-03-17
install-test-dependencies playbook update

vojtechtrefny:
It was really outdated. I've also added a simple Vagrantfile. We already have similar configs in our other projects ([libblockdev](https://github.com/storaged-project/libblockdev/tree/master/misc), [blivet-gui](https://github.com/storaged-project/blivet-gui/tree/master/misc)).

Fixes: #945


VladimirSlavik:
I think pocketlint is missing for centos?

vojtechtrefny:
> I think pocketlint is missing for centos?

Oops. Good catch, thanks.

misc/install-test-dependencies.yml
```diff
@@ -0,0 +1,163 @@
...
+      - python3-six
+      - python3-kickstart
+      - python3-pyudev
+      - python3-pyparted
+      - libselinux-python3
+      - python3-blockdev
+      - python3-bytesize
+      - python3-libvirt
+      - python3-paramiko
+      - python3-pip
```
dwlehman, line 105: You can pass `name` as a list, eg:

```
- package
    name:
       - one
       - two
       - three

```

which _should_ be substantially faster than the loop.

dwlehman (approved):
Looks okay to me; one suggestion, inline.

misc/install-test-dependencies.yml
vojtechtrefny: Fixed. Is this a new thing? I think I remember using the `with_items` loop because the `dnf` module didn't work with lists. Anyway, it works now and it is faster.

VladimirSlavik: I'd actually suggest formatting all package lists like that, if you're touching the lines anyway.

PR #949, opened 2021-04-06
Avoid AttributeError for DiskLabel formats without disklabel type

vojtechtrefny:
Simple instance of DiskLabel without label_type specified in the
constructor results in a AttributeError when trying to print the
format name. This makes sure the name is always valid.

Resolves: rhbz#1945914

------

Anaconda hits this when it's gathering format types for some DBus shenanigans. Simple reproducer is to get disklabel without specifying type `blivet.formats.get_format("disklabel")` and it will fail when printing the disklabel info (because `__repr__` prints `name`). Alternatively we could make sure the `_default_label_type` is never `None` but that might break more things. It might be also better to have a special `name` for this "disklabel without type" but I wasn't able to come up with something that makes sense, so feel free to suggest something better.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -292,8 +292,11 @@ class DiskLabel(DeviceFormat):
             if self._label_type:
                 lt = self._label_type
             else:
-                lt = self._get_best_label_type()
-
+                try:
+                    lt = self._get_best_label_type()
+                except Exception:  # pylint: disable=broad-except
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
I reworked this a bit. We have another bug, this time on s390 where we try to check whether the device is FBA DASD and the check fails if `device` is `None`.

japokorn (approved):
Looks good to me.

PR #950, opened 2021-04-12
Ignore ArithmeticError when trying to align partition size down

vojtechtrefny:
For 512 B partitions, pyparted throws ArithmeticError when trying
to align their size down which crashes the installer when trying
to get max size of such partitions.

Resolves: rhbz#1947985


japokorn (approved):
Looks good to me.

PR #951, opened 2021-04-15
Provide better error message for LVM with inconsistent sector sizes

vojtechtrefny:
List of pvs with sector sizes is now part of the error message. I've also added a new exception for Anaconda to catch so they can add more information in the UI (workarounds in kickstart etc.).

### DIFF
--- a/blivet/devices/lvm.py
+++ b/blivet/devices/lvm.py
@@ -25,7 +25,7 @@ import pprint
 import re
 import os
 import time
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 from functools import wraps
 from enum import Enum
 import six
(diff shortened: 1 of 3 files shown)


vojtechtrefny:
I've updated the error message, our documentation team recommended a small change, it now looks like this:

> The specified volume group cannot be created because inconsistent sector
> sizes have been detected on the following disks:
> vda5: 512
> vdg1: 4096

vojtechtrefny:
One more update based on feedback from documentators and installer team.

> Cannot create volume group 'testvg'. The following disks have inconsistent sector size:
pv1: 512
pv2: 4096

dwlehman (approved):
Looks good.

PR #958, opened 2021-05-20
Fix activating old style LVM snapshots

vojtechtrefny:
The old style snapshots are activated together with the origin LV
so we need to make sure it is activated to be able to remove the
snapshot or its format.

Resolves: rhbz#1961739


japokorn (approved):
Looks good to me.

PR #959, opened 2021-05-24
Fix adding/removing devices to the LVM reject filter

vojtechtrefny:
Because the filter is a list internally we can add a device multiple times to it and remove it only once when we decide to use the device.


japokorn (approved):
Looks good to me

PR #960, opened 2021-05-25
When sorting devices make sure partitions are sorted correctly

vojtechtrefny:
We need a special sorting key for partitions to make sure they are
sorted like "sda1, sda2, ..., sda10" and not like "sda1, sda10,
sda2, ...".

Resolves: rhbz#1915333

----

When removing a logical partition, the "higher number" partitions are automatically renumbered (e.g. with sda5 and sda6, sda6 becomes sda5 after removing sda5). We have a code to adjust to this behaviour (sorting destroy actions etc.) but it still causes issues for Anaconda when allowing to remove partitions in the UI where Anaconda uses the displayed name to remove the partition which no longer exists in the model. By correctly sorting the partitions we can make sure Anaconda removes them in correct order (see also https://github.com/rhinstaller/anaconda/pull/2755).

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -42,7 +42,7 @@ from .devicetree import DeviceTree
 from .formats import get_default_filesystem_type
 from .flags import flags
 from .formats import get_format
-from .util import capture_output
+from .util import capture_output, natural_sort_key
 from . import arch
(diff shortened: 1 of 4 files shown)


japokorn (approved):
Looks good to me

PR #963, opened 2021-06-10
Fix/unify importing mock module in tests

vojtechtrefny:
mock is available in the unittest module since Python 3.3, we need
to use the old mock module not only with Python 2 but also with
the early versions of Python 3.

----

We are currently using mix of our custom `six.moves` solution, `try-except` imports and `six.PY3` based check (and in some tests we simply `import mock` which means we have hard dependency on python-mock even with Python >= 3.3). This change unifies everything to the `try-except` solution which should work everywhere.


japokorn (approved):
Looks good to me.
Pylint issues are false positives addressed in #966.

PR #964, opened 2021-06-10
Fix util.virt_detect on Xen

vojtechtrefny:
Xen is apparently still alive so we should return True for it too.

-----

Discover by the new Fedora/RHEL tests that run in EC2. I guess we could also adjust the `MiscTest.test_detect_virt` test but I think it makes more sense to return `True` for Xen too.


japokorn (approved):
Looks good to me.
Pylint issues are false positives addressed in #966.

PR #966, opened 2021-06-24
Ignore pylint false positive no-member warning

vojtechtrefny:


vojtechtrefny:
The rawhide test failures are caused by a change in python 3.10, Size class is no longer consider to be integer, this needs to be fixed in libbytesize, see https://github.com/storaged-project/libbytesize/pull/96

vojtechtrefny:
Jenkins, test this please.

PR #967, opened 2021-07-19
Revert "Use PARTITION_ESP flag for EFIFS partitions (#1930486)"

vojtechtrefny:
This reverts commit aeffc4d50d5059e6a782ee41ccd49061d28996ab.

Using the correct 0xef partition type for EFI System Partition
unfortunately doesn't work with RaspberryPi so we need to keep
using 0x06 with "normal" boot flag to make it boot.

Resolves: rhbz#1975375

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -29,13 +29,7 @@ import uuid as uuid_mod
 import random
 import stat
 
-from parted import fileSystemType
-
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
Jenkins, test this please.

japokorn (approved):
Looks good to me

PR #968, opened 2021-07-22
Tell LVM to ignore the new devices file for now

vojtechtrefny:
We currently don't support working with the devices file and it's
not possible to use lvm.conf filters together with the devices
file so we need to tell LVM to ignore it until we can support it.


japokorn (approved):
Looks good to me

PR #969, opened 2021-07-29
Do not set chunk size for RAID 1

vojtechtrefny:
Setting chunk size for RAID 1 doesn't make sense and latest
mdadm started returning error instead of ignoring the --chunk
option when creating an array.

Resolves: rhbz#1987170


japokorn (approved):
Looks good to me

PR #971, opened 2021-08-04
Fix running tests against installed version of blivet

vojtechtrefny:


jkonecny12:
Nice seems that our tests works as they should :).

japokorn (approved):
Looks good to me. 
Packit issue is not connected to this change.

PR #972, opened 2021-08-05
LVM devices support

vojtechtrefny:
Work in progress version of the LVM devices file and "filtering" support (check [lvmdevices manpage](https://man7.org/linux/man-pages/man8/lvmdevices.8.html) for more information about this new LVM feature).

Because we need to support the "old" lvm.conf based filtering too (the new LVM devices feature is currently available only on CentOS/RHEL 9) the first commit switches from the existing *reject* based filtering to *accept* based. This is the biggest change in this PR, the LVM devices support itself is just about switching from calling LVM with `--config devices` to `--devices`. Libblockdev support needs to merged first, see https://github.com/storaged-project/libblockdev/pull/646

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -259,10 +259,9 @@ class ActionList(object):
         for action in self._actions:
             log.debug("action: %s", action)
 
-            # Remove lvm filters for devices we are operating on
-            lvm.lvm_cc_removeFilterRejectRegexp(action.device.name)
             for device in (d for d in devices if d.depends_on(action.device)):
-                lvm.lvm_cc_removeFilterRejectRegexp(device.name)
+                if device.format.type == "lvmpv":
+                    lvm.lvm_devices_add(device.path)
 
     def _post_process(self, devices=None):
         """ Clean up relics from action queue execution. """
--- a/blivet/devicelibs/lvm.py
+++ b/blivet/devicelibs/lvm.py
@@ -65,6 +65,16 @@ LVMETAD_SOCKET_PATH = "/run/lvm/lvmetad.socket"
 
 safe_name_characters = "0-9a-zA-Z._-"
 
+if hasattr(blockdev.LVMTech, "DEVICES"):
+    try:
+        blockdev.lvm.is_tech_avail(blockdev.LVMTech.DEVICES, 0)  # pylint: disable=no-member
+    except blockdev.LVMError:
+        HAVE_LVMDEVICES = False
+    else:
+        HAVE_LVMDEVICES = True
+else:
+    HAVE_LVMDEVICES = False
+
 # list of devices that LVM is allowed to use
 # with LVM >= 2.0.13 we'll use this for the --devices option and when creating
 # the /etc/lvm/devices/system.devices file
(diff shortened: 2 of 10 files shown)


blivet/actionlist.py
```diff
@@ -259,10 +259,9 @@ def _pre_process(self, devices=None):
         for action in self._actions:
             log.debug("action: %s", action)
 
-            # Remove lvm filters for devices we are operating on
-            lvm.lvm_cc_removeFilterRejectRegexp(action.device.name)
             for device in (d for d in devices if d.depends_on(action.device)):
-                lvm.lvm_cc_removeFilterRejectRegexp(device.name)
```
dwlehman, line 265: It will be wonderful to see this code go away. It made so much noise in the logs, but I never took the time to quiet it down.

blivet/devicelibs/lvm.py
```diff
@@ -65,78 +65,90 @@
...
 
 
 @needs_config_refresh
-def lvm_cc_addFilterRejectRegexp(regexp):
-    """ Add a regular expression to the --config string."""
-    log.debug("lvm filter: adding %s to the reject list", regexp)
-    config_args_data["filterRejects"].add(regexp)
+def lvm_devices_add(path):
+    """ Add a device (PV) to the list of devices LVM is allowed to use """
+    log.debug("lvm filter: device %s added to the list of allowed devices")
```
dwlehman, line 134: Oh, hmm, I see the noise level in the logs will remain the same.

blivet/devicelibs/lvm.py
```diff
@@ -65,78 +65,90 @@
...
-    log.debug("lvm filter: removing %s from the reject list", regexp)
+def lvm_devices_remove(path):
+    """ Remove a device (PV) to the list of devices LVM is allowed to use """
+    log.debug("lvm filter: device %s removed from the list of allowed devices")
     try:
-        config_args_data["filterRejects"].remove(regexp)
+        _lvm_devices.remove(path)
     except KeyError:
-        log.debug("%s wasn't in the reject list", regexp)
+        log.debug("%s wasn't in the devices list", path)
```
dwlehman, line 145: Another (quieter) approach would be to check for `path` and only emit the log message if it is indeed being added/removed as requested -- IOW, if the call is a no-op nothing would be logged. Maybe that's less clear for debugging the devices list, but less noise (at least the way we use the old/current filter) in general.

dwlehman (approved):
Looks like an overall improvement to me.

blivet/devicelibs/lvm.py
vojtechtrefny: The excessive was helpful when debugging some of the filtering issue. Ideally there should be zero no-op calls here. We could change this in the future when (if) we decide to drop support for the "old style" lvm.conf filtering and the device file based filtering is mature enough.

vojtechtrefny:
I've added new commit that adds to functions to `LVMPhysicalVolume` to add and remove the PV from the LVM system.devices file. Anaconda will use these to add existing PVs used during the installation to the config file. It will be easier for them than using libblockdev directly.

PR #977, opened 2021-08-16
pylint: Ignore deprecation warning about threading.currentThread

vojtechtrefny:
We use it only with Python 2, Python 3 branch uses current_thread.


japokorn (approved):
Looks good to me

PR #980, opened 2021-08-24
Do not use FS.mount for btrfs temporary mounts

vojtechtrefny:
The FS.mount/setup does additional things, notably setting SELinux
context when running in the installer mode, which we don't want to
do when mounting the btrfs volume just to get list of subvolumes.


japokorn (approved):
Looks good to me.

PR #981, opened 2021-09-01
Distutils deprecation

vojtechtrefny:
We have some hacks in `setup.py` so it's possible this breaks *something* but everything I was able to manually test seems to be working. I guess we won't be sure until next upstream release.


japokorn (approved):
Looks good to me.

Issue #982, opened 2021-09-05
unable to launch on linux mint 20.2 (OBS repository/manual install, ubuntu 20.04 base)


buck2202:
I've installed `blivet-gui` from OBS on a very fresh install of linux mint 20 (ubuntu 20.04 base). Trying to launch from the system menu fails silently; launching from a terminal, I get this
```
$ sudo blivet-gui
Traceback (most recent call last):
  File "/usr/bin/blivet-gui", line 40, in <module>
    from blivetgui.dialogs.message_dialogs import ErrorDialog
  File "/usr/lib/python3/dist-packages/blivetgui/dialogs/message_dialogs.py", line 31, in <module>
    from .helpers import adjust_scrolled_size
  File "/usr/lib/python3/dist-packages/blivetgui/dialogs/helpers.py", line 34, in <module>
    from blivet import devicefactory
  File "/usr/lib/python3/dist-packages/blivet/devicefactory.py", line 27, in <module>
    from .devices import BTRFSDevice, DiskDevice
  File "/usr/lib/python3/dist-packages/blivet/devices/__init__.py", line 24, in <module>
    from .storage import StorageDevice
  File "/usr/lib/python3/dist-packages/blivet/devices/storage.py", line 33, in <module>
    from ..formats import get_format, DeviceFormat
  File "/usr/lib/python3/dist-packages/blivet/formats/__init__.py", line 46, in <module>
    from ..tasks import fsinfo
  File "/usr/lib/python3/dist-packages/blivet/tasks/fsinfo.py", line 29, in <module>
    from . import availability
  File "/usr/lib/python3/dist-packages/blivet/tasks/availability.py", line 377, in <module>
    technologies={blockdev.LVMTech.VDO: (blockdev.LVMTechMode.CREATE |
AttributeError: type object 'LVMTech' has no attribute 'VDO'
```

The trace seems to suggest that I should've gotten an error message--does this suggest a dependency issue, and/or should the `AttributeError` have been fatal? I've never found an ubuntu repo that _didn't_ work on mint, but I guess it's possible that the packages installed by default are different and something isn't getting pulled in explicitly?

buck2202:
thanks for your help!

vojtechtrefny:
It took me a little longer, sorry for that, but the fixed version of blivet (python3-blivet 3.4.1-2) is now available in the OBS repository. I'll keep this issue open and let GitHub close it after we merge #983.

buck2202:
[...]

PR #983, opened 2021-09-05
Fix checking for LVM VDO support with libblockdev 2.23 and older

vojtechtrefny:
Fixes: #982

This will make possible to use latest blivet (and blivet-gui) on Ubuntu 20.04 which doesn't have libblockdev 2.24.


japokorn (approved):
Looks good to me.

PR #985, opened 2021-09-07
Do not add device name as a parameter for errors.DeviceError

vojtechtrefny:
DeviceError doesn't expect additional arguments which leads to the
exception message being tuple of (msg, name) instead of a simple
string.

------

I've tried to find out why the name is added for some `DeviceError`s, but as far as I can tell it never expected device name as a second parameter. We have some exceptions like that (notably `UnusableConfigurationError`, but not `DeviceError`). I added the name to the exception string where it makes sense and for most calls it's also logged by `log_method_call` so we shouldn't lose any information. Feel free to suggest better messages for the exceptions.

### DIFF
--- a/blivet/devices/container.py
+++ b/blivet/devices/container.py
@@ -146,7 +146,7 @@ class ContainerDevice(StorageDevice):
             This method writes the member addition to disk.
         """
         if not self.exists:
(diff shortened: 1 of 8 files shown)


japokorn (approved):
Looks good to me.

PR #986, opened 2021-09-07
Remove unused flag multipath

vojtechtrefny:
We are not using the flag internally and Anaconda is not setting it, see also https://github.com/rhinstaller/anaconda/pull/3586#issuecomment-910462263


japokorn (approved):
Looks good to me.

Issue #944, opened 2021-03-15
Replace uses of IOError


VladimirSlavik:
`IOError` has been deprecated a long time ago, it is now an alias for `OSError`. Blivet generally contains two kinds uses:

1. Guarding against failures of file access (`open`, `read` etc.). These can be just replaced with `OSError`.
2. It raises it in several places. These need replacing with something else.

See also https://github.com/rhinstaller/anaconda/pull/3239

vojtechtrefny:
Closing the issue, both #946 and #990 are now merged and there are no more `IOError`s in blivet.

PR #990, opened 2021-09-14
Do not raise deprecated IOError from iscsi and fcoe modules

vojtechtrefny:
IOError is now deprecated and only an alias to OSError. Raising
IOError doesn't really makes sense for our use cases so we are
replacing it with custom FCoE/ISCSIError exceptions that are also
subclasses of OSError to keep backward compatibility.

Fixes: #944

### DIFF
--- a/blivet/errors.py
+++ b/blivet/errors.py
@@ -307,3 +307,13 @@ class EventHandlingError(StorageError):
 
 class ThreadError(StorageError):
     """ An error occurred in a non-main thread. """
+
+# other
+
+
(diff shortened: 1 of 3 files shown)


japokorn (approved):
Looks good to me.

vojtechtrefny:
@VladimirSlavik can you check this change? It would be nice if Anaconda switched to using the new exception but catching `IOError` or `OSError` should work too.

VladimirSlavik (approved):
Looks good to me. Thank you!

PR #991, opened 2021-09-14
Set correct map name for existing LUKS devices

vojtechtrefny:
We assume that all LUKS device are named luks-<UUID>. This is true
for devices activated by us or UDisks, but manually opened devices
can have a different name. This change uses sysfs holders to check
the "real" name for opened LUKS devices and makes sure we always
set correct state for the LUKS format based on the name (we check
whether /dev/mapper/<map_name> exists when checking whether the
LUKS format is active or not).


japokorn (approved):
Looks good to me.

PR #997, opened 2021-10-28
Fix removing zFCP SCSI devices

jstodola:
Values parsed from `/proc/scsi/scsi` were not correctly used to assemble
paths to SCSI devices.

For example:
`/sys/bus/scsi/devices/0:0:00:00/`
was incorrectly accessed instead of:
`/sys/bus/scsi/devices/0:0:0:0/`

Switch to a more reliable way of listing the available SCSI devices.

### DIFF
--- a/blivet/zfcp.py
+++ b/blivet/zfcp.py
@@ -20,6 +20,7 @@
 #
 
 import os
+import re
 from . import udev
 from . import util
 from .i18n import _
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
Thank you, looks good to me.

PR #998, opened 2021-11-10
Use bigger chunk size for thinpools bigger than ~15.88 TiB

vojtechtrefny:
With our default chunk size of 64 KiB we cannot create bigger
thin pools than 15.88 TiB. Unfortunately we need to specify chunk
size to be able to calculate thin metadata properly so we can't
simply leave this to LVM to determine the correct chunk size.


japokorn (approved):
Looks good to me.

PR #999, opened 2021-12-01
iscsi: Replace all log_exception_info calls with log.info

vojtechtrefny:
We don't get any useful information from the exception, it's
always the same traceback from a failed DBus call and we only use
these when a called failed because firmware ISCSI is not supported.
The resulting log message also looks like a failure with the
traceback logged and not just as a debug information.

Resolves: rhbz#2028134

----

Similar to #943, we had the `log_exception_info` function on two other places that are not triggered so often and as visible as the previous one, but it's still confusing when we log the entire traceback for something that isn't critical.

### DIFF
--- a/blivet/iscsi.py
+++ b/blivet/iscsi.py
@@ -22,7 +22,6 @@ from . import udev
 from . import util
 from .flags import flags
 from .i18n import _
-from .storage_log import log_exception_info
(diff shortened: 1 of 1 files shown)


japokorn (approved):
Looks good to me.

Issue #1003, opened 2021-12-22
Separate actions for LVM cache attach/detach


vojtechtrefny:
Currently both `LVMLogicalVolumeDevice.attach_cache` and `LVMCache.detach` are simple functions that immediately call the libblockdev functions to attach or detach the cache volume. This isn't currently a problem because we are not supporting creating cache volumes and only way to create a cached LV is to do create a new LV with a cache in a one step using `cache_request`. I'm currently working on support for creating cache pools and attaching them to an existing LV (this will be used in the storage role, see https://github.com/linux-system-roles/storage/pull/235 and in the future could be useful for blivet-gui too). In this case having an action would useful -- without an action it's not possible to create the cache pool and attach it to an existing LV in one step, it would be necessary to call `do_it` and then `attach_cache` with the newly created cache pool as an argument.

This would of course require addition of two new actions for attaching and detaching the cache making the actions API potentially more complicated.

Strictly speaking we don't need the actions for the storage role -- it uses actions internally, but executes them all immediately for every task so manually running `attach_cache` for this case wouldn't be a huge issue.

Thoughts @dwlehman @japokorn

PR #1004, opened 2021-12-30
LVM cache pool support

vojtechtrefny:
This adds support for creating LVM cache pools, the goal is to be able to add cache to an existing LV, we can currently only attach an existing cache pool to an LV or create cached LV (create the LV and the cache pool together).

This doesn't contain support for attaching (or detaching) the cache pool using an action, this is currentl being discussed in #1003


japokorn (approved):
Looks good to me.

PR #1005, opened 2022-01-06
Adjust to Stratis 3.0 API

vojtechtrefny:
We plan to support only the new 3.0 API so we don't need to be
backward compatible.


japokorn (approved):
Looks good to me.

PR #1006, opened 2022-01-07
Do no try to read cache MD size for inactive LVs from cache stats

vojtechtrefny:
We read cache stats from the DM table so it's not available for
inactive logical volumes. We can use the metadata size we get from
the internal LV in this case.

Resolves: rhbz#1893553


blivet/devices/lvm.py
```diff
@@ -2692,7 +2692,10 @@ def size(self):
     @property
     def md_size(self):
```
japokorn, line 2693: This condition could be perhaps shortened as:
```
if self.exists and self.stats:
   return self.stats.md_size
else:
    return self._md_size
```

japokorn (approved):
One minor nitpick. Otherwise looks good to me.

PR #1007, opened 2022-01-07
Use subvolume mountpoints when listing btrfs subvolumes

vojtechtrefny:
We are currently using the btrfs volume mountpoint when trying to
get list of subvolumes and if the volume isn't mounted, we don't
show the subvolumes. But if one of the subvolumes is mounted
(like with the default Fedora installation where / and /home are
mounted) we can also use the subvolume mountpoint to run the
"btrfs subvol list <mnt>" command. This change makes sure we check
all the possible mountpoints to show the subvolumes.

Resolves: rhbz#1856048

### DIFF
--- a/blivet/devices/btrfs.py
+++ b/blivet/devices/btrfs.py
@@ -38,6 +38,7 @@ from .. import udev
 from .. import util
 from ..formats import get_format, DeviceFormat
 from ..size import Size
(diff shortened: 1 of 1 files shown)


japokorn (approved):
Looks good to me.

PR #1008, opened 2022-01-10
Fix object type for ActionConfigureDevice

vojtechtrefny:
It should be ACTION_OBJECT_DEVICE not ACTION_OBJECT_FORMAT.


japokorn (approved):
Looks good to me.

PR #1011, opened 2022-01-13
Exclude unusable disks from PartitionFactory 

vojtechtrefny:
For multipath devices anaconda creates the `PartitionFactory` with both the multipath device and the disks (so for example with `disks=["sda", "sdb", "mpatha"]`). Technically speaking this isn't correct but we allow this in other factories and when `LVMFactory` is used we exclude the disks in the `PartitionSetFactory` when creating the physical volumes so we should do the same with the `PartitionFactory`.


japokorn (approved):
Looks good to me.

PR #1015, opened 2022-01-27
Make sure we mount the top level subvolume when mounting btrfs

vojtechtrefny:
If we don't specify the subvolid=5 we end up mounting the default
subvolume and if it isn't the top volume one, we can't remove
other subvolumes "higher" in the hierarchy.

This makes the installer crash when reinstalling over a preexisting
btrfs installation, for example the default openSUSE installation
where the default subvolume is a post installation snapshot.

Resolves: rhbz#2026205

### DIFF
--- a/blivet/devices/btrfs.py
+++ b/blivet/devices/btrfs.py
@@ -123,7 +123,8 @@ class BTRFSDevice(StorageDevice):
         else:
             tmpdir = tempfile.mkdtemp(prefix=self._temp_dir_prefix)
             try:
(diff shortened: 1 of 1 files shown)


japokorn (approved):
Looks good to me.

cmurf:
Not to suggest a change in strategy, but just extra info: it is possible to delete subvolumes without "line of sight" via path-to-suvolume. One is `btrfs subvolume delete --subvolid <subvolid>` option. Less obvious, is the example in libbtrfsutil, both C API and Python bindings have an _fd variant for removing subvolumes. There's also a recursive delete subvolume option, not available in the CLI commands.

PR #1018, opened 2022-02-08
Fix exception message when trying to format an non-existing device

vojtechtrefny:


vojtechtrefny:
The rawhide failure is caused by a kernel bug so it is expected to fail but it now fails with a nicer exception.

japokorn (approved):
Looks good to me.

PR #1019, opened 2022-02-14
Set partition flags after setting parted filesystem (#2033875)

vojtechtrefny:
With latest parted setting the parted filesystem property changes
the GPT type set by the flags to the default type for the fs, e.g.
to "Microsoft basic data" for FAT instead of "EFI System" which
is set with the boot flag on GPT.


vojtechtrefny:
Jenkins, test this please.

japokorn (approved):
Looks good to me.

PR #1020, opened 2022-02-15
anaconda tests: Install older version of npm in the container

vojtechtrefny:
Related: https://github.com/rhinstaller/anaconda/commit/aae34587bc246d31eff2b229f40b1ccc87adb9b5


vojtechtrefny:
@jkonecny12 can you check this? I have no idea what am I doing here, but the tests failed repeatedly on #1019 but passed here, so I assume I found the correct fix :-)

VladimirSlavik:
I believe you have it right but don't know enough to ack, either.

jkonecny12:
Hi, thanks for asking. I think this is the correct fix https://github.com/rhinstaller/anaconda/pull/3868 .

PR #1021, opened 2022-02-16
Correctly cancel configure actions in cancel()

vojtechtrefny:
The action also must be marked as not applied which is what the
DeviceAction.cancel() call does for all actions.


japokorn (approved):
Looks good to me.

PR #1023, opened 2022-02-16
Make sure configure actions obsolete only actions with same attribute

vojtechtrefny:
By default an action obsoletes all actions with the same type on
the same device. For configure actions this makes sense only if
the actions changes the same attribute.


japokorn (approved):
CI failures seem to be unrelated to this PR.
Looks good to me.

PR #1025, opened 2022-02-21
Create loop devices for tests with --partscan

vojtechtrefny:
On latest rawhide newly created partitions are not added when the
loop device is created without the partscan option.
Related: rhbz#2055708


japokorn (approved):
Looks good to me.

Issue #996, opened 2021-10-14
Separate blivet test suite into "unit" and "integration" test


vojtechtrefny:
The blivet test suite is currently a mix of unit and integration tests, or to be more precise mix of tests that work with real storage (disk files or loop devices) and tests that use fake devices and mock, without clear distinction between those two. Sometimes tests that need "real" storage devices are mixed with tests that are 100 % mocked in a single file.

This causes some issue like #870 when running tests as non-root user doesn't correctly skip all tests that needs root privileges to create or probe devices or #995 where we try to probe an existing device in a test that should be just a pure unit test without touching real hardware.

We should fully "separate" these tests into two test suites: *unit tests* that can run as a non-root user and don't need access to block devices and *integration tests* that create their own block devices (so mostly `LoopBackedTestCase` and `ImageBackedTestCase` test cases).

PR #1028, opened 2022-03-09
Separate blivet test suite into "unit" and "storage" parts

vojtechtrefny:
This is mostly moving the tests from one place to another and making sure the "unit" tests can be run without root privileges and with missing dependencies. Next step is introducing targetcli devices in the "storage" tests instead of using loop devices and start covering blivet API with tests that use "real" storage (we have a good coverage with the `fstesting` module, but devices are basically not covered at all).

Fixes: #996


japokorn (approved):
Cosmetic typo. Looks good to me.

PR #1029, opened 2022-03-10
Disable Fedora ELN builds in packit

vojtechtrefny:
Fedora ELN builds are broken all the time because of repository
or GPG signatures issues, it doesn't make sense to have a test
that is broken all the time and outside of our control.


japokorn (approved):
Looks good to me.

Issue #1027, opened 2022-03-02
Add support for LVM writecache


vojtechtrefny:
With the recently [added support for LVM cache](https://github.com/storaged-project/blivet/pull/1004) we should also consider adding support for LVM writecache which shouldn't be much different from blivet perspective, we already [have "read-only" support](https://github.com/storaged-project/blivet/pull/854) and libblockdev [already supports creating LVM cache](https://github.com/storaged-project/libblockdev/pull/546).

vojtechtrefny:
Closed via #1033

PR #1033, opened 2022-04-07
Add support for attaching and creating LVM writecached LVs

vojtechtrefny:
This adds support to attach an existing LV as a writecache to
another LV and to create a writecached LV.

Fixes: #1027 

This is a good candidate for more tests, especially something that will actually try to attach and detach the writecache LV so I'll wait for #1028 to be merged and add more tests later.


japokorn (approved):
Looks good to me.

PR #1034, opened 2022-04-13
ActionDestroyDevice should not obsolete ActionRemoveMember

vojtechtrefny:
If we want to remove a PV from a VG and then remove the PV device,
the ActionDestroyDevice must not obsolete the ActionRemoveMember
action. Eventhough we are going to remove the device, we still
need to call "vgreduce" first.


japokorn (approved):
Looks good to me.
Loop device related issues are addressed by https://github.com/storaged-project/blivet/pull/1025

PR #1035, opened 2022-04-16
Add support for NPIV-enabled zFCP devices

jstodola:
This change adds a support for [zFCP NPIV (N_Port ID virtualization)](https://www.ibm.com/docs/en/linux-on-systems?topic=later-n-port-id-virtualization) devices.
NPIV-enabled devices are activated just by using the device ID. The kernel module will detect the WWPNs and LUNs and bring all the devices up automatically. This means the user doesn't have to provide the WWPN and LUN IDs.

Related: rhbz#[1937030](https://bugzilla.redhat.com/show_bug.cgi?id=1937030)

TODO:

- [x] additional testing in anaconda

### DIFF
--- a/blivet/zfcp.py
+++ b/blivet/zfcp.py
@@ -397,6 +397,44 @@ class ZFCPDevice(ZFCPDeviceBase):
         return True
 
 
+class ZFCPNPIVDevice(ZFCPDeviceBase):
+    """Class for zFCP devices configured in NPIV mode. Only a zFCP device number is
+    needed for such devices.
+    """
+
+    def online_device(self):
+        """Initialize the device and make its storage block device(s) ready to use.
+
+        :returns: True if success
+        :raises: ValueError if the device cannot be initialized
+        """
+
+        super().online_device()
+
+        if not is_npiv_enabled(self.devnum):
+            raise ValueError(_("zFCP device %s cannot be used in NPIV mode.") % self)
+
+        return True
+
+    def offline_device(self):
+        """Remove the zFCP device from the system.
+
+        :returns: True if success
+        :raises: ValueError if the device cannot be brought offline
+         """
+
+        try:
+            self.offline_scsi_device()
+        except OSError as e:
+            raise ValueError(_("Could not correctly delete SCSI device of "
+                               "zFCP %(zfcpdev)s (%(e)s).")
+                             % {'zfcpdev': self, 'e': e})
+
+        self._set_zfcp_device_offline()
+
+        return True
+
+
 class zFCP:
 
     """ ZFCP utility class.
@@ -439,7 +477,12 @@ class zFCP:
 
             fields = line.split()
 
-            if len(fields) == 3:
+            # NPIV enabled device
+            if len(fields) == 1:
+                devnum = fields[0]
+                wwpn = None
+                fcplun = None
+            elif len(fields) == 3:
                 devnum = fields[0]
                 wwpn = fields[1]
                 fcplun = fields[2]
@@ -458,8 +501,12 @@ class zFCP:
             except ValueError as e:
                 log.warning("%s", str(e))
 
-    def add_fcp(self, devnum, wwpn, fcplun):
-        d = ZFCPDevice(devnum, wwpn, fcplun)
+    def add_fcp(self, devnum, wwpn=None, fcplun=None):
+        if wwpn and fcplun:
+            d = ZFCPDevice(devnum, wwpn, fcplun)
+        else:
+            d = ZFCPNPIVDevice(devnum)
+
         if d.online_device():
             self.fcpdevs.add(d)
 
--- a/blivet/zfcp.py
+++ b/blivet/zfcp.py
@@ -21,6 +21,7 @@
 
 import os
 import re
+from abc import ABC
 from . import udev
 from . import util
 from .i18n import _
@@ -46,29 +47,19 @@ zfcpsysfs = "/sys/bus/ccw/drivers/zfcp"
 scsidevsysfs = "/sys/bus/scsi/devices"
 zfcpconf = "/etc/zfcp.conf"
 
+class ZFCPDeviceBase(ABC):
+    """An abstract base class for zFCP storage devices."""
 
-class ZFCPDevice:
-    """
-        .. warning::
-            Since this is a singleton class, calling deepcopy() on the instance
-            just returns ``self`` with no copy being created.
-    """
-
-    def __init__(self, devnum, wwpn, fcplun):
+    def __init__(self, devnum):
         self.devnum = blockdev.s390.sanitize_dev_input(devnum)
-        self.wwpn = blockdev.s390.zfcp_sanitize_wwpn_input(wwpn)
-        self.fcplun = blockdev.s390.zfcp_sanitize_lun_input(fcplun)
-
         if not self.devnum:
             raise ValueError(_("You have not specified a device number or the number is invalid"))
-        if not self.wwpn:
-            raise ValueError(_("You have not specified a worldwide port name or the name is invalid."))
-        if not self.fcplun:
-            raise ValueError(_("You have not specified a FCP LUN or the number is invalid."))
+
+        self._device_online_path = os.path.join(zfcpsysfs, self.devnum, "online")
 
     # Force str and unicode types in case any of the properties are unicode
     def _to_string(self):
-        return "%s %s %s" % (self.devnum, self.wwpn, self.fcplun)
+        return str(self.devnum)
 
     def __str__(self):
         return stringize(self._to_string())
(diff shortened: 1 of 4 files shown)


blivetghoul:

vojtechtrefny:
Jenkins, ok to test.

jstodola:
pep8 errors fixed.

vojtechtrefny:
Btw. my first reaction was that all the code around switching the zFCP devices online/offline and similar stuff should be moved to libblockdev, but we [already did do that](https://github.com/storaged-project/libblockdev/commit/bfa6ceec76c2978762aaf96154a706e0d677f620) and just never changed blivet to actually use it. But that's out of scope for this PR.

jstodola:
Tested successfully together with anaconda [PR-4188](https://github.com/rhinstaller/anaconda/pull/4188) and pykickstart [PR-410](https://github.com/pykickstart/pykickstart/pull/410) changes. No other work is expected on this PR.

blivet/zfcp.py
```diff
@@ -248,20 +384,53 @@ def offline_device(self):
...
+        """Initialize the device and make its storage block device(s) ready to use.
+
+        :returns: True if success
+        :raises: ValueError if the device cannot be initialized
+        """
+
+        super().online_device()
+
+        if not is_npiv_enabled(self.devnum):
+            raise ValueError(_("zFCP device %s cannot be used in NPIV mode.") % self)
```
steffen-maier, line 415: raise ValueError(_("zFCP device %s cannot use auto LUN scan.") % self)

blivet/zfcp.py
```diff
@@ -327,8 +501,12 @@ def read_config(self):
             except ValueError as e:
                 log.warning("%s", str(e))
 
-    def add_fcp(self, devnum, wwpn, fcplun):
-        d = ZFCPDevice(devnum, wwpn, fcplun)
+    def add_fcp(self, devnum, wwpn=None, fcplun=None):
+        if wwpn and fcplun:
+            d = ZFCPDevice(devnum, wwpn, fcplun)
+        else:
+            d = ZFCPNPIVDevice(devnum)
```
steffen-maier, line 508: d = ZFCPDeviceAutoLunScan(devnum)

blivet/zfcp.py
```diff
@@ -327,8 +501,12 @@ def read_config(self):
             except ValueError as e:
                 log.warning("%s", str(e))
 
-    def add_fcp(self, devnum, wwpn, fcplun):
-        d = ZFCPDevice(devnum, wwpn, fcplun)
+    def add_fcp(self, devnum, wwpn=None, fcplun=None):
+        if wwpn and fcplun:
+            d = ZFCPDevice(devnum, wwpn, fcplun)
```
steffen-maier, line 506: optionally instead:
`d = ZFCPDeviceFullPath(devnum, wwpn, fcplun)`

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+    except OSError as e:
+        log.warning("Couldn't read the port_type attribute of the %s device: %s", device_id, str(e))
+        port_in_npiv_mode = False
+
+    return port_in_npiv_mode
+
+
+def is_npiv_enabled(device_id):
+    """Return True if the given zFCP device ID is configured and usable in
+    NPIV (N_Port ID Virtualization) mode.
```
steffen-maier, line 82: """Return True if the given zFCP device ID is configured in
    NPIV (N_Port ID Virtualization) mode and zfcp auto LUN scan is not disabled.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        scsi_device_found = False
+        for scsidev in scsi_devices:
+            fcpsysfs = os.path.join(scsidevsysfs, scsidev)
+
+            with open(os.path.join(fcpsysfs, "hba_id")) as f:
+                fcphbasysfs = f.readline().strip()
+            with open(os.path.join(fcpsysfs, "wwpn")) as f:
+                fcpwwpnsysfs = f.readline().strip()
+            with open(os.path.join(fcpsysfs, "fcp_lun")) as f:
+                fcplunsysfs = f.readline().strip()
```
steffen-maier, line 198: FYI.
Since this is only updated old code, it's OK to keep deprecated and zfcp-specific sysfs attributes rather than using common code mechanisms, which I would prefer for new code.
https://www.ibm.com/docs/en/linux-on-systems?topic=devices-mapping-sysfs-representations

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        try:
+            logged_write_line_to_file(self._device_online_path, "0")
+        except OSError as e:
+            raise ValueError(_("Could not set zFCP device %(devnum)s "
+                               "offline (%(e)s).")
+                             % {'devnum': self.devnum, 'e': e})
+
+    def _is_scsi_associated_with_fcp(self, fcphbasysfs, _fcpwwpnsysfs, _fcplunsysfs):
+        """Decide if the SCSI device with the provided SCSI attributes
+        corresponds to the zFCP device.
```
steffen-maier, line 165: """Decide if the provided FCP addressing
corresponds to the zFCP device.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        if not self.fcplun:
+            raise ValueError(_("You have not specified a FCP LUN or the number is invalid."))
+
+    # Force str and unicode types in case any of the properties are unicode
+    def _to_string(self):
+        return "{} {} {}".format(self.devnum, self.wwpn, self.fcplun)
+
+    def _is_scsi_associated_with_fcp(self, fcphbasysfs, fcpwwpnsysfs, fcplunsysfs):
+        """Decide if the SCSI device with the provided SCSI attributes
+        corresponds to the zFCP device.
```
steffen-maier, line 232: """Decide if the provided FCP addressing
corresponds to the path stored in the zFCP device.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+
+        self.fcplun = blockdev.s390.zfcp_sanitize_lun_input(fcplun)
+        if not self.fcplun:
+            raise ValueError(_("You have not specified a FCP LUN or the number is invalid."))
+
+    # Force str and unicode types in case any of the properties are unicode
+    def _to_string(self):
+        return "{} {} {}".format(self.devnum, self.wwpn, self.fcplun)
+
+    def _is_scsi_associated_with_fcp(self, fcphbasysfs, fcpwwpnsysfs, fcplunsysfs):
```
steffen-maier, line 230: I would drop "_scsi" as this method does not take or use a SCSI device.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        """
+
+        try:
+            logged_write_line_to_file(self._device_online_path, "0")
+        except OSError as e:
+            raise ValueError(_("Could not set zFCP device %(devnum)s "
+                               "offline (%(e)s).")
+                             % {'devnum': self.devnum, 'e': e})
+
+    def _is_scsi_associated_with_fcp(self, fcphbasysfs, _fcpwwpnsysfs, _fcplunsysfs):
```
steffen-maier, line 163: I would drop "_scsi" as this method does not take or use a SCSI device.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        """Decide if the SCSI device with the provided SCSI attributes
+        corresponds to the zFCP device.
+
+        :returns: True or False
+        """
+
+        return fcphbasysfs == self.devnum
```
steffen-maier, line 170: Oh, hang on, shouldn't this method be defined in class `ZFCPDeviceAutoLunScan`, as opposed to this abstract `ZFCPDeviceBase`, which does not know whether auto lun scan is available or not?
But then again you seem to need it here to move method offline_scsi_device() to the base class so both derived classes can use it as common method.

blivet/zfcp.py
```diff
@@ -167,36 +327,9 @@ def online_device(self):
...
-                logged_write_line_to_file(scsidel, "1")
-                udev.settle()
-                return
-
-        log.warning("no scsi device found to delete for zfcp %s %s %s",
-                    self.devnum, self.wwpn, self.fcplun)
-
     def offline_device(self):
-        offline = "%s/%s/online" % (zfcpsysfs, self.devnum)
+        """Remove the zFCP device from the system."""
```
steffen-maier, line 331: nit pick: Setting an FCP device offline (which was online before) does not remove it from the system, it keeps all the state info and Linux driver core sysfs tree with zfcp_adapter,  zfcp_port-s, and zfcp_unit-s and only sets it into an offline state. (Reason: it can recover the previous state on the next online operation.)
No need to change the comment, though.

steffen-maier (commented):
Hi Jan, thanks a lot for the code changes. Very much appreciated. I cannot wait to try them. Regarding user messages and code maintenance, I suggest small changes to comments, user messages, and code indentifiers. They are based on potential conceptual misunderstandings:

- The code change is about enabling users to take advantage of a Linux kernel (module) feature called zfcp automatic LUN scan.
- It is not about enabling NPIV-enabled FCP devices. (Users deploy them since NPIV support appeared in IBM Z HBAs and the zfcp device driver; using NPIV is best practice since many years)

Those are two different things.

I'm aware it's already merged, so maybe a simple fixup commit would be possible?

Apologies that I review the entire PR as one chunk, instead of individual commits. I intended to do the latter but confused myself and had already accidentally put most review comments in this entire PR diff.

blivet/zfcp.py
```diff
@@ -46,63 +47,117 @@ def logged_write_line_to_file(fn, value):
...
+        :returns: True if success
+        :raises: ValueError if the device cannot be initialized
+        """
+
+        self._free_device()
+        self._set_zfcp_device_online()
+        return True
+
+
+class ZFCPDevice(ZFCPDeviceBase):
```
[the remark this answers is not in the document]
jstodola: I intentionally kept this class name so that it doesn't cause any problems for possible projects using the previous code. This could be a problem for RHEL-8/9.

blivet/zfcp.py
```diff
@@ -47,62 +49,218 @@ def logged_write_line_to_file(fn, value):
...
+        portdir = "%s/%s/%s" % (zfcpsysfs, self.devnum, self.wwpn)
+        unitadd = "%s/unit_add" % (portdir)
+        unitdir = "%s/%s" % (portdir, self.fcplun)
+        failed = "%s/failed" % (unitdir)
+
+        # Activating an NPIV enabled device using devnum, WWPN and LUN should still be possible
+        # as this method was used as a workaround until the support for NPIV enabled devices has
+        # been implemented. Just log a warning message and continue.
+        if is_npiv_enabled(self.devnum):
+            log.warning("zFCP device %s in NPIV mode brought online. All LUNs will be activated "
```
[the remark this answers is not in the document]
jstodola: I think that warning is more appropriate in this case. When the user activates a zFCP device using the device number, WWPN and LUN, he might want to activate just the one LUN and not all available LUNs. It's good to be warned that more LUNs might be activated (and possibly erased during the installation process).

jstodola:
@steffen-maier thanks for you additional review and comments. I've created a new PR https://github.com/storaged-project/blivet/pull/1061 implementing most of the suggestions.

blivet/zfcp.py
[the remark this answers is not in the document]
steffen-maier: > ZFCPDeviceFullPath could be done like this, what do you think? [216fb82](https://github.com/storaged-project/blivet/commit/216fb82c386449547f542b5ce43bca4bd99eccf6)

Yes, like this. Not sure how python works regarding the singleton `zfcp = zFCP()` assigning an object of either subclass to a (non?!-member) variable `d` and then invoking (virtual) member functions `d.online_device()` or  `d.offline_device()` and remembering the objects in a set `self.fcpdevs`. Maybe python infers `None` (or whatever the world's top-most unknown pseudo class would be) as type of `d` and as type of set members. In another more strictly-typed OO language, I might have used an abstract ZFCPDevice (instead of ZFCPDeviceBase), use it as type for `d`, and derive the two concrete ZFCPDevice... classes from it. IIRC, ZFCPDevice... is not used anywhere else in the code so it's really encapsulated here in the singleton and there should be no other depencies. The only used (by anaconda UI or pykickstart) interfaces of the singleton seem to be:

```
pyanaconda/modules/storage/kickstart.py:273:            zfcp.add_fcp(fcp.devnum, fcp.wwpn, fcp.fcplun)
pyanaconda/modules/storage/reset.py:78:            zfcp.startup()
pyanaconda/modules/storage/zfcp/discover.py:80:            zfcp.add_fcp(self._device_number, self._wwpn, self._lun)
pyanaconda/modules/storage/zfcp/zfcp.py:58:        zfcp.startup()
pyanaconda/modules/storage/zfcp/zfcp.py:78:        zfcp.write(conf.target.system_root)
```

This stuff here in the singleton is only for explicit online/offline and/or zfcp path management. In contrast, ZFCPDiskDevice exists for each SCSI disk device enumerated by libudev and backed by zfcp-attachment, so that's independent of ZFCPDevice... here. The stuff here triggers the appearance of the SCSI disk devices over there (handling `dracut_setup_args`).

> 
> But at this point I do not plan doing any other changes in the code due to other priorities.

Fair enough. Let's get in what we have.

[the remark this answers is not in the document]
jstodola: > Yes, like this. Not sure how python works regarding the singleton `zfcp = zFCP()` assigning an object of either subclass to a (non?!-member) variable `d` and then invoking (virtual) member functions `d.online_device()` or `d.offline_device()` and remembering the objects in a set `self.fcpdevs`. 

I tested the change before sending it and there was no issue when using the `zFCP()` singleton. But I could only test devices with auto LUN scan disabled (by `zfcp.allow_lun_scan=0`) or enabled, but not mixing both device types at the same time (if it's at all possible).

Anyway, I've updated https://github.com/storaged-project/blivet/pull/1061 with renaming the ZFCPDevice class.

PR #1037, opened 2022-04-28
Do not crash when changing disklabel on disks with active devices

vojtechtrefny:
The _find_active_devices_on_action_disks function originally
prevented from making any changes on disks with active devices
(active LVs, mounted partitions etc.)  This was changed in
b72e957d2b23444824316331ae21d1c594371e9c and the check currently
prevents only reformatting the disklabel on such disks which
should be already impossible on disks with an existing partition.

This change for the 3.4 stable branch keeps the current behaviour
where the active devices are teared down when running in installer
mode to avoid potential issues with the installer.

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -211,9 +211,8 @@ class ActionList(object):
                     except StorageError as e:
                         log.info("teardown of %s failed: %s", device.name, e)
             else:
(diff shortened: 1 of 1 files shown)


japokorn (approved):
Looks good to me.
Loop device related issues are addressed by https://github.com/storaged-project/blivet/pull/1025

PR #1038, opened 2022-04-28
Do not check for "problematic" disks in ActionList._pre_process

vojtechtrefny:
The _find_active_devices_on_action_disks function originally
prevented from making any changes on disks with active devices
(active LVs, mounted partitions etc.) This was changed in
b72e957 and the check currently prevents only reformatting
the disklabel on such disks which should be already impossible
on disks with an existing partition.

The check currently also makes it impossible to add a new disk to
an existing VG -- the new disk has an active device (the VG) in
our storage model so creating a new partition table (even if it
didn't have one before) marks it as "problematic".

----

This is a less conservative version of #1037 removing the check completely. I mean, what could go wrong, right?

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -29,8 +29,7 @@ from .deviceaction import ActionCreateDevice
 from .deviceaction import action_type_from_string, action_object_from_string
 from .devicelibs import lvm
 from .devices import PartitionDevice
-from .errors import DiskLabelCommitError, StorageError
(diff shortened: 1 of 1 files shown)


japokorn (approved):
Looks good to me.

vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
Jenkins, test this please.

PR #1040, opened 2022-05-02
Correctly set vg_name after adding/removing a PV from a VG

vojtechtrefny:
Without setting the LVMPhysicalVolume.vg_name argument to None
after removing the PV from its VG, the PV is still considered
active and cannot be removed.

Yet another issue discovered when working on https://github.com/linux-system-roles/storage/pull/264


vojtechtrefny:
Jenkins, test this please.

vojtechtrefny:
The failed XFS resize test is addressed in https://github.com/storaged-project/blivet/pull/1025 which went only to 3.5-devel (I expected that 3.5.0 will be released by this time already).

japokorn (approved):
Looks good to me.

PR #1042, opened 2022-05-05
Update DeviceFormat.target_size in update_size_info

vojtechtrefny:
DeviceFormat.size returns target_size for resizable format so
without updating it in update_size_info we'll always get old or
invalid value for size.

---

One more issue discovered thanks to https://github.com/linux-system-roles/storage/pull/264 -- because we don't know the lvmpv format size we fail to call `pvmove` before running `vgreduce` and removing a non-empty PV from a VG will fail.

### DIFF
--- a/blivet/formats/__init__.py
+++ b/blivet/formats/__init__.py
@@ -412,6 +412,9 @@ class DeviceFormat(ObjectID):
 
     def update_size_info(self):
         """ Update this format's current and minimum size (for resize). """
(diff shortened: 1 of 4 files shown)


vojtechtrefny:
Replaced by https://github.com/storaged-project/blivet/pull/1043 the dependency of `size` on `target_size` is actually expected and we need to use `current_size` instead.

PR #1043, opened 2022-05-05
Use LVM PV format current_size in LVMVolumeGroupDevice._remove

vojtechtrefny:
The member format size is 0 when target size is not set.


japokorn (approved):
Looks good to me.
loop related errors in CI are not related to this PR.

PR #1045, opened 2022-05-09
Add a very simple NVMe module

vojtechtrefny:
This covers only the basic functionallity needed by Anaconda right
now: populating the config files in /etc/nvme and copying them to
the installed system. The API for the NVMe singleton is based on
the similar modules for iSCSI and FCoE.

Resolves: rhbz#2073008


tbzatek (approved):
Alright, go for it!

PR #1046, opened 2022-05-16
Do not fail when we can't get LVM cache information (#2086310)

vojtechtrefny:


blivet/devices/lvm.py
```diff
@@ -2854,7 +2854,13 @@ def exists(self):
     def stats(self):
         # to get the stats we need the cached LV to exist and be activated
         if self._exists and self._cached_lv.status:
-            return LVMCacheStats(blockdev.lvm.cache_stats(self._cached_lv.vg.name, self._cached_lv.lvname))
+            try:
+                stats = blockdev.lvm.cache_stats(self._cached_lv.vg.name, self._cached_lv.lvname)
+            except blockdev.LVMError as lvmerr:
+                log.error("Failed to get cache stats for '%s': %s", self._cached_lv.name, lvmerr)
+                return None
+            else:
```
japokorn, line 2862: I think that `else` can be omitted here (also in the second changed block of code).

japokorn (approved):
Looks good to me.

PR #1047, opened 2022-05-16
Add support for LVM RAID raid0 level

vojtechtrefny:
When we initially added support for LVM RAID, raid0 wasn't
supported yet so we added support only for the old "striped"
segment type. RAID0 is now supported by LVM so we should support
it too.

Related: https://github.com/linux-system-roles/storage/pull/271


japokorn (approved):
Looks good to me.

PR #1050, opened 2022-05-30
Add option to attach a newly created cache pool to existing LV

vojtechtrefny:
Because we do not have action for attaching the cache pool, we
cannot schedule both adding the fast PV to the VG and attaching
the cache pool to existing LV. This hack allows to schedule the
attach to happen after the cache pool is created.


japokorn (approved):
Looks good to me.
Tests are in #1051

PR #1051, opened 2022-05-30
tests: Add test for creating and attaching cache pools

vojtechtrefny:
Test for #1050 on 3.6-branch.


japokorn (approved):
Looks good to me. 
`lvm.py` changes are the same as those in #1050 so the tests can be run. After #1050 is merged this one should be rebased.

PR #1053, opened 2022-06-10
ci: Run static analysis checks in GitHub actions

vojtechtrefny:
We don't really need to run the checks (pylint, pycodestyle and translation-canary) in our CI, everything can run in the GH actions on Ubuntu without a problem. This uses our ansible playbook to install dependencies including pocketlint and latest pylint from pip so the action itself is actually really simple.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,0 +1,28 @@
+name: Static Analysis
+
+on:
+  pull_request:
+    branches:
+     - 3.*-devel
+     - master
+  push:
+    branches:
+     - 3.*-devel
+     - master
+  schedule:
+    - cron: 0 0 * * 0
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install dependencies
+      run:   |
+         sudo apt-get -qq update
+         sudo apt-get -y -qq install make ansible
+         sudo make install-requires
+    - name: Run checks
+      run: make check


vojtechtrefny:
The test itself now fails, but these pylint issues are fixed in #1052, the action itself seems to be working well.

.github/workflows/check.yml
```diff
@@ -0,0 +1,28 @@
...
+  schedule:
+    - cron: 0 0 * * 0
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
```
VladimirSlavik, line 21: Latest is v3, but of course v2 works too.

.github/workflows/check.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    branches:
+     - 3.*-devel
+     - master
+  push:
+    branches:
+     - 3.*-devel
+     - master
+  schedule:
+    - cron: 0 0 * * 0
```
VladimirSlavik, line 13: I'm not sury why run this regularly too? You should have 100% coverage with the triggers above. Also, it won't run on the 3.*-devel kind of branch.

I guess the cron trigger would be useful for external changes that can break the checks regardless of code?

VladimirSlavik (approved):
Looks good to me.

.github/workflows/check.yml
vojtechtrefny: Thanks, looks like I was copy-pasting from an old how-to :-)

.github/workflows/check.yml
vojtechtrefny: >Also, it won't run on the 3.*-devel kind of branch.

If I understand it correctly it will run on the default branch, which should be enough.

> I guess the cron trigger would be useful for external changes that can break the checks regardless of code?

Yes, that's the general idea. It's probably not as important for blivet which is a relatively active project, but for some of our less active projects where there could be few weeks between patches (like blivet-gui where I added this action too) and I'd like to be notified when something goes wrong sooner than later. With pylint this mostly means new warnings that were detect by a new pylint (or astroid) version. It's always annoying to discover that pylint started complaining about something completely unrelated to my PR just because there was a new version few weeks ago.

VladimirSlavik: Ah, ok, that does make sense.

PR #1055, opened 2022-06-13
Use the new Stratis tool to predict pool and fs used size

vojtechtrefny:
stratis-predict-usage from the latest Stratis 3.1.0 can now
predict used/metadata size for pools with non-existing blockdevs
and also supports predicting used/metadata size for filesystems
so we no longer need to guess these ourselves.


vojtechtrefny:
Replaced by #1058

PR #1058, opened 2022-06-16
Stratis support update for stratis 3.1.0

vojtechtrefny:
This uses the updated tool for predicting pool and filesystem metadata sizes and adds support for specifying size when creating a file system.

### DIFF
--- a/blivet/devicefactory.py
+++ b/blivet/devicefactory.py
@@ -35,6 +35,7 @@ from .devicelibs import mdraid
 from .devicelibs import lvm
 from .devicelibs import raid
 from .devicelibs import crypto
+from .devicelibs import stratis
 from .partitioning import SameSizeSet
 from .partitioning import TotalSizeSet
 from .partitioning import do_partitioning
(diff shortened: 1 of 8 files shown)


vojtechtrefny:
This should be now ready for review. There still might be some issues, but with #1060 I'm pretty sure at least the basics work as expected. The rawhide failure is caused by Python 3.11, pylint doesn't work with it yet.

blivet/devicefactory.py
```diff
@@ -2095,13 +2103,56 @@ def _get_total_space(self):
 
             return space
```
japokorn, line 2104: Is there a reason that this branch is the only one returning a result?

japokorn (approved):
Looks good to me.

blivet/devicefactory.py
vojtechtrefny: Yes, this `return` is unnecessary, the `return` at the end of the function will take care of all branches. Fixed, thanks.

PR #1059, opened 2022-06-24
Add a forced delay to udev settle in chroot environments

win32asm:
When running in a chroot environment (e.g.mock) there is a race between
udev actually settling the discovered loopback/dm device names and a
program running 'udevadm settle', as latter becomes a no-op in chroot:

   <mock-chroot> sh-5.1# udevadm settle
   Running in chroot, ignoring request.

This patch grabs stdout of 'udevadm settle' command, tries to
determine whether the command was ignored, and adds an 1s delay if it was.

1s delay was chosen arbitrary, to succeed in used environment.
However, it should be sufficient for most chrooted environments, as
significant device changes are not expected there.

### DIFF
--- a/blivet/udev.py
+++ b/blivet/udev.py
@@ -26,6 +26,7 @@ import re
 import subprocess
 import logging
 import pyudev
+import time
 
 from . import util
 from .size import Size
@@ -41,6 +42,29 @@ log = logging.getLogger("blivet")
 
 ignored_device_names = []
 """ device name regexes to ignore; this should be empty by default """
+
+
+def running_in_chroot():
+    """ Mimic running_in_chroot() logic from systemd.
+        Simplifications:
+            - can't be running as PID 1, so skip the checks
+            - don't have statfs() call, so can't check for fake /proc
+
+        :returns: True if we detect a chroot environment, False otherwise
+        :rtype: bool
+    """
+    try:
+        # compare inode # for "/" in init with inode # for "/" here
+        # if inode numbers are different - we're in chroot
+        pid1_root_stat = os.stat("/proc/1/root", follow_symlinks=True)
+        my_root_stat = os.stat("/")
+        return pid1_root_stat.st_ino != my_root_stat.st_ino
+    except FileNotFoundError:
+        # "/proc/1/root" does not exist; default to not chroot
+        return False
+    except PermissionError:
+        # not superuser enough; default to not chroot
+        return False
 
 
 def device_to_dict(device):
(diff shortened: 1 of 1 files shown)


StorageGhoul:

blivet/udev.py
```diff
@@ -87,16 +88,23 @@ def get_devices(subsystem="block"):
...
+
     argv = ["udevadm", "settle", "--timeout=300"]
     if quiet:
-        subprocess.call(argv, close_fds=True)
+        proc = subprocess.run(argv, close_fds=True, capture_output=True, text=True, check=True)
+        out = proc.stderr
     else:
-        util.run_program(argv)
+        out = util.capture_output(argv, stderr_to_stdout=True)
+    if out.find('ignoring'):
```
vojtechtrefny, line 103: I was going to ask if we could do something else than parsing the output to check for the error message, like checking the return code, but `udevadm settle` just return 0 in this case. Maybe we could do some chroot detection ourselves?

vojtechtrefny (commented):
I have nothing against this in general and if it helps your use case, I guess there's no harm in merging this. Our relationship with udev is complicated and we multiple discussions about removing the `udevadm settle` calls altogether because these no longer works with the async handlers and are basically `time.sleep` calls anyway. But that's something we need to spend some more time on. Any thoughts @dwlehman @japokorn?

blivet/udev.py
win32asm: For reference:
Per udev logic it bypasses **trigger**, **settle**, **control** and **monitor** commands in chroot, and [chroot detection](https://github.com/systemd/systemd/blob/4e5c39597add4c0e3c479fc11d562f6ea92f22de/src/basic/virt.c#L861) is done by checking SYSTEMD_IGNORE_CHROOT env var and comparing whether **/proc/1/root** and **/** are the same directory.
We can copy this code for our purposes to bypass output parsing.
However, as of now udevadm prints the same message for all bypass scenarios, and it might be 'cleaner' to leave chroot detection to udevadm and analyse output instead.

japokorn: I would like to avoid parsing of the error message. In case the message ever changes, the whole issue will reappear.

I think that we should adapt referenced udev logic. Ideally with a comment describing what and why it is being done. Also it might be actually better for us that it covers only this scenario. Generic check may silently catch something we do not want to be caught.

win32asm: @japokorn do we want to rely on SYSTEMD_IGNORE_CHROOT or only pid 1 root dir?

vojtechtrefny (commented):
The pylint errors related to the `_ped` module should be fixed in #1071 so you can ignore them. @win32asm Can you please rebase this and change the target branch to 3.7-devel (latest devel branch)? Other than that, I think this looks good and can be merged (if you don't plan to add more changes).

win32asm:
@vojtechtrefny please proceed with the merge.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

PR #1060, opened 2022-06-28
Add storage tests for Stratis

vojtechtrefny:


vojtechtrefny:
The failure on rawhide is caused by Python 3.11 and the failures on Debian by old version of python-rtslib.

PR #1062, opened 2022-07-14
Misc pylint fixes for rawhide

vojtechtrefny:
Pylint finally works with Python 3.11 on rawhide and shows some new warnings.


japokorn (approved):
Looks good to me.

PR #1066, opened 2022-08-17
DDF RAID support using mdadm

vojtechtrefny:


blivet/populator/helpers/disk.py
```diff
@@ -68,6 +68,8 @@ def run(self):
         log_method_call(self, name=name)
 
         kwargs = self._get_kwargs()
+        if not kwargs:
+            return
```
dwlehman, line 72: How can `kwargs` be empty if you always set `kwargs['biosraid']`? (Or is there some way these changes introduce a way for that to happen for some other type of disk?)

vojtechtrefny: This is the older fix for the issue https://github.com/storaged-project/blivet/pull/1031 when the RAID was still handled by `DMRaidFormatPopulator` which can return empty kwargs. So strictly speaking this isn't necessary now but I wanted to include it just in case.

dwlehman (approved):
Thanks for answering my question. This looks good to me. It's almost as simple as I hoped it would be.

vojtechtrefny:
Merged manually

PR #1068, opened 2022-08-18
Fix missing whitespaces around not keyword

vojtechtrefny:
Latest pycodestyle is complaining about missing whitespace around
"not" which we were (at some places) using as a function instead
of keyword like "not(...)".


VladimirSlavik (approved):
Looks good to me.

PR #1070, opened 2022-08-23
Adjust to new XFS min size

vojtechtrefny:
Starting with xfsprogs 5.19 minimal size for XFS is 300 MiB.


vojtechtrefny:
Jenkins, test this please.

PR #1077, opened 2022-10-19
Add a basic support for NVMe and NVMe Fabrics devices

vojtechtrefny:
This adds two new device types: NVMeNamespaceDevice and NVMeFabricsNamespaceDevice mostly to allow to differentiate between "local" and "remote" NVMe devices. The new libblockdev NVMe plugin is required for full functionality.


tbzatek (approved):
Didn't understand much of it but looks good when reading it like a book...

blivet/devices/disk.py
```diff
@@ -725,3 +728,101 @@ def description(self):
     @property
     def sector_size(self):
         return self._sector_size
+
+
+NVMeController = namedtuple("NVMeController", ["name", "serial", "nvme_ver", "id", "subsysnqn"])
+
+
+class NVMeNamespaceDevice(DiskDevice):
```
dwlehman, line 736: It would be nice if this could be called `NVMeDevice` and so on for the other device and the populators, but I wouldn't want it to be confusing or vague if that's a possibility.

blivet/udev.py
```diff
@@ -1023,6 +1023,39 @@ def device_is_nvdimm_namespace(info):
...
+        # nvme plugin is not available -- even if this is an nvme fabrics device we
+        # don't have tools to work with it, so we should pretend it's just a normal nvme
+        return False
+
+    controllers = blockdev.nvme_find_ctrls_for_ns(info.get("SYS_PATH", ""))
+    if not controllers:
+        return False
+
+    transport = util.get_sysfs_attr(controllers[0], "transport")
+    return transport in ("rdma", "fc", "tcp", "loop")
```
dwlehman, line 1056: At some point we might want to store this information somewhere, but I guess we shouldn't until we have a use for it.

blivet/util.py
```diff
@@ -449,6 +449,15 @@ def get_sysfs_path_by_name(dev_node, class_name="block"):
...
 
 
+def get_path_by_sysfs_path(sysfs_path, dev_type="block"):
+    """ Return device path for a given device sysfs path. """
+
+    dev = get_sysfs_attr(sysfs_path, "dev")
+    if not dev or not os.path.exists("/dev/%s/%s" % (dev_type, dev)):
+        raise RuntimeError("get_path_by_sysfs_path: Could not find device for %s" % sysfs_path)
+    return os.path.realpath("/dev/%s/%s" % (dev_type, dev))
+
```
dwlehman, line 459: A comment noting that this will not return "friendly" names would be good, even though I realize it's a non-issue for how you're using it here.

dwlehman (approved):
Looks good. I have a few comments, but nothing big or critically important.

blivet/util.py
vojtechtrefny: Ok, I added a note to the docstring to clarify the returned symlink.

blivet/devices/disk.py
vojtechtrefny: Strictly speaking the device is a namespace device (the controller is the "nvme device") and there can be multiple namespaces on one device. And we already have `NVDIMMNamespaceDevice` for nvdimms. But we are not going to have a device for the character device controller so there shouldn't be a conflict if we decide to name the namespaces simply `NVMeDevice`. 
Any thoughts about the naming here @tbzatek ?

tbzatek: From my point of view the plain `NVMeDevice` is somewhat misleading. While at the moment only namespaces are of interest, certain operations need to be performed on a controller device. Modelling both namespace and controller devices separately is more future-proof. Of course, depends on the future plans.

dwlehman: Sounds like there are several reasons to leave it as-is.

blivet/devices/disk.py
```diff
@@ -730,7 +730,7 @@ def sector_size(self):
         return self._sector_size
 
 
-NVMeController = namedtuple("NVMeController", ["name", "serial", "nvme_ver", "id", "subsysnqn"])
+NVMeController = namedtuple("NVMeController", ["name", "serial", "nvme_ver", "id", "subsysnqn", "transport"])
```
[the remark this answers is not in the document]
vojtechtrefny: Ok, done.

tbzatek (approved):
I think this is good to go, no more identifiers were requested so far.

See https://github.com/storaged-project/libblockdev/pull/827 for a related fix, not blocking this PR though.

PR #1078, opened 2022-10-19
Improve and document VM tests

berrange:
I've struggled to figure out how to actually run the VM tests in blivet since there's no docs.

This series attempts to help in that respect by adding a tests/vmtests/README.rst file, whose content I inferred / reverse engineered from terse description in

```
commit aa9160875482debdf6eacc0cbfa6ca1906cf5cb1
Author: Vojtech Trefny <vtrefny@redhat.com>
Date:   Thu Aug 25 11:08:42 2016 +0200

    Add a script to run tests in a virtual machine
    
    This script will start the machine, make a snapshot of its initial
    state and run the tests on it (reverting to the snashot after each
    test). The VM must automatically boot to an live image with ssh
    enabled.
```

With the steps I document, I was able to successfully run the VM tests. If there is a better way to achieve this, that the maintainers typically follow, then please let me know and I'll improve/change the docs correspondingly.

In figuring this out, I felt the runvmtests.py script would benefit from logging. To speed up debugging failed tests I also make it possible to run just one test and to avoid starting/stopping the test VM. Indeed avoiding starting/stopping is key with my documented instructions since the Live CD image I suggested doesn't have SSH enabled by default.

There was also a small bug in password arg handling for SSH

### DIFF
--- a/tests/vmtests/README.rst
+++ b/tests/vmtests/README.rst
@@ -1,0 +1,57 @@
+Running VM tests
+================
+
+The VM tests are intended to run aginst a libvirt managed KVM
+virtual machine, running a Live OS with two unprovisioned
+disks present. A suitable environment can be created using
+the virt-install program and Fedora Workstation Live CD
+media. As root follow the steps::
+
+  $ cd /var/lib/libvirt/images
+  $ wget https://fedora.mirrorservice.org/fedora/linux/releases/36/Workstation/x86_64/iso/Fedora-Workstation-Live-x86_64-36-1.5.iso
+  $ virt-install \
+      --cdrom /var/lib/libvirt/images/Fedora-Workstation-Live-x86_64-36-1.5.iso \
+      --name f36live \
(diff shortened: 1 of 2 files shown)


StorageGhoul:

vojtechtrefny (approved):
Thank you, this is a great improvement!

vojtechtrefny:
> I've struggled to figure out how to actually run the VM tests in blivet since there's no docs.

Looking back, we probably should've make this a separate test suite outside the repository because it always required to have a special setup to actually be able to run the test -- in our CI we actually crate a special LiveCD ISO using [livemedia-creator](http://weldr.io/lorax/livemedia-creator.html) for the test runs. We don't have many outside contributors here so a lot of things isn't well documented, so thank you again for making the documentation better and sorry you had to spend so much time on this.

vojtechtrefny:
Jenkins, ok to test.

PR #1080, opened 2022-11-03
Support configuring partition type UUIDs for GPT partition auto-discovery

berrange:
The discoverable partitions specification defines a way to identify partitions used for various well known mount points:

  https://uapi-group.org/specifications/specs/discoverable_partitions_specification/

This relies on setting well known GPT partition GUIDs, instead of parted's generic "Linux Data"  GUID which more or less everything gets set to today (swap being a notable exception).

My goal is to make it possible for applications like Anaconda to "do the right thing(tm)" with little to no effort when doing disk partitioning. This series aims to make that possible, while still retaining the current behaviour by default for the sake of backwards compatibility.

With this series, a simple app usage might not need anything more than to opt-in to discoverable partitions with:

```
from blivet.git import gpt_set_discoverable_partitions

gpt_set_discoverable_partitions(True)
```

Assuming the disk has 'gpt' disk label and the app is already passing 'mountpoint' to the ``blivet.new_partition()`` method then it will "just work" and pick the right GUID for the given mount point. The `mountpoint` parameter would historically only be needed if a format is requested with `new_partition`, but now it can be passed regardless so it can be forwarded onto the `PartitionDevice` constructor.

The `examples/partitioning.py` script illustrates this, as does a comprehensive test at ``tests/vmtests/gpt_test.py``

If the app is more advanced and creating `PartitionDevice` instances directly instead of via the convenient `blivet.py` API entrypoints, then it will need to be changed to pass `mountpoint` to this object constructor too, but this is still hopefully fairly straightforward. 

Note that this functionality relies on `pypartd` exposing the `type_uuid` attribute on the Partition object. This is new functionality added to pyparted last month in https://github.com/dcantrell/pyparted/pull/95  Note `pyparted` only exposes this if compiled against a new enough `parted`. As a result some code additions in this series have 'hasattr' changes to attempt to gracefully degrade when the functionality is missing.

Since this is my first time looking at Blivet code in any detail, its taken me a while to get my head around the design and expected usage patterns. I'm still not entirely confident in my understanding, but the `examples/partitioning.py` script is at least doing the right thing with this series of changes.

Happy to have feedback on things I've overlooked / approached in the wrong/suboptimal way.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -487,7 +487,7 @@ class Blivet(object):
         """
         if 'fmt_type' in kwargs:
             kwargs["fmt"] = get_format(kwargs.pop("fmt_type"),
-                                       mountpoint=kwargs.pop("mountpoint",
+                                       mountpoint=kwargs.get("mountpoint",
                                                              None),
                                        **kwargs.pop("fmt_args", {}))
 
--- a/blivet/deviceaction.py
+++ b/blivet/deviceaction.py
@@ -26,12 +26,13 @@ from six import add_metaclass
 
 from . import util
 from . import udev
-from .errors import DependencyError
+from .errors import DependencyError, PartitioningError
 from .util import get_current_entropy
 from .devices import StorageDevice
 from .devices import PartitionDevice, LVMLogicalVolumeDevice
 from .formats import get_format, luks
 from parted import partitionFlag, PARTITION_LBA
+import parted
 from .i18n import _, N_
 from .callbacks import CreateFormatPreData, CreateFormatPostData
 from .callbacks import ResizeFormatPreData, ResizeFormatPostData
@@ -635,6 +636,17 @@ class ActionCreateFormat(DeviceAction):
 
             if self.format.parted_flag is not None:
                 self.device.set_flag(self.format.parted_flag)
+
+            # If an explicit part type UUID is requested against
+            # the partition, that needs to take priority over any
+            # UUID assignment that was done implicitly when the
+            # above unset_flag/set_flag methods were called, or
+            # when the parted_partition.system flag is set.
+            if self.device.part_type_uuid_req is not None:
+                if not hasattr(parted.Partition, 'type_uuid'):
+                    raise PartitioningError("parted too old to set partition type UUID")
+
+                self.device.parted_partition.type_uuid = self.device.part_type_uuid_req.bytes
 
             self.device.disk.format.commit_to_disk()
             udev.settle()
--- a/blivet/devicelibs/gpt.py
+++ b/blivet/devicelibs/gpt.py
@@ -1,0 +1,251 @@
+# gpt.py
+# GPT partitioning helpers
+#
+# Copyright (C) Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions of
+# the GNU General Public License v.2, or (at your option) any later version.
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
+
+import uuid
+
+from blivet.arch import get_arch
+from blivet.errors import GPTVolUUIDError
+
+GPT_VOL_ARCH_ROOT = "root"
+GPT_VOL_ARCH_ROOT_VERITY = "root_verity"
+GPT_VOL_ARCH_ROOT_VERITY_SIG = "root_verity_sig"
+GPT_VOL_ARCH_USR = "usr"
+GPT_VOL_ARCH_USR_VERITY = "usr_verity"
+GPT_VOL_ARCH_USR_VERITY_SIG = "usr_verity_sig"
+
+GPT_VOL_ESP = "esp"
+GPT_VOL_XBOOTLDR = "xbootldr"
+GPT_VOL_SWAP = "swap"
+GPT_VOL_HOME = "home"
+GPT_VOL_SRV = "srv"
+GPT_VOL_VAR = "var"
+GPT_VOL_TMP = "tmp"
+GPT_VOL_USER_HOME = "user_home"
+GPT_VOL_LINUX_GENERIC = "linux_generic"
+
+_gpt_common_uuid = {
+    GPT_VOL_ESP: uuid.UUID("c12a7328-f81f-11d2-ba4b-00a0c93ec93b"),
+    GPT_VOL_XBOOTLDR: uuid.UUID("bc13c2ff-59e6-4262-a352-b275fd6f7172"),
+    GPT_VOL_SWAP: uuid.UUID("0657fd6d-a4ab-43c4-84e5-0933c84b4f4f"),
+    GPT_VOL_HOME: uuid.UUID("933ac7e1-2eb4-4f13-b844-0e14e2aef915"),
+    GPT_VOL_SRV: uuid.UUID("3b8f8425-20e0-4f3b-907f-1a25a76f98e8"),
+    GPT_VOL_VAR: uuid.UUID("4d21b016-b534-45c2-a9fb-5c16e091fd2d"),
+    GPT_VOL_TMP: uuid.UUID("7ec6f557-3bc5-4aca-b293-16ef5df639d1"),
+    GPT_VOL_USER_HOME: uuid.UUID("773f91ef-66d4-49b5-bd83-d683bf40ad16"),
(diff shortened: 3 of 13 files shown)


StorageGhoul:

dwlehman:
[Disclaimer: I'm not an active developer here anymore, so I may miss important things.]

Looks like a really nice addition to me. Your comprehension of the bits you're using here seems fine. The only thing I'm unsure of is whether it's necessary to set/manage `PartitionDevice.req_part_type_uuid` in addition to `PartitionDevice._part_type_uuid`; @vojtechtrefny will be able to say definitively. Generally, the `req_` attributes are for 'saving' data that will be stored in parted objects, which may get modified or replaced in the course of partition allocation; I think since you store `_part_type_uuid` outside of the parted object you can skip its `req_` counterpart.

Other than that, it might be good to extend the attribute checking to the places where you're setting it in `blivet/formats/disklabel.py` and in `blivet/deviceaction.py`

Thanks for your contribution!

vojtechtrefny:
Jenkins, ok to test.

blivet/gpt.py
```diff
@@ -0,0 +1,228 @@
...
+
+    remap = {
+        "amd64": "x86_64",
+        "i386": "i686",
+        "i486": "i686",
+        "i586": "i686",
+        "powerpc": "ppc",
+    }
+
+    if arch is None:
```
vojtechtrefny, line 212: You can use the `get_arch` function from the `blivet.arch` module here. (Probably more complicated than the `platform` module, but it's probably better to use something we already have.

blivet/gpt.py
```diff
@@ -0,0 +1,228 @@
...
+# ANY WARRANTY expressed or implied, including the implied warranties of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.  You should have received a copy of the
+# GNU General Public License along with this program; if not, write to the
+# Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.  Any Red Hat trademarks that are incorporated in the
+# source code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission of
+# Red Hat, Inc.
+#
```
vojtechtrefny, line 19: The better place for this file would probably be `blivet/devicelibs`. Strictly speaking GPT is not a device, but this where we already have some definitions for LUKS or btrfs too.

blivet/gpt.py
```diff
@@ -0,0 +1,228 @@
...
+        "parisc": uuid.UUID("450dd7d1-3224-45ec-9cf2-a43a346d71ee"),
+        "ppc": uuid.UUID("7007891d-d371-4a80-86a4-5cb875b9302e"),
+        "ppc64": uuid.UUID("0b888863-d7f8-4d9e-9766-239fce4d58af"),
+        "ppc64el": uuid.UUID("c8bfbd1e-268e-4521-8bba-bf314c399557"),
+        "riscv32": uuid.UUID("c3836a13-3137-45ba-b583-b16c50fe5eb4"),
+        "riscv64": uuid.UUID("d2f9000a-7a18-453f-b5cd-4d32f77a7b32"),
+        "s390": uuid.UUID("17440e4f-a8d0-467f-a46e-3912ae6ef2c5"),
+        "s390x": uuid.UUID("3f324816-667b-46ae-86ee-9b0c0c6c11b4"),
+        "tilegx": uuid.UUID("4ede75e2-6ccc-4cc8-b9c7-70334b087510"),
+        "x86_64": uuid.UUID("e7bb33fb-06cf-4e81-8273-e543b413e2e2"),
```
vojtechtrefny, line 181: This is more a note to myself: we should move these definitions to libblockdev in the future (possibly using the GUID list provided by libfdisk). We had similar discussion recently about these definitions in UDisks https://github.com/storaged-project/udisks/pull/1032 and it makes sense to have these in one place if possible. (But that's out of scope for this PR.)

blivet/gpt.py
```diff
@@ -183,6 +183,18 @@
 }
 
 
+_gpt_discoverable_partitions = False
```
vojtechtrefny, line 186: Please use `blivet.flags.Flags` here instead. We already have multiple flags for controlling various default behaviours, so it make more sense having this there too. Anaconda is also setting these flags, so it will make it easier for the installer too.

vojtechtrefny (changes requested):
Looks good to me in general, I have just few nitpicks. Thank you for the contribution, it is a great improvement, this feature has been requested for years.

I agree with Dave about the `req_part_type_uuid`, which is probably not necessary (but we can keep it, if you think it's better or clearer that way). I also plan to do some manual testing, but everything looks to be covered by tests so I don't expect any problems.

blivet/gpt.py
berrange: I'll have a look and see if it has the right semantics

blivet/gpt.py
berrange: Sure, works for me.

blivet/gpt.py
berrange: Yeah, it would make sense to have them in a place that can maximise sharing.

blivet/gpt.py
berrange: Ah yes, that's a much better place for this.

dwlehman:
> WRT tracking req_part_type_uuid directly, I was having some trouble distinguishing the user specified UUIDs from the default UUIDs that parted sets. Essentially every parted partition reports a UUID all the time. What's even more complex is that while we set a custom UUID when creating the partition, when we later format the partition with a filesystem, parted will replace the UUID with a default one again. I'll have another look at this all though, to see if there's a saner way.

We aren't using parted to create file systems, so I don't think parted would have an opportunity to change the UUID at this point. It might be a bug in parted that it's doing that when you use it to create the file system.

berrange:
> We aren't using parted to create file systems, so I don't think parted would have an opportunity to change the UUID at this point. It might be a bug in parted that it's doing that when you use it to create the file system.

I presume you're referring to Anaconda in this context ?  Blivet at least has functionality for formatting filesystems using parted, which is where I was hitting the problems in testing. I've not actually tried anything with anaconda yet.

dwlehman:
> > We aren't using parted to create file systems, so I don't think parted would have an opportunity to change the UUID at this point. It might be a bug in parted that it's doing that when you use it to create the file system.
> 
> I presume you're referring to Anaconda in this context ? Blivet at least has functionality for formatting filesystems using parted, which is where I was hitting the problems in testing. I've not actually tried anything with anaconda yet.

Wow, I forgot all about that. it's in `ActionCreateFormat.execute`, just above where you are setting the UUID. We aren't using parted to create the file system, but we are telling it to take note of the file system type, which I forgot about. Maybe the best way is to expand the conditional that sets `self.device.parted_partition.system` so that it handles both cases in one place.

berrange:
> Wow, I forgot all about that. it's in `ActionCreateFormat.execute`, just above where you are setting the UUID. We aren't using parted to create the file system, but we are telling it to take note of the file system type, which I forgot about. Maybe the best way is to expand the conditional that sets `self.device.parted_partition.system` so that it handles both cases in one place.

The `ActionCreateFormat.execute` code is affecting UUID in several ways. The `self.device.parted_partition.system` property will change it, as you mention, but also calling `self.device.unset_flag` or `self.device.set_flag` methods will also change the UUID depending on which flag is touched, as some of the flags have associated UUID mappings defined in parted's `flag_uuid_mapping` table (see https://git.savannah.gnu.org/cgit/parted.git/tree/libparted/labels/gpt.c#n179).

I didn't feel it was sensible to try to second guess which flags might have side effects on the partition type UUID. So took the approach of just explicitly setting the partition type UUID at the end after everything else has been done, which should be reliable and clearer to understand.

berrange:
I've pushed an updated series

 * switched to blivet.get_arch  and use blivet arch names in GPT UUID lookup tables
 * Use blivet.flags for the global toggle for use of discoverable partitions
 * moved the gpt.py module under blivet.devicelibs
 * Made more use of hasattr() and raise exceptions if parted doesn't support setting the type UUID

berrange:
Also, I switched the base to be the 3.7-devel branch

berrange:
The Fedora 36 jenkins job failure looks like it might be an infra problem, if someone can re-trigger it

vojtechtrefny (approved):
Thank you, looks good to me.

vojtechtrefny:
> > Wow, I forgot all about that. it's in `ActionCreateFormat.execute`, just above where you are setting the UUID. We aren't using parted to create the file system, but we are telling it to take note of the file system type, which I forgot about. Maybe the best way is to expand the conditional that sets `self.device.parted_partition.system` so that it handles both cases in one place.
> 
> The `ActionCreateFormat.execute` code is affecting UUID in several ways. The `self.device.parted_partition.system` property will change it, as you mention, but also calling `self.device.unset_flag` or `self.device.set_flag` methods will also change the UUID depending on which flag is touched, as some of the flags have associated UUID mappings defined in parted's `flag_uuid_mapping` table (see https://git.savannah.gnu.org/cgit/parted.git/tree/libparted/labels/gpt.c#n179).
> 
> I didn't feel it was sensible to try to second guess which flags might have side effects on the partition type UUID. So took the approach of just explicitly setting the partition type UUID at the end after everything else has been done, which should be reliable and clearer to understand.

I think your solution to set the GUID after the flags is fine. The parted flags that are sometimes flags, sometimes GPT attributes, sometimes partition GUIDs or IDs were always confusing so setting the UUID after the flags should be safe.

vojtechtrefny:
> Also, I switched the base to be the 3.7-devel branch

Thanks. I have few more changes planned for the 3.7 release, but if everything goes as planned, it should be released in time for Fedora 38 beta.

PR #1085, opened 2022-11-22
ci: Run Anaconda tests correctly on rhel9-branch

vojtechtrefny:
So we can avoid issues like https://github.com/rhinstaller/anaconda/pull/4430 in the future when making changes in blivet.

### DIFF
--- a/.github/workflows/anaconda_tests.yml
+++ b/.github/workflows/anaconda_tests.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
     env:
-      TARGET_BRANCH: 'master'
+      TARGET_BRANCH: 'rhel-9'
 
     steps:
       - name: Clone Anaconda repository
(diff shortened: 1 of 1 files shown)


.github/workflows/anaconda_tests.yml
```diff
@@ -43,10 +43,11 @@ jobs:
           # - Install dependencies for blivet.
           # - Install blivet to the container.
           # - Run Anaconda tests.
-          podman run -i --rm -v ./blivet:/blivet:z -v ./anaconda:/anaconda:z quay.io/rhinstaller/anaconda-ci:$TARGET_BRANCH sh -c " \
+          podman run -i --rm -v ./blivet:/blivet:z -v ./anaconda:/anaconda:z quay.io/rhinstaller/anaconda-ci:master sh -c " \
```
VladimirSlavik, line 46: I'm afraid I don't understand why master, if this is for rhel 9?

vojtechtrefny: You don't have RHEL images (or at least I don't see one on quay). If this isn't possible, I can close the PR, I was just trying to make sure we won't break Anaconda in the future on RHEL.

PR #1090, opened 2022-11-29
Do not report as `Mactel` for Apple Macs with the T2 chip

sharpenedblade:
T2 apple macs behave much more like normal `EFI` computers than the current use of `Mactel`. This PR will make T2 Apple Macs report as x86-64 EFI computers, which is probably the right behavior. This PR also adds the `is_t2mac()` function to arch.py for detecting T2 macs.

### DIFF
--- a/blivet/arch.py
+++ b/blivet/arch.py
@@ -232,6 +232,8 @@ def is_mactel():
         mactel = False
     elif not os.path.isfile(DMI_CHASSIS_VENDOR):
         mactel = False
+    elif is_t2mac():
+        mactel = False
     else:
         try:
             buf = open(DMI_CHASSIS_VENDOR).read()
--- a/blivet/arch.py
+++ b/blivet/arch.py
@@ -211,7 +211,7 @@ def is_t2mac():
         "MacBookAir8,1", "MacBookPro16,4", "MacBookPro16,3", "MacBookPro16,2",
         "MacBookPro16,1", "MacBookPro15,4", "MacBookPro15,3", "MacBookPro15,2",
         "MacBookPro15,1",
-        }
+    }
     if not is_x86():
         t2_mac = False
     elif not os.path.isfile(DMI_PRODUCT_ID):
--- a/blivet/arch.py
+++ b/blivet/arch.py
@@ -40,6 +40,7 @@ log = logging.getLogger("blivet")
 
 # DMI information paths
 DMI_CHASSIS_VENDOR = "/sys/class/dmi/id/chassis_vendor"
+DMI_PRODUCT_ID = "/sys/class/dmi/id/product_id"
 
 
 def get_ppc_machine():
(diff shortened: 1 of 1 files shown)


StorageGhoul:

sharpenedblade:
This also solves [https://bugzilla.redhat.com/show_bug.cgi?id=2140462](https://bugzilla.redhat.com/show_bug.cgi?id=2140462)

tbzatek:
Yeah, this looks good enough for T2-based Macs. Could you please add similar change for the T1 chip? I've had `MacBookPro14,1` that came with regular FAT32 EFI partition, I put my Gentoo rootfs in there right away and it worked just fine.

vojtechtrefny:
Jenkins, ok to test.

bcl:
The problem with how this is implemented is that any new T1 or T2 system will fall back to being treated like an old mactel system. It would be better if it detected the old ones and defaulted to the new behavior on newer systems.

sharpenedblade:
> The problem with how this is implemented is that any new T1 or T2 system will fall back to being treated like an old mactel system. It would be better if it detected the old ones and defaulted to the new behavior on newer systems.

New T2 macs will be treated like normal EFI computers, which is almost always the right thing to do. The code detects new macs, and leaves old macs alone. Normally this would be a bad approach, but Apple no longer makes T2 intel macs, so our list does not need to be updated.

sharpenedblade:
> Yeah, this looks good enough for T2-based Macs. Could you please add similar change for the T1 chip? I've had `MacBookPro14,1` that came with regular FAT32 EFI partition, I put my Gentoo rootfs in there right away and it worked just fine.

I could not find a list of T1 macs, but if there is a list it is just a matter of adding the DMI ids to the set.

blivet/arch.py
```diff
@@ -198,6 +199,29 @@ def is_cell():
...
+    :return: True if the hardware is an Intel-based Apple Mac with the T2 security chip, False otherwise.
+    :rtype boolean
+
+    """
+    APPLE_T2_PRODUCT_NAMES = {
+        "iMac20,2", "iMac20,1", "iMacPro1,1",
+        "MacPro7,1", "Macmini8,1", "MacBookAir9,1", "MacBookAir8,2",
+        "MacBookAir8,1", "MacBookPro16,4", "MacBookPro16,3", "MacBookPro16,2",
+        "MacBookPro16,1", "MacBookPro15,4", "MacBookPro15,3", "MacBookPro15,2",
+        "MacBookPro15,1",
```
vojtechtrefny, line 213: I didn't really like the idea of having a list of products here at first, but it's short enough and if Apple isn't really going to make more Intel Macs, I think it should be ok.

blivet/arch.py
```diff
@@ -198,6 +199,29 @@ def is_cell():
...
+    }
+    if not is_x86():
+        t2_mac = False
+    elif not os.path.isfile(DMI_PRODUCT_ID):
+        t2_mac = False
+    else:
+        buf = open(DMI_PRODUCT_ID).read()
```
vojtechtrefny, line 220: Leaking file descriptor.

vojtechtrefny (approved):
Looks good and simple enough. Thank you!

blivet/arch.py
vojtechtrefny: Ok, we do exactly the same below with the `DMI_CHASSIS_VENDOR`. I guess I can fix both later in a separate PR.

tbzatek:
> EDIT: After some research, I think that T1 macs should continue to use the old `Mactel` system.

Fair enough, I guess the secure boot enforcement the original bugreport was about works differently than on the T2 platforms.

blivet/arch.py
sharpenedblade: Apple is not going to make more Intel macs, so this list should never expand. The other method would be to look for a PCI device that represents the T2, but it would have more complicated code, which I think would lead to more bugs.

blivet/arch.py
sharpenedblade: Should I leave this as is.

sharpenedblade:
> > EDIT: After some research, I think that T1 macs should continue to use the old `Mactel` system.
> 
> Fair enough, I guess the secure boot enforcement the original bugreport was about works differently than on the T2 platforms.

Its not just that, the EFI in T1 macs will only properly honor some boot icon/name stuff on HFS+ partitions.

blivet/arch.py
vojtechtrefny: Yes, I will fix this later, we have this problem at more places. I want to give some more time to other people to review this, but other than that, this is ready to be merged. Thank you again.

PR #1091, opened 2022-12-02
Make sure we close the streams when reading a file

vojtechtrefny:
Python's garbage collector should close these automatically, but we shouldn't depend on that and close the file descriptor when no longer needed.

### DIFF
--- a/blivet/util.py
+++ b/blivet/util.py
@@ -280,17 +280,17 @@ def get_mount_paths(dev):
 def get_mount_device(mountpoint):
     """ Given a mountpoint, return the device node path mounted there. """
     mountpoint = os.path.realpath(mountpoint)  # eliminate symlinks
-    mounts = open("/proc/mounts").readlines()
     mount_device = None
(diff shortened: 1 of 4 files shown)


blivet/util.py
```diff
@@ -280,17 +280,17 @@ def get_mount_paths(dev):
...
+                (device, path, _rest) = mnt.split(None, 2)
+            except ValueError:
+                continue
 
-        if path == mountpoint:
-            mount_device = device
-            break
+            if path == mountpoint:
+                mount_device = device
+                break
```
VladimirSlavik, line 293: I would expect this fix to be a separate commit.

VladimirSlavik (approved):
LGTM, thanks.

Not a team member of course.

blivet/util.py
vojtechtrefny: This isn't a separate change, I am just using the `open` contextmanager here so I needed to indent this block. I guess it's just git diff showing it in a weird way.

PR #1093, opened 2022-12-14
Fstab support

japokorn:
- added support for blivet to directly modify fstab based on blivet actions
 - added tests

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -40,6 +40,7 @@ from .devicelibs.crypto import LUKS_METADATA_SIZE
 from .errors import StorageError, DependencyError
 from .size import Size
 from .devicetree import DeviceTree
+from .fstab import FSTabManager
 from .formats import get_default_filesystem_type
 from .flags import flags
 from .formats import get_format
@@ -70,6 +71,10 @@ class Blivet(object):
 
         self.size_sets = []
         self.set_default_fstype(get_default_filesystem_type())
+
+        # fstab write location purposedly set to None. It has to be overriden
+        # manually when using blivet.
+        self.fstab = FSTabManager(src_file="/etc/fstab", dest_file=None)
 
         self._short_product_name = 'blivet'
 
@@ -111,7 +116,8 @@ class Blivet(object):
 
         """
 
-        self.devicetree.actions.process(callbacks=callbacks, devices=self.devices)
+        self.devicetree.actions.process(callbacks=callbacks, devices=self.devices, fstab=self.fstab)
+        self.fstab.read()
 
     @property
     def next_id(self):
@@ -140,6 +146,8 @@ class Blivet(object):
         self.devicetree.populate(cleanup_only=cleanup_only)
         self.edd_dict = get_edd_dict(self.partitioned)
         self.devicetree.edd_dict = self.edd_dict
+
+        self.fstab.read()
 
         if flags.include_nodev:
             self.devicetree.handle_nodev_filesystems()
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -88,7 +88,8 @@ class FS(DeviceFormat):
     # value is already unpredictable and can change in the future...
     _metadata_size_factor = 1.0
 
-    config_actions_map = {"label": "write_label"}
+    config_actions_map = {"label": "write_label",
+                          "mountpoint": "change_mountpoint"}
 
     def __init__(self, **kwargs):
         """
@@ -626,6 +627,11 @@ class FS(DeviceFormat):
             self._chrooted_mountpoint = None
 
         udev.settle()
+
+    def change_mountpoint(self, dry_run=False):
+        # This function is intentionally left blank. Mountpoint change utilizes
+        # only the generic part of this code branch
+        pass
 
     def read_label(self):
         """Read this filesystem's label.
--- a/blivet/fstab.py
+++ b/blivet/fstab.py
@@ -1,0 +1,364 @@
+import os
+
+import gi
+
+gi.require_version("BlockDev", "2.0")
+from gi.repository import BlockDev as blockdev
+
+from libmount import Table, Fs, MNT_ITER_FORWARD
+from libmount import Error as LibmountException
+
+import logging
+log = logging.getLogger("blivet")
+
+
+def parser_errcb(tb, fname, line):  # pylint: disable=unused-argument
+    print("{:s}:{:d}: parse error".format(fname, line))
+    return 1
+
+
+class FSTabManager():
+    # Read, write and modify fstab file
+    # This class is meant to work even without blivet.
+    # However some of its methods require blivet and will not function without
+    # it. These methods will import what they need when they are run.
+
+    def __init__(self, src_file=None, dest_file=None):
+        self._table = Table()   # Space for parsed fstab contents
+
+        self.src_file = src_file
+        self.dest_file = dest_file
+
+        if self.src_file is not None:
+            self.read(self.src_file)
+
+    def __deepcopy__(self, memo):
+        clone = FSTabManager(src_file=self.src_file, dest_file=self.dest_file)
+        clone._table = Table()
+        clone._table.enable_comments(True)
+
+        entry = self._table.next_fs()
+        entries = [entry]
+        while entry:
+            entries.append(entry)
+            entry = self._table.next_fs()
+
+        # Libmount does not allow to simply use clone._table.add_fs(entry), so...
+        for entry in entries:
+            new_entry = Fs()
+            new_entry.source = entry.source
+            new_entry.target = entry.target
+            new_entry.fstype = entry.fstype
+            new_entry.append_options(entry.options)
+            new_entry.freq = entry.freq
+            new_entry.passno = entry.passno
+            if entry.comment is not None:
+                new_entry.comment = entry.comment
+            clone._table.add_fs(new_entry)
+
+        return clone
+
+    def _get_containing_device(self, path, devicetree):
+        # Return the device that a path resides on
+        if not os.path.exists(path):
+            return None
+
+        st = os.stat(path)
+        major = os.major(st.st_dev)
+        minor = os.minor(st.st_dev)
+        link = "/sys/dev/block/%s:%s" % (major, minor)
+        if not os.path.exists(link):
+            return None
+
+        try:
+            device_name = os.path.basename(os.readlink(link))
+        except Exception:  # pylint: disable=broad-except
+            log.error("failed to find device name for path %s", path)
+            return None
+
+        if device_name.startswith("dm-"):
+            # have I told you lately that I love you, device-mapper?
+            # (code and comment copied from anaconda, kept for entertaining purposes)
+            device_name = blockdev.dm.name_from_node(device_name)
+
+        return devicetree.get_device_by_name(device_name)
+
+    def _from_device(self, device):
+        # Return the list of fstab options obtained from device
+        # *(result) of this method is meant to be used as a parameter in other methods
+
+        spec = getattr(device, 'fstab_spec', None)
+
+        file = None
+        if device.format.mountable:
+            file = device.format.mountpoint
+        if device.format.type == 'swap':
+            file = 'swap'
+
+        vfstype = device.format.type
+        mntops = None
+
+        return spec, file, vfstype, mntops
+
+    def _from_entry(self, entry):
+        return entry.source, entry.target, entry.fstype, ','.join(entry.options)
+
+    def read(self, src_file=''):
+        # Read fstab file
+
+        # Reset table
+        self._table = Table()
+        self._table.enable_comments(True)
+
+        # resolve which file to read
+        if src_file == '':
+            if self.src_file is None:
+                # No parameter given, no internal value
+                return
+        elif src_file is None:
+            return
+        else:
+            self.src_file = src_file
+
+        self._table.errcb = parser_errcb
+        self._table.parse_fstab(self.src_file)
+
+    def find_device_by_specs(self, blivet, spec, file_dummy=None, vfstype_dummy=None, mntops=""):  # pylint: disable=unused-argument
+        # Parse an fstab entry for a device, return the corresponding device,
+        # return None if not found
+        # dummy arguments allow using result of _from_device/_from_entry as a parameter of this method
+        return blivet.devicetree.resolve_device(spec, options=mntops)
+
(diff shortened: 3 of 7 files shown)


vojtechtrefny:
Moving to draft, because this is clearly not yet ready for a review. The core functionality is missing and the fstab module itself has some serious issues reported by pylint:

```
E0602(undefined-variable):blivet/fstab.py:40,12: FSTabManager._get_containing_device: Undefined variable 'log_exception_info'
E0602(undefined-variable):blivet/fstab.py:46,26: FSTabManager._get_containing_device: Undefined variable 'blockdev'
E0602(undefined-variable):blivet/fstab.py:87,21: FSTabManager.get_device: Undefined variable 'NFSDevice'
E0602(undefined-variable):blivet/fstab.py:120,24: FSTabManager.get_device: Undefined variable 'get_device_format_class'
E0602(undefined-variable):blivet/fstab.py:126,12: FSTabManager.get_device: Undefined variable 'log'
E0602(undefined-variable):blivet/fstab.py:133,12: FSTabManager.get_device: Undefined variable 'log'
E0602(undefined-variable):blivet/fstab.py:143,12: FSTabManager.get_device: Undefined variable 'log'
E0602(undefined-variable):blivet/fstab.py:148,33: FSTabManager.get_device: Undefined variable '_'
E0203(access-member-before-definition):blivet/fstab.py:59,15: FSTabManager.read: Access to member '_src_file' before its definition line 63
```

blivet/formats/fs.py
```diff
@@ -627,6 +628,9 @@ def _teardown(self, **kwargs):
 
         udev.settle()
 
+    def change_mountpoint(self, dry_run=False):
```
vojtechtrefny, line 631: The support for changing mountpoint with `ActionConfigureFormat` can be a separate commit, but it should be finished before the PR is set to be ready to be merged.

blivet/fstab.py
```diff
@@ -0,0 +1,279 @@
...
+        except Exception:  # pylint: disable=broad-except
+            log_exception_info(fmt_str="failed to find device name for path %s", fmt_args=[path])
+            return None
+
+        if device_name.startswith("dm-"):
+            # have I told you lately that I love you, device-mapper?
+            # (code and comment copied from anaconda, kept for entertaining purposes)
+            device_name = blockdev.dm.name_from_node(device_name)
+
+        return devicetree.get_device_by_name(device_name)
```
vojtechtrefny, line 48: So the module will always require a populated devicetree to work? If that is true, we might be able to simplify some of the things above with something like `resolve_device`.

blivet/fstab.py
```diff
@@ -0,0 +1,279 @@
...
+                    "There is an entry in your /etc/fstab file that contains "
+                    "an invalid or incorrect file system type. The file says that "
+                    "{detected_type} at {mount_point} is {fstab_type}.").format(
+                    detected_type=dtype,
+                    mount_point=file,
+                    fstab_type=ftype
+                ))
+
+        del ftype
+        del dtype
```
vojtechtrefny, line 158: Is this necessary? I would expect Python to deal correctly with the memory management here. If not, a comment might be a good idea.

blivet/fstab.py
```diff
@@ -0,0 +1,279 @@
...
+        # the device's format we found matches what's in the fstab
+        ftype = getattr(fmt, "mount_type", fmt.type)
+        dtype = getattr(device.format, "mount_type", device.format.type)
+        if hasattr(fmt, "test_mount") and vfstype != "auto" and ftype != dtype:
+            log.info("fstab says %s at %s is %s", dtype, file, ftype)
+            if fmt.test_mount():     # pylint: disable=no-member
+                device.format = fmt
+            else:
+                device.teardown()
+                raise ValueError(_(
```
vojtechtrefny, line 148: If this is a pre-existing entry, we might not want to raise an error here, it's probably not our job to tell users their fstab is wrong (unless they ask us explicitly to write something we think is wrong), especially because it simply might be correct and we just don't understand/support the filesystem.

tests/fstab_test.py
```diff
@@ -0,0 +1,78 @@
+import os
+
+from storage_tests.storagetestcase import StorageTestCase
+
+import blivet
+
+FSTAB_WRITE_FILE = "/var/tmp/test-blivet-fstab"
```
vojtechtrefny, line 7: `/var` is persistent, probably not a best place to add temporary files. Using the `tempfile` module to create a temporary directory would be better.

tests/fstab_test.py
```diff
@@ -0,0 +1,78 @@
+import os
```
vojtechtrefny, line 1: The test needs to be placed either in `storage_tests` (if it uses "real" storage) or in `unit_tests` (if it doesn't require root privileges). Ideally this functionality would be covered in both test suites.

vojtechtrefny (changes requested):

blivet/fstab.py
```diff
@@ -0,0 +1,279 @@
...
+            fstab_entries.append(entry)
+            entry = self._table.next_fs()
+
+        for device in devices:
+            args = self.from_device(device)
+            entry = self.find(*args)
+
+            # remove from the list if found
+            try:
+                fstab_entries.remove(entry)
```
vojtechtrefny, line 265: I might be missing something here, but `fstab_entries` isn't used after this so this part actually doesn't remove the devices from the table?

japokorn: Fixed.

japokorn:
[citest]

japokorn:
test this please

blivet/formats/fs.py
japokorn: Added an explanation comment to make the intention clearer.

tests/fstab_test.py
japokorn: Changed path to /tmp/...
I am not using `tempfile` on purpose. Created file has to be closed before blivet modifies it. Closing the `tempfile` will immediately remove it.

blivet/fstab.py
japokorn: Removed.

japokorn: Whole method was copied from anaconda. Removing `del`s does not seem to affect the behavior. Removed.

tests/fstab_test.py
japokorn: The tests are now placed in `storage_tests` and `unit_tests` .

blivet/fstab.py
japokorn: It might not be that clear from anaconda perspective. I have added a new method called `find_device` that utilizes `resolve_device` and does not create the device if it does not exist.

blivet/fstab.py
japokorn: Changed Exception to log.info, so the unusual state is logged. If that happens the method will return None instead of a device.

tests/fstab_test.py
vojtechtrefny: The tempfile library also allows creating temporary directories, see https://docs.python.org/3.8/library/tempfile.html#tempfile.TemporaryDirectory I'd prefer using that rather than a custom cleanup.

vojtechtrefny (changes requested):
Please also rebase the PR to the latest 3.8-devel branch.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+            fmt = get_format(vfstype)
+            fmt_class = get_device_format_class("nodev")
+            if spec == "none" or \
+               (fmt_class and isinstance(fmt, fmt_class)):
+                device = NoDevice(fmt=fmt)
+
+        if device is None:
+            log.error("failed to resolve %s (%s) from fstab", spec,
+                      vfstype)
+            raise ValueError()
```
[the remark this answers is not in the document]
japokorn: Fixed.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+            freq = 0
+
+        # Order of fsck run at boot time. '/' should have 1, other 2, defaults to 0
+        if passno is None:
+            if file is None:
+                file = ''
+            if file == '/':
+                passno = 1
+            elif file.startswith('/boot'):
+                passno = 2
```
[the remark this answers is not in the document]
japokorn: '/' should have 1, other checked filesystems, like '/boot' should have 2. Defaults to 0 (no filesystem check).
Changed the comment to clarify.

blivet/blivet.py
```diff
@@ -71,6 +72,10 @@ def __init__(self):
         self.size_sets = []
         self.set_default_fstype(get_default_filesystem_type())
 
+        # fstab write location purposedly set to none. It has to be overriden
+        # manually when using blivet.
+        self.fstab = FSTabManager(src_file="/etc/fstab", dest_file=None)
```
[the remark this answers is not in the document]
japokorn: Replaced the string with the constant.

blivet/fstab.py
```diff
@@ -0,0 +1,367 @@
...
+
+from libmount import Table, Fs, MNT_ITER_FORWARD
+from libmount import Error as LibmountException
+
+import logging
+log = logging.getLogger("blivet")
+
+
+def parser_errcb(tb, fname, line):  # pylint: disable=unused-argument
+    print("{:s}:{:d}: parse error".format(fname, line))
```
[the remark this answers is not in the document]
japokorn: Fixed.

blivet/fstab.py
```diff
@@ -0,0 +1,367 @@
...
+    # However some of its methods require blivet and will not function without
+    # it. These methods will import what they need when they are run.
+
+    def __init__(self, src_file=None, dest_file=None):
+        self._table = Table()   # Space for parsed fstab contents
+
+        self.src_file = src_file
+
+        self.dest_file = dest_file
+        if dest_file is None:
```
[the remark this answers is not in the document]
japokorn: Fixed.

blivet/fstab.py
```diff
@@ -0,0 +1,367 @@
...
+log = logging.getLogger("blivet")
+
+
+def parser_errcb(tb, fname, line):  # pylint: disable=unused-argument
+    print("{:s}:{:d}: parse error".format(fname, line))
+    return 1
+
+
+class FSTabManager():
+    # Read, write and modify fstab file
```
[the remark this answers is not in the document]
japokorn: Changed comments to docstrings.

tests/fstab_test.py
japokorn: Changed to TemporaryDirectory.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def get_device_by_specs(self, blivet, spec, file, vfstype, mntops):
+        # Parse an fstab entry for a device, return the corresponding device,
+        # create new one if it does not exist
+
+        from blivet.formats import get_format
+        from blivet.devices import DirectoryDevice, NFSDevice, FileDevice, NoDevice
+        from blivet.formats import get_device_format_class
+
+        # no sense in doing any legwork for a noauto entry
```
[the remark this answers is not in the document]
japokorn: Anaconda used this for skipping processing of devices that wouldn't be mounted. I don't think it makes sense for us to keeping it here. Removed.

tests/storage_tests/fstab_test.py
```diff
@@ -0,0 +1,75 @@
+import os
+
+from .storagetestcase import StorageTestCase
+
+import blivet
+
+FSTAB_WRITE_FILE = "/tmp/test-blivet-fstab"
+
+
+class FstabTestCase(StorageTestCase):
```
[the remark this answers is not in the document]
japokorn: Creating devices is already part of the tests. I have added testing of removing as well.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+        # dummy arguments allow using result of _from_device/_from_entry as a parameter of this method
+        return blivet.devicetree.resolve_device(spec, options=mntops)
+
+    def find_device_by_entry(self, blivet, entry):
+        args = self._from_entry(entry)
+        return self.find_device_by_specs(blivet, *args)
+
+    def get_device_by_specs(self, blivet, spec, file, vfstype, mntops):
+        # Parse an fstab entry for a device, return the corresponding device,
+        # create new one if it does not exist
```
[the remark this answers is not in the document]
japokorn: I was not able to find any reason why these types of devices should be there, so I dropped them.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+                return
+        elif src_file is None:
+            return
+        else:
+            self.src_file = src_file
+
+        self._table.errcb = parser_errcb
+        self._table.parse_fstab(self.src_file)
+
+    def find_device_by_specs(self, blivet, spec, file_dummy=None, vfstype_dummy=None, mntops=""):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
japokorn: I have reworked the whole API, to better cover up internal libmount use.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+    print("{:s}:{:d}: parse error".format(fname, line))
+    return 1
+
+
+class FSTabManager():
+    # Read, write and modify fstab file
+    # This class is meant to work even without blivet.
+    # However some of its methods require blivet and will not function without
+    # it. These methods will import what they need when they are run.
+
```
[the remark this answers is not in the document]
japokorn: I have added methods for better access. It is now possible to iterate over fstab entries, print, and compare them.

blivet/fstab.py
```diff
@@ -0,0 +1,364 @@
...
+            if spec == "none" or \
+               (fmt_class and isinstance(fmt, fmt_class)):
+                device = NoDevice(fmt=fmt)
+
+        if device is None:
+            log.error("failed to resolve %s (%s) from fstab", spec,
+                      vfstype)
+            raise ValueError()
+
+        device.setup()
```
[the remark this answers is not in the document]
japokorn: Removed.

blivet/formats/fs.py
japokorn: The point of this is to be able to revert any mountpoint changes in case the action gets cancelled. Since it is just a property value change, it does not need any code.
I am open to suggestions but this came to me as a simple solution that utilizes structures already in place.

blivet/fstab.py
japokorn: I have reworked the whole method, dropping some stuff that is probably not needed.

PR #1095, opened 2023-01-10
Use DMI product_name for t2 mac detection.

sharpenedblade:
`/sys/class/dmi/id/product_name` works on more model/firmware combinations than DMI `product_id`. Apple's firmware always sets `product_name`, but only sometimes sets `product_id`, so the t2 detection will work on more macs with this change.


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

PR #1096, opened 2023-01-12
Default to encryption sector size 512 for LUKS devices

vojtechtrefny:
We are currently letting cryptsetup decide the optimal encryption sector size for LUKS. The problem is that for disks with physical sector size 4096 cryptsetup will default to 4096 encryption sector size even if the drive logical sector size is 512 which means these disks cannot be combined with other 512 logical sector size disks in LVM. This requires a more sophisticated solution in the future, but for now just default to 512 if not specified by the user otherwise.

Resolves: rhbz#2103800


vojtechtrefny:
Merged manually.

PR #1098, opened 2023-01-18
Remove support for DMRAID devices 

vojtechtrefny:
This is planned for Fedora 38: https://fedoraproject.org/wiki/Changes/UseMdadmForBIOSRAIDInAnaconda


vojtechtrefny:
I have successfully finished installation with this patch (with an extra patch for mdadm udev rules and only on Fedora 37, because the array cannot be assembled by mdadm with kernel 6.2 from rawhide) so I am marking this as ready for review.

PR #1100, opened 2023-01-26
Ignore parent dependencies during action execute

vojtechtrefny:
We don't need full support for the entire stack when creating a device, having dependencies just for the device the action is working with should be enough.

Resolves: rhbz#2144812


dwlehman (commented):
Looks good to me.

PR #1103, opened 2023-02-07
Ignore missing parted disk in ActionList._post_process (#2102960)

vojtechtrefny:
If we don't have a parted disk instance here, we should just ignore it some of the devices we are trying to update here are not even used.


vojtechtrefny:
Jenkins, test this please.

PR #1106, opened 2023-02-15
Remove support for Python 2 from spec and Makefile

vojtechtrefny:
Python 2 EOL was 3 years ago so we can stop supporting it now.

I'd like to do a bigger cleanup in the future (like removing the `six` module), but for now we can just remove Python 2 from the build and release process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PYTHON?=python3
+PYTHON=python3
 PKG_INSTALL?=dnf
 
 L10N_REPOSITORY=git@github.com:storaged-project/blivet-weblate.git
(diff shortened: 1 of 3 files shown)


vojtechtrefny:
Jenkins, test this please.

japokorn:
Shouldn't the line `Conflicts: python-blivet < 1:2.0.0` in the spec file be removed as well?
Otherwise it looks good to me.

vojtechtrefny:
> Shouldn't the line `Conflicts: python-blivet < 1:2.0.0` in the spec file be removed as well?

We probably could remove all the `Conflicts` statements simply because 2.0 release was really long time ago. But the conflict is valid for the Python 2 package here too -- 2.0 is where we split the package into two (python-blivet and blivet-data) and even the old Python 2 package would conflict with our new blivet-data package because it contains (some of) the same files.

japokorn (approved):
Looks good to me.

PR #1110, opened 2023-03-01
Use onPart attribute which exists in definition of pykickstart.

mmatsuya:
populate_ksdata() uses data.on_part for the value of --onpart option. But, it doesn't exist in pykickstart side. It should be data.onPart, instead, since it's defined in pykickstart. Also, use UUID for onPart value since it may be better.

This causes a problem of missing --onpart option of the part commands in /root/anaconda-ks.cfg.

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -36,6 +36,7 @@ from ..flags import flags
 from ..storage_log import log_method_call
 from .. import udev
 from ..formats import DeviceFormat, get_format
+from ..formats.lvmpv import LVMPhysicalVolume
 from ..devicelibs.gpt import gpt_part_uuid_for_mountpoint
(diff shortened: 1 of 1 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (commented):
Thank you for this fix. I went through our other `populate_ksdata` functions and found even more problems so I created a new PR fixing all of them: #1111 including the `onPart` one.

We did a PEP8 compliance rewrite few years ago changing all camel case names to snake case and apparently changed the pykickstart attributes too. It's interesting that nobody noticed until now.

vojtechtrefny:
Obsoleted by #1111

Issue #1014, opened 2022-01-24
Make sure libblockdev exceptions are not raised from Blivet


vojtechtrefny:
Libblockdev is a low level library we use for storage operations and the exception we get from it shouldn't be propagated to Blivet users which currently need to import libblockdev and catch these low level exceptions, all Blivet calls to libblockdev should be protected by `try-exec` and we should raise our specific exceptions if these calls fail.

PR #1115, opened 2023-03-16
Avoid raising libblockdev exceptions from our code

vojtechtrefny:
We shouldn't force our users to import libblockdev just to be able to catch storage exceptions. This commit makes sure all libblockdev calls are protected with try-except and we raise our own exceptions like StorageError.

Fixes: #1014


VladimirSlavik (commented):
LGTM, although I can't really verify that these are all the places to fix :)

PR #1119, opened 2023-03-22
Fstab support (rebased to 3.9)

japokorn:
- added support for blivet to directly modify fstab based on blivet actions
- added tests

### DIFF
--- a/blivet/fstab.py
+++ b/blivet/fstab.py
@@ -1,0 +1,364 @@
+import os
+
+import gi
+
+gi.require_version("BlockDev", "2.0")
+from gi.repository import BlockDev as blockdev
+
+from libmount import Table, Fs, MNT_ITER_FORWARD
+from libmount import Error as LibmountException
+
+import logging
+log = logging.getLogger("blivet")
+
+
+def parser_errcb(tb, fname, line):  # pylint: disable=unused-argument
+    print("{:s}:{:d}: parse error".format(fname, line))
+    return 1
+
+
+class FSTabManager():
+    # Read, write and modify fstab file
+    # This class is meant to work even without blivet.
+    # However some of its methods require blivet and will not function without
+    # it. These methods will import what they need when they are run.
+
+    def __init__(self, src_file=None, dest_file=None):
+        self._table = Table()   # Space for parsed fstab contents
+
+        self.src_file = src_file
+        self.dest_file = dest_file
+
+        if self.src_file is not None:
+            self.read(self.src_file)
+
+    def __deepcopy__(self, memo):
+        clone = FSTabManager(src_file=self.src_file, dest_file=self.dest_file)
+        clone._table = Table()
+        clone._table.enable_comments(True)
+
+        entry = self._table.next_fs()
+        entries = [entry]
+        while entry:
+            entries.append(entry)
+            entry = self._table.next_fs()
+
+        # Libmount does not allow to simply use clone._table.add_fs(entry), so...
+        for entry in entries:
+            new_entry = Fs()
+            new_entry.source = entry.source
+            new_entry.target = entry.target
+            new_entry.fstype = entry.fstype
+            new_entry.append_options(entry.options)
+            new_entry.freq = entry.freq
+            new_entry.passno = entry.passno
+            if entry.comment is not None:
+                new_entry.comment = entry.comment
+            clone._table.add_fs(new_entry)
+
+        return clone
+
+    def _get_containing_device(self, path, devicetree):
+        # Return the device that a path resides on
+        if not os.path.exists(path):
+            return None
+
+        st = os.stat(path)
+        major = os.major(st.st_dev)
+        minor = os.minor(st.st_dev)
+        link = "/sys/dev/block/%s:%s" % (major, minor)
+        if not os.path.exists(link):
+            return None
+
+        try:
+            device_name = os.path.basename(os.readlink(link))
+        except Exception:  # pylint: disable=broad-except
+            log.error("failed to find device name for path %s", path)
+            return None
+
+        if device_name.startswith("dm-"):
+            # have I told you lately that I love you, device-mapper?
+            # (code and comment copied from anaconda, kept for entertaining purposes)
+            device_name = blockdev.dm.name_from_node(device_name)
+
+        return devicetree.get_device_by_name(device_name)
+
+    def _from_device(self, device):
+        # Return the list of fstab options obtained from device
+        # *(result) of this method is meant to be used as a parameter in other methods
+
+        spec = getattr(device, 'fstab_spec', None)
+
+        file = None
+        if device.format.mountable:
+            file = device.format.mountpoint
+        if device.format.type == 'swap':
+            file = 'swap'
+
+        vfstype = device.format.type
+        mntops = None
+
+        return spec, file, vfstype, mntops
+
+    def _from_entry(self, entry):
+        return entry.source, entry.target, entry.fstype, ','.join(entry.options)
+
+    def read(self, src_file=''):
+        # Read fstab file
+
+        # Reset table
+        self._table = Table()
+        self._table.enable_comments(True)
+
+        # resolve which file to read
+        if src_file == '':
+            if self.src_file is None:
+                # No parameter given, no internal value
+                return
+        elif src_file is None:
+            return
+        else:
+            self.src_file = src_file
+
+        self._table.errcb = parser_errcb
+        self._table.parse_fstab(self.src_file)
+
+    def find_device_by_specs(self, blivet, spec, file_dummy=None, vfstype_dummy=None, mntops=""):  # pylint: disable=unused-argument
+        # Parse an fstab entry for a device, return the corresponding device,
+        # return None if not found
+        # dummy arguments allow using result of _from_device/_from_entry as a parameter of this method
+        return blivet.devicetree.resolve_device(spec, options=mntops)
+
+    def find_device_by_entry(self, blivet, entry):
+        args = self._from_entry(entry)
+        return self.find_device_by_specs(blivet, *args)
+
+    def get_device_by_specs(self, blivet, spec, file, vfstype, mntops):
+        # Parse an fstab entry for a device, return the corresponding device,
+        # create new one if it does not exist
+
+        from blivet.formats import get_format
+        from blivet.devices import DirectoryDevice, NFSDevice, FileDevice, NoDevice
+        from blivet.formats import get_device_format_class
+
+        # no sense in doing any legwork for a noauto entry
+        if "noauto" in mntops.split(","):
+            raise ValueError("Unrecognized fstab entry value 'noauto'")
+
+        # find device in the tree
+        device = blivet.devicetree.resolve_device(spec, options=mntops)
+
+        if device:
+            # fall through to the bottom of this block
+            pass
+        elif ":" in spec and vfstype.startswith("nfs"):
+            # NFS -- preserve but otherwise ignore
+            device = NFSDevice(spec,
+                               fmt=get_format(vfstype,
+                                              exists=True,
+                                              device=spec))
+        elif spec.startswith("/") and vfstype == "swap":
+            # swap file
+            device = FileDevice(spec,
+                                parents=self._get_containing_device(spec, blivet.devicetree),
+                                fmt=get_format(vfstype,
+                                               device=spec,
+                                               exists=True),
+                                exists=True)
+        elif vfstype == "bind" or "bind" in mntops:
+            # bind mount... set vfstype so later comparison won't
+            # turn up false positives
+            vfstype = "bind"
+
+            # This is probably not going to do anything useful, so we'll
+            # make sure to try again from FSSet.mount_filesystems. The bind
+            # mount targets should be accessible by the time we try to do
+            # the bind mount from there.
+            parents = self._get_containing_device(spec, blivet.devicetree)
+            device = DirectoryDevice(spec, parents=parents, exists=True)
+            device.format = get_format("bind",
+                                       device=device.path,
+                                       exists=True)
+        elif file in ("/proc", "/sys", "/dev/shm", "/dev/pts",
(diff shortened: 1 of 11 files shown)


japokorn:
Summary of replies to review comments from #1093:

- changed API to avoid using confusing arguments
- added docstrings
- `_get_containing_device` method renamed to `get_device` and reworked (unclear parts were dropped or changed)
- added methods for better access; it is now possible to iterate over fstab entries, print, and compare them directly
- extended tests
- other small changes

japokorn:
`blivet/formats/fs.py`: `change_mountpoint` intentionally left blank
The point of this is to be able to revert any mountpoint changes in case the action gets cancelled. Since it is just a property value change, it does not need any code but functionality remains. I am open to suggestions but this came to me as a simple solution that utilizes structures already in place.

vojtechtrefny:
image

These tests run using [GitHub Actions](https://docs.github.com/en/actions) and not in our CI, the missing dependencies need to be resolved in the specific [workflow](https://github.com/storaged-project/blivet/tree/3.8-devel/.github/workflows) configuration.

blivet/fstab.py
```diff
@@ -0,0 +1,790 @@
...
+    def freq(self, value):
+        self._entry.freq = value
+
+    @property
+    def passno(self):
+        return self._entry.passno
+
+    @passno.setter
+    def passno(self, value):
+        self._entry.passno = value
```
dwlehman, line 225: My recollection is that the only valid values for `passno` are 0, 1, 2. Should this be checked/enforced here. Perhaps somewhere else?

blivet/fstab.py
```diff
@@ -664,6 +685,34 @@ def find_entry(self, spec=None, file=None, *, entry=None):
...
+        elif spec_type == "PARTLABEL" and device.format.name:
+            spec = "PARTLABEL=%s" % device.format.name
+        elif spec_type == "PATH":
+            spec = device.path
+        elif device.format.uuid:
+            # default choice
+            spec = "UUID=%s" % device.format.uuid
+        else:
+            # if everything else failed, let blivet decide
+            return None
```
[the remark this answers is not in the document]
japokorn: Returning `None` in this case falls back into blivet logic which then determines the best usable option. Forcing `device.path` here might cause some issues if it is not available for some reason.

blivet/fstab.py
```diff
@@ -0,0 +1,790 @@
...
+        if entry is None:
+            self._entry = Fs()
+        else:
+            self._entry = entry
+
+        if spec is not None:
+            self.spec = spec
+        if file is not None:
+            self.file = file
+        if vfstype is not None:
```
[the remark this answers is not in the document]
japokorn: For the fstab fields I am using the same names as in man pages for fstab (without `fs_` prefix). I hope this way it will be easier to cross reference the blivet change to the fstab.

blivet/fstab.py
japokorn: Fixed.

blivet/fstab.py
```diff
@@ -0,0 +1,790 @@
...
+        self._table = Table()
+        self._table.enable_comments(True)
+
+        # resolve which file to read
+        if src_file == "":
+            if self.src_file is None:
+                # No parameter given, no internal value
+                return
+        elif src_file is None:
+            return
```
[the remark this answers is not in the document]
japokorn: Fixed.

blivet/fstab.py
```diff
@@ -0,0 +1,790 @@
...
+
+        if entry is None:
+            self._entry = Fs()
+        else:
+            self._entry = entry
+
+        if spec is not None:
+            self.spec = spec
+        if file is not None:
+            self.file = file
```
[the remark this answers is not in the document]
japokorn: See my response to the `vfstype` comment. Still, I do agree that `file` is weird but for the sake of consistency I left it as is.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        if device.format.uuid:
+            entry.spec = "UUID=%s" % device.format.uuid
+        else:
+            entry.spec = getattr(device, "fstab_spec", None)
+
+        entry.file = None
+        if device.format.mountable:
+            entry.file = device.format.mountpoint
+        if device.format.type == "swap":
+            entry.file = "swap"
```
vojtechtrefny, line 382: This should definitely raise an exception when the format is not mountable or swap. I can now generate entry for my LUKS device which is definitely not the best idea.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        entry = FSTabEntry()
+
+        if device.format.uuid:
+            entry.spec = "UUID=%s" % device.format.uuid
+        else:
+            entry.spec = getattr(device, "fstab_spec", None)
+
+        entry.file = None
+        if device.format.mountable:
+            entry.file = device.format.mountpoint
```
vojtechtrefny, line 380: What if the device doesn't have a mount point set? I don't think that `None` is a valid value in fstab:

```
In [13]: f.entry_from_device(nvme0n1p1)
Out[13]: UUID=3E7F-3333         efi     None    0       0
```

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+
+            :keyword device: device to process
+            :type device: :class: `blivet.devices.StorageDevice`
+            :returns: fstab entry object based on device
+            :rtype: :class: `FSTabEntry`
+        """
+
+        entry = FSTabEntry()
+
+        if device.format.uuid:
```
vojtechtrefny, line 373: Shouldn't this also use the preferred spec type from `FSTabOptions`?

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        """
+
+        # Reset table
+        self._table = Table()
+        self._table.enable_comments(True)
+
+        # resolve which file to read
+        if src_file == "":
+            if self.src_file is None:
+                # No parameter given, no internal value
```
vojtechtrefny, line 403: So if no `src_file` is given the `read` call is no-op? Shouldn't that be error?

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        else:
+            self.src_file = src_file
+
+        self._table.errcb = self._parser_errcb
+        try:
+            self._table.parse_fstab(self.src_file)
+        except Exception as e:  # pylint: disable=broad-except
+            # libmount throws general Exception. Okay...
+            if str(e) == "No such file or directory":
+                log.warning("Fstab file '%s' does not exist", self.src_file)
```
vojtechtrefny, line 416: You can check for the file existence yourself instead of parsing the error message. Also this should raise an exception.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        if device.format.type == "swap":
+            entry.file = "swap"
+
+        entry.vfstype = device.format.type
+
+        return entry
+
+    def read(self, src_file=""):
+        """ Read the fstab file from path stored in self.src_file. Setting src_file parameter overrides
+            it with new value.
```
vojtechtrefny, line 390: I don't really see a point in having the optional `src_file` argument here if it overrides `self.src_file`. I'd prefer having it as property and telling users to set it instead if they want to change it.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        except Exception as e:  # pylint: disable=broad-except
+            # libmount throws general Exception. Okay...
+            if str(e) == "No such file or directory":
+                log.warning("Fstab file '%s' does not exist", self.src_file)
+            else:
+                raise
+
+    def find_device(self, blivet, spec=None, mntops=None, blkid_tab=None, crypt_tab=None, *, entry=None):
+        """ Find a blivet device, based on spec or entry. Mount options can be used to refine the search.
+            If both entry and spec/mntops are given, spec/mntops are prioritized over entry values.
```
vojtechtrefny, line 422: I'd prefer having two functions here instead of this. For example something like `find_device_by_entry` and `find_device_by_spec`.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        try:
+            self._table.parse_fstab(self.src_file)
+        except Exception as e:  # pylint: disable=broad-except
+            # libmount throws general Exception. Okay...
+            if str(e) == "No such file or directory":
+                log.warning("Fstab file '%s' does not exist", self.src_file)
+            else:
+                raise
+
+    def find_device(self, blivet, spec=None, mntops=None, blkid_tab=None, crypt_tab=None, *, entry=None):
```
vojtechtrefny, line 420: I know that `resolve_device` can use `blkid_tab` and `crypt_tab`, but I don't really see the use case for those here? Do you have an example where it would be useful?

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+            :rtype: :class: `~.devices.StorageDevice` or None
+        """
+
+        _spec = spec or (entry.spec if entry is not None else None)
+        _mntops = mntops or (entry.mntops if entry is not None else None)
+        _mntops_str = ",".join(_mntops) if mntops is not None else None
+
+        return blivet.devicetree.resolve_device(_spec, options=_mntops_str, blkid_tab=blkid_tab, crypt_tab=crypt_tab)
+
+    def get_device(self, blivet, spec=None, file=None, vfstype=None,
```
vojtechtrefny, line 446: Why do we have two very similar functions `find_device` and `get_device`? Is there a use case for creating `FileDevice` and `DirectoryDevice` devices that are not in devicetree and returning them?

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+            :type spec: str
+            :keyword file: device mount path (see man fstab(5) fs_file)
+            :type file: str
+            :keyword entry: fstab entry as an alternative input type
+            :type: :class: `FSTabEntry`
+        """
+
+        fs = self.find_entry(spec, file, entry=entry)
+        if fs:
+            self._table.remove_fs(fs.entry)
```
vojtechtrefny, line 594: This should raise an exception when the entry is not found.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+        if file is not None:
+            try:
+                found_entry = self._table.find_target(file, MNT_ITER_FORWARD)
+            except LibmountException:
+                return None
+            return FSTabEntry(entry=found_entry)
+
+        return None
+
+    def get_spec(self, device):
```
vojtechtrefny, line 690: Do we need to have this as a public function? Looks like it is useful only internally.

vojtechtrefny (commented):
Just a few observations based on trying to use `blivet.fstab.FSTabManager` as a standalone module without blivet.

blivet/fstab.py
japokorn: `entry_from_device` is meant to produce even incomplete entries which can then be used for example to look for existing devices in fstab using `find_entry`. To prevent invalid entries to get into the output file there is a sanity check in the `FSTabManager.write`.

blivet/fstab.py
japokorn: It is not a no-op. If no `src_file` is given it behaves like it is asked to read an empty file and (re)sets its internal table.

blivet/fstab.py
japokorn: Fixed. Now raises ValueError if device is neither mountable nor swap. Modified code and tests as well.

blivet/fstab.py
japokorn: This function is meant to be able to produce incomplete entries as well (these are used for looking up existing entries). There is a sanity check in `FSTabManager.write` that prevents invalid entries to be written into the file in case of misuse.

blivet/fstab.py
japokorn: It is not a no-op. Without set `src_file` the `read` method (re)sets its internal table. Same result as reading an empty fstab file.

blivet/fstab.py
japokorn: Fixed. Now checks the for file existence. Can raise `FileNotFoundError`.

blivet/fstab.py
japokorn: Fixed. `src_file` argument completely removed. Now `self.src_file` variable has to be modified to change the read location.

blivet/fstab.py
japokorn: Combining these functions together allows finding the device based on multiple criteria if needed.

blivet/fstab.py
japokorn: These options are available for the `resolve_device` that this method wraps. I see no reason to limit its functionality. Anaconda works with `blkid_tab` and `crypt_tab` so I expect they might have some usecase.

blivet/fstab.py
japokorn: `get_device` is a method taken from anaconda. I am preserving it for compatibility reasons.

blivet/fstab.py
japokorn: Fixed. Raises ValueError if no matching entry is found in fstab.

blivet/fstab.py
japokorn: Fixed. Renamed to `_get_spec`.

blivet/fstab.py
```diff
@@ -0,0 +1,792 @@
...
+            entries.append(entry)
+            entry = self._table.next_fs()
+
+        for entry in entries:
+            new_entry = self._copy_fs_entry(entry)
+            clone._table.add_fs(new_entry)
+
+        return clone
+
+    def __repr__(self):
```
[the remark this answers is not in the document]
japokorn: Fixed. Renamed method to `__str__`.

blivet/fstab.py
japokorn: Fixed. Using `_get_spec` instead.

blivet/fstab.py
vojtechtrefny: Ok, please make a note about this in the docstring, thanks.

vojtechtrefny (approved):

PR #1120, opened 2023-04-04
Add support for filesystem online resize

vojtechtrefny:


blivet/formats/fs.py
```diff
@@ -436,12 +439,27 @@ def _post_create(self, **kwargs):
...
-        # file systems need a check before being resized
-        self.do_check()
+        if self.status:
+            if flags.allow_online_fs_resize:
+                if self.target_size > self.size and not self._resize_support & FSResize.ONLINE_GROW:
+                    raise FSError("This filesystem doesn't support online growing")
+                if self.target_size < self.size and not self._resize_support & FSResize.ONLINE_SHRINK:
+                    raise FSError("This filesystem doesn't support online shrinking")
+            else:
+                raise FSError("Resizing of mounted filesystems is disabled")
```
vojtechtrefny, line 449: In general should this be considered an error or do we want to try to unmount the filesystem right now? Currently for partitions and LVs the block device `_pre_resize` functions unmount (`teardown`) the format if the `allow_online_fs_resize` flag isn't set, but we might try to unmount the format and do the online resize only when the unmount fails.

blivet/formats/fs.py
```diff
@@ -838,6 +856,7 @@ class Ext2FS(FS):
     _formattable = True
     _supported = True
     _resizable = True
+    _resize_support = FSResize.ONLINE_GROW | FSResize.OFFLINE_GROW | FSResize.OFFLINE_SHRINK
```
vojtechtrefny, line 859: We'll eventually get these values from libblockdev once we release 3.0.

PR #1122, opened 2023-04-26
WIP : update dbus example to walk the tree and attempt to create a real volume via the dbus interface.

trgill:
The commit() is failing with:

`LookupError: <blivet.dbus.format.DBusFormat at 0x7f3209b0a190> is not exported`

Also the Factory() method takes size as a dbus Int32, we should probably update it.  (I'm assuming it is in bytes)

I added --blockdevs parameter to the example client.  Use with caution.  The devices specified will be used to construct a new LVM volume.


StorageGhoul:

PR #1123, opened 2023-04-26
Add support for creating shared LVM setups

vojtechtrefny:
This feature is requested by GFS2 for the storage role. This adds support for creating shared VGs and activating LVs in shared mode.

### DIFF
--- a/blivet/__init__.py
+++ b/blivet/__init__.py
@@ -63,15 +63,20 @@ gi.require_version("BlockDev", "3.0")
 from gi.repository import GLib
 from gi.repository import BlockDev as blockdev
 if arch.is_s390():
-    _REQUESTED_PLUGIN_NAMES = set(("lvm", "btrfs", "swap", "crypto", "loop", "mdraid", "mpath", "dm", "s390", "nvdimm"))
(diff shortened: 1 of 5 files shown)


japokorn (dismissed):
Looks good to me.

japokorn:
When creating the first shared VG (and using `sanlock` lock manager) `vgcreate` has to be given non-PV device otherwise VG creation will fail. This functionality does not seem to be handled. (When testing the thing I got the error message described in ref.)

Ref: [lvmlockd(8)](https://man7.org/linux/man-pages/man8/lvmlockd.8.html) 'Creating the first sanlock VG'.

vojtechtrefny:
> When creating the first shared VG (and using `sanlock` lock manager) `vgcreate` has to be given non-PV device otherwise VG creation will fail. This functionality does not seem to be handled. (When testing the thing I got the error message described in ref.)
> 
> Ref: [lvmlockd(8)](https://man7.org/linux/man-pages/man8/lvmlockd.8.html) 'Creating the first sanlock VG'.

I added two more commits that should address this -- we don't really need to run `pvcreate` first, so we can just skip it for all shared VGs. I really hope this doesn't mean that **only** the first created VG must be created this way. There might be some more caveats so I am moving this to draft, let me know if the fix works.

japokorn (approved):
LGTM
I managed to test shared VG creation using storage role and ha_cluster system role.
I think this PR can be merged.

PR #1125, opened 2023-04-28
DBus: remove extra callback invocations

dwlehman:
The device_added/device_removed/format_added/format_removed callbacks are for changes to blivet's model -- not when changes are written to disk.

### DIFF
--- a/blivet/dbus/blivet.py
+++ b/blivet/dbus/blivet.py
@@ -141,12 +141,6 @@ class DBusBlivet(DBusObject):
         self._manager.add_object(added)
 
     def _action_executed(self, action):
-        if action.is_destroy:
-            if action.is_device:
-                self._device_removed(action.device, keep=False)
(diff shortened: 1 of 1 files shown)


StorageGhoul:

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

dwlehman:
@vojtechtrefny is this a good choice for a target branch? I don't think there is much of a pressing need to have this in fedora, but I would rather not wait for more than a fedora release cycle or so.

vojtechtrefny:
> @vojtechtrefny is this a good choice for a target branch? I don't think there is much of a pressing need to have this in fedora, but I would rather not wait for more than a fedora release cycle or so.

I plan to release 3.8 for Fedora 39, but we can do an extra release from 3.7 for Fedora 38 if we need this for testing.

PR #1127, opened 2023-05-03
tests: add dbus example to traverse the devices and call test the Factory()

trgill:
Added block device list as "--blockdevs" parameter.  It accepts block devices in the format "/dev/xxx,/dev/xxx".

Call the Factory() to test DEVICE_TYPE_LVM, DEVICE_TYPE_BTRFS, DEVICE_TYPE_MD and DEVICE_TYPE_STRATIS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ tests/pylint/.pylint.d/
 MANIFEST
 misc/.vagrant
 ChangeLog
+.vscode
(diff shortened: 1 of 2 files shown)


trgill:
@vojtechtrefny @dwlehman would it make sense to merge this as an example?

I think we might want to add something similar to automated tests?

These calls are what the CSI driver uses - just FYI.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

vojtechtrefny:
> @vojtechtrefny @dwlehman would it make sense to merge this as an example?

Definitely, more examples could never hurt.

PR #1128, opened 2023-05-04
Do not set memory limit for LUKS2 when running in FIPS mode

vojtechtrefny:
With FIPS enabled LUKS uses pbkdf and not argon so the memory limit is not a valid parameter.

### DIFF
--- a/blivet/formats/luks.py
+++ b/blivet/formats/luks.py
@@ -303,11 +303,13 @@ class LUKS(DeviceFormat):
             if luks_data.pbkdf_args:
                 self.pbkdf_args = luks_data.pbkdf_args
             else:
-                mem_limit = crypto.calculate_luks2_max_memory()
-                if mem_limit:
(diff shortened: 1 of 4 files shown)


blivet/formats/luks.py
```diff
@@ -303,11 +303,12 @@ def _create(self, **kwargs):
             if luks_data.pbkdf_args:
                 self.pbkdf_args = luks_data.pbkdf_args
             else:
-                mem_limit = crypto.calculate_luks2_max_memory()
-                if mem_limit:
-                    self.pbkdf_args = LUKS2PBKDFArgs(max_memory_kb=int(mem_limit.convert_to(KiB)))
-                    luks_data.pbkdf_args = self.pbkdf_args
-                    log.info("PBKDF arguments for LUKS2 not specified, using defaults with memory limit %s", mem_limit)
+                if not crypto.is_fips_enabled():
```
jstodola, line 306: Would it be possible to add a comment here to make it obvious why the code is not needed in FIPS mode?

vojtechtrefny: Ok, I added a comment.

PR #1132, opened 2023-05-26
Always prefer GPT disk labels on x86_64 (and clean up the logic)

AdamWill:
See: https://bugzilla.redhat.com/show_bug.cgi?id=2092091#c6 There was a Fedora 37 Change to prefer GPT disk labels on x86_64 BIOS installs, but for some reason, this was implemented in anaconda by having it ignore blivet's ordering of the disk label types and just universally prefer GPT if it's in the list at all, which resulted in a preference for GPT in more cases than just x86_64 BIOS. This is one step towards fixing that, by putting the 'always prefer GPT on x86_64' logic here in the blivet ordering code where it belongs. Step 2 will be to drop the anaconda code that overrides blivet's preference order.

This also simplifies the logic a bit; it had gotten rather a lot of conditions piled on top of each other and was rather hard to read. This should achieve the same effect as before in a clearer and more concise way.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -223,11 +223,8 @@ class DiskLabel(DeviceFormat):
         label_types = ["msdos", "gpt"]
         if arch.is_pmac():
             label_types = ["mac"]
-        elif arch.is_aarch64():
-            label_types = ["gpt", "msdos"]
-        elif arch.is_efi() and arch.is_arm():
-            label_types = ["msdos", "gpt"]
-        elif arch.is_efi() and not arch.is_aarch64():
+        # always prefer gpt on aarch64, x86_64, and EFI plats except 32-bit ARM
+        elif arch.is_aarch64() or arch.is_x86(bits=64) or (arch.is_efi() and not arch.is_arm()):
             label_types = ["gpt", "msdos"]
         elif arch.is_s390():
             label_types += ["dasd"]
--- a/tests/storage_tests/devices_test/partition_test.py
+++ b/tests/storage_tests/devices_test/partition_test.py
@@ -99,7 +99,7 @@ class PartitionDeviceTestCase(unittest.TestCase):
     def test_min_max_size_alignment(self):
         with sparsetmpfile("minsizetest", Size("10 MiB")) as disk_file:
             disk = DiskFile(disk_file)
-            disk.format = get_format("disklabel", device=disk.path)
+            disk.format = get_format("disklabel", device=disk.path, label_type="msdos")
             grain_size = Size(disk.format.alignment.grainSize)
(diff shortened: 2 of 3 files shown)


StorageGhoul:

AdamWill:
This failure isn't very actionable :( It just says that something called `run_ruff.sh` failed, with no output that I can find. Nothing named `*ruff*` or containing the string `ruff` exists in the blivet git repo.

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
> This failure isn't very actionable :( It just says that something called `run_ruff.sh` failed, with no output that I can find. Nothing named `*ruff*` or containing the string `ruff` exists in the blivet git repo.

We run Anaconda test suite on our PRs to make sure we don't break Anaconda with our changes. Looks like they added some tests that use *ruff* few days ago https://github.com/rhinstaller/anaconda/pull/4783 I'm not sure why it fails, this is all I get from the logs:

```
./ruff/run_ruff.sh: line 8: ruff: command not found
./ruff/run_ruff.sh: line 10: ruff: command not found
FAIL ruff/run_ruff.sh (exit status: 127)
```

The test runs in a container created by Anaconda CI so I'd assume *ruff* would be included there. @VladimirSlavik can you please look at this? Do we need to add *ruff* as dependency [here](https://github.com/storaged-project/blivet/blob/3.8-devel/.github/workflows/anaconda_tests.yml)?

vojtechtrefny (approved):
Looks good to me, thank you.

vojtechtrefny:
I am not sure why the `test_min_max_size_alignment` fails only on Debian, but I've seen some alignment issues with libparted on Debian before in libblockdev and we create very small images for the test (10 MiB) so it's probably just a different between metadata size between GPT and MSDOS, I think adding `label_type="msdos"` [here](https://github.com/storaged-project/blivet/blob/71f2037e07f10ed89885216d0cac5f8969a459cb/tests/storage_tests/devices_test/partition_test.py#LL102C27-L102C37) should fix it.

VladimirSlavik:
> The test runs in a container created by Anaconda CI so I'd assume _ruff_ would be included there. @VladimirSlavik can you please look at this? Do we need to add _ruff_ as dependency [here](https://github.com/storaged-project/blivet/blob/3.8-devel/.github/workflows/anaconda_tests.yml)?

The anaconda-ci container should have it. Please try re-running the test. I think it's due to poor timing, the order of events likely was: The test was added, you tried running it, the container was rebuilt with the dependency.

AdamWill:
fixed tests with @vojtechtrefny 's suggestion, thanks. as discussed in the anaconda PR, we need to merge this and then cut a blivet release and package it into rawhide *before* we make the anaconda change...

PR #1135, opened 2023-06-08
Prepare for libblockdev 3.0 release

vojtechtrefny:
We need to make some changes related to the upcoming libblockdev 3.0 release. There are some API changes (see https://github.com/storaged-project/libblockdev/pull/877) and we also need to start requiring 3.0 from GI repository.

Note that all tests will fail until we actually bump libblockdev version to 3.0.


vojtechtrefny:
Two new commits added to deal with the crypto plugin API changes, see https://github.com/storaged-project/libblockdev/pull/898 for details.

PR #1142, opened 2023-07-12
Make sure that LUKS.has_key always returns a boolean value

vojtechtrefny:
We don't want to return 'None' here.

Related: https://github.com/rhinstaller/anaconda/pull/4873

### DIFF
--- a/blivet/devices/stratis.py
+++ b/blivet/devices/stratis.py
@@ -129,8 +129,8 @@ class StratisPoolDevice(StorageDevice):
 
     @property
     def has_key(self):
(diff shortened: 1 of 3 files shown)


poncovka (commented):
Thanks! Btw. I think there are three different definitions of `has_key` that look like this.

jkonecny12 (approved):
Thanks for the fix.

vojtechtrefny:
> Thanks! Btw. I think there are three different definitions of `has_key` that look like this.

Oh, I forgot about Stratis. Fixed, thanks.

japokorn (approved):
Looks good to me.

PR #1154, opened 2023-09-11
nvme: Require additional rpms for dracut

tbzatek:
The '95nvmf' dracut module needs a couple more packages for the NBFT (NVMe over TCP) to work - such as networking. Local PCIe NVMe devices have no special needs.

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -725,7 +725,8 @@ class NVMeFabricsNamespaceDevice(NVMeNamespaceDevice, NetworkStorageDevice):
 
     """ NVMe fabrics namespace """
     _type = "nvme-fabrics"
-    _packages = ["nvme-cli"]
+    # dracut '95nvmf' module dependencies
+    _packages = ["nvme-cli", "dracut-network"]
 
     def __init__(self, device, **kwargs):
         """


tbzatek:
This is mostly untested at the moment, additional dependencies may be required. This is also different between Fedora and RHEL.

blivet/devices/disk.py
```diff
@@ -725,7 +725,8 @@ class NVMeFabricsNamespaceDevice(NVMeNamespaceDevice, NetworkStorageDevice):
 
     """ NVMe fabrics namespace """
     _type = "nvme-fabrics"
-    _packages = ["nvme-cli"]
+    # dracut '95nvmf' module dependencies
+    _packages = ["nvme-cli", "dracut-network", "jq"]
```
vojtechtrefny, line 729: `jq` sounds like a really weird dependency for us to install

tbzatek: Agree, this looks out of scope of blivet, though I haven't found a better place. This would perhaps need to be somewhere in Anaconda otherwise.

vojtechtrefny (approved):
If no one wants to add the `jq` dependency, I guess we can do that. But I still don't like it :-)

tbzatek:
So the `jq` dependency won't be needed anymore, the following dracut changes should take care of satisfying additional dependencies:
* c9s: https://gitlab.com/redhat/centos-stream/rpms/dracut/-/merge_requests/28
* Fedora: https://src.fedoraproject.org/rpms/dracut/pull-request/45

vojtechtrefny (approved):

PR #1159, opened 2023-10-03
Revert unneeded changes for shared LVM support

vojtechtrefny:
This part of #1123 was needed only for shared LVM with sanlock which we are not going to support.


japokorn (approved):
Looks good to me.

PR #1165, opened 2023-10-31
Fix DMDevice hard dependency on kpartx

vojtechtrefny:
`DMDevice` now depends on `kpartx` which means all DM devices (LUKS, LVM etc.) depend on `kpartx` which is really needed only for multipath. This PR moves the dependency to multipath and also allows avoids unnecessary exceptions when `kpartx` is not available.


japokorn (approved):
Looks good to me.

PR #1166, opened 2023-11-02
 Added missing fstab object to SwapSpace

japokorn:
Anaconda tests revealed missing `FSTabOptions` object in `SwapSpace` class.
This issue is related to the recent merge of fstab rework #1119 which
added this object to `FS` class. `SwapSpace` does not inherit from `FS` and
was overlooked. This change adds the missing object.
    
It also adds logic that skips all fstab related operations if fstab file
is not to be written (i.e. `fstab.dest_file is None`). The reason for this is to
eliminate the risk of potential issues caused by unused component.
    
The test for above has been added as well.

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -281,6 +281,8 @@ class ActionList(object):
         devices = devices or []
         self._pre_process(devices=devices)
 
+        skip_fstab = fstab is None or fstab.dest_file is None
+
         for action in self._actions[:]:
             log.info("executing action: %s", action)
             if dry_run:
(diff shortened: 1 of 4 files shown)


vojtechtrefny (changes requested):
Please add a test case that will cover this.

japokorn:
Added the test. Also modified the code so now unnecessary logic is skipped if no fstab is to be written (i.e. `fstab.dest_file is None`).

vojtechtrefny (approved):

PR #1163, opened 2023-10-27
Use libblockdev FS plugin for filesystem operations

vojtechtrefny:
With libblockdev 3.0 now released and required in Blivet, we can now start using the FS plugin for various filesystem operations we previously used the command line tools directly.

Most of the changes are just replacing the function calls with libblockdev. One major change is removing some filesystem support. We'll remove support for creating and managing of ReiserFS, JFS and HFS. These filesystems are either unsupported (ReiserFS userspace tools were removed from Fedora and ReiserFS is also deprecated in the upstream kernel) or very old and not used (HFS is supported only on PPC Mac which is no longer supported in Fedora, we are not removing HFS+ support).

TODO:
- [x] FS min size support -- blivet uses different tools for getting filesystem min size (`ntfsresize` used by blivet vs `ntfsinfo` used by libblockdev to get filesystem free space), we need to add support to libblockdev first
- [x] swap UUID and label format check, libblockdev support pending release: https://github.com/storaged-project/libblockdev/pull/978
- [ ]  fsck support -- blivet uses error codes for custom error messages, which libblockdev doesn't support

PR #1167, opened 2023-11-03
fcoe/iscsi: Use libblockdev to laod modules instead of modprobe

vojtechtrefny:
Related: #1163


tbzatek (approved):
No objections apart that typo in the commit message.

PR #1170, opened 2023-11-14
Blivet-GUI reverse dependency testing in GH actions

vojtechtrefny:
A GH action to run blivet-gui test suite so we can avoid issues like https://github.com/storaged-project/blivet-gui/pull/418 in the future.


VladimirSlavik (approved):
Thank you!

PR #1174, opened 2023-11-23
Support partitioning of hybrid boot disks

poncovka:
Anaconda needs to be able to create hybrid boot disks. For example:

```
  clearpart --all --initlabel --disklabel=gpt
  part prepboot  --size=4    --fstype=prepboot
  part biosboot  --size=1    --fstype=biosboot
  part /boot/efi --size=100  --fstype=efi
  part /boot     --size=1000 --fstype=ext4 --label=boot
  part /         --grow      --fstype xfs
```
However, this kickstart snippet is not working with two or more disks. The bootloader-related partitions should be all created on the disk the computer will boot from, but Blivet does that only for platform -specific partitions. The rest of them are created on any disk with enough space.

It looks like this can be easily fixed by setting the same weight to all of these partitions regardless of the current platform.

See: https://github.com/rhinstaller/anaconda/pull/5298

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -452,28 +452,24 @@ class PartitionDevice(StorageDevice):
         if isinstance(self.req_base_weight, int):
             return self.req_base_weight
 
-        # now we have the weights for varying mountpoints and fstypes by platform
-        weight = 0
-        if self.format.mountable and self.format.mountpoint == "/boot":
-            weight = 2000
-        elif (self.format.mountable and
-              self.format.mountpoint == "/boot/efi" and
-              self.format.type in ("efi", "macefi") and
-              arch.is_efi()):
-            weight = 5000
-        elif arch.is_x86() and self.format.type == "biosboot" and not arch.is_efi():
-            weight = 5000
-        elif self.format.mountable and arch.is_arm():
-            # On ARM images '/' must be the last partition.
-            if self.format.mountpoint == "/":
-                weight = -100
-        elif arch.is_ppc():
-            if arch.is_pmac() and self.format.type == "appleboot":
-                weight = 5000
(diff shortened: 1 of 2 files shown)


StorageGhoul:

poncovka:
@vojtechtrefny Does it make sense? Can we do this change globally or do you want to control it via a flag?

vojtechtrefny:
Jenkins, ok to test.

blivet/devices/partition.py
```diff
@@ -454,23 +454,22 @@ def _get_weight(self):
...
         if self.format.mountable and self.format.mountpoint == "/boot":
             weight = 2000
         elif (self.format.mountable and
               self.format.mountpoint == "/boot/efi" and
-              self.format.type in ("efi", "macefi") and
-              arch.is_efi()):
+              self.format.type in ("efi", "macefi")):
             weight = 5000
-        elif arch.is_x86() and self.format.type == "biosboot" and not arch.is_efi():
+        elif self.format.type == "biosboot":
```
vojtechtrefny, line 464: I think biosboot, appleboot and prepboot can be grouped here, just to simplify it a little bit more.

vojtechtrefny (approved):
Looks good to me (except the bad indentation pylint is complaining about). I think we change this globally, I don't see a situation where it could be a problem and I don't like adding another flag.

poncovka:
Updated.

vojtechtrefny (approved):

poncovka:
Just an update, Anaconda is not going to add the support for hybrid boot (at least for now, see https://github.com/rhinstaller/anaconda/pull/5298#issuecomment-1827687218). I talked to @vojtechtrefny about this pull request and we think it's still a useful fix for anyone who wants to experiment with this feature in the `%post` section. So the conclusion is that we leave it merged.

PR #1183, opened 2023-12-06
tests: Add a simple unit test for the NVMe module

vojtechtrefny:
Just a simple heavily mocked test case, but it should prevent issues like the one currently blocking https://github.com/rhinstaller/anaconda/pull/5357


tbzatek (approved):
Probably looks good

PR #1196, opened 2024-02-02
Fix crash when scanning degraded/not fully assembled MD arrays

vojtechtrefny:
When we don't have the array name we cannot construct device ID for the devicetree lookup, we need to let the code fallback to other means like UUID.


KKoukiou (approved):

vojtechtrefny:
Jenkins, test this please.

PR #1197, opened 2024-02-08
Fstab cleanup fix

japokorn:
Formats with multiple layers (e.g. LUKS) were not handled properly in fstab. This resulted in fstab entries not being properly cleaned from fstab when destroy actions were executed. This fixes the issue by using different variables with correct values based on action type.

### DIFF
--- a/blivet/actionlist.py
+++ b/blivet/actionlist.py
@@ -292,7 +292,7 @@ class ActionList(object):
             # (device may not exist afterwards)
             if not skip_fstab:
                 try:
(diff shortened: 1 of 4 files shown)


vojtechtrefny (changes requested):
Looks good to me, but please add a test case covering this issue.

japokorn:
> Looks good to me, but please add a test case covering this issue.

Added the test.

vojtechtrefny (approved):

PR #1204, opened 2024-03-04
Fixes for latest Ubuntu/Debian support

vojtechtrefny:
The biggest change is libmount requirement which is not available on Debian/Ubuntu, so we should be able to work without it if possible. The other changes are mostly small fixes in the dependencies installation.


japokorn (approved):
Looks good to me.

PR #1209, opened 2024-03-12
Fix scanning partitions on RAID arrays (#2269133)

vojtechtrefny:


AdamWill:
would this have problems if we had two devices with the same 'name'? haven't we run into issues with that recently?

AdamWill:
Still, aside from that, can confirm this fixes the bug at least in a simple case. Was able to reproduce the bug (both elements), then booted with the updates image, and both the custom interface and 'reclaim space' worked as expected.

vojtechtrefny:
> would this have problems if we had two devices with the same 'name'? haven't we run into issues with that recently?

Yes, unfortunately. The [unique device identifier change](https://github.com/storaged-project/blivet/pull/1182) works only if we know the device type we are dealing with and because partitions can be on any device (in theory, nobody is putting partitions on LVs, at least I hope) we cannot use it here, so we need to use name and hope for the best. We had this issue mostly with btrfs and LVM so it shouldn't be a big issue here.

PR #1211, opened 2024-03-14
nvme: Skip startup/write when NVMe plugin isn't available

vojtechtrefny:
This is similar to other modules like iSCSI where these methods are silently skipped if the technology isn't supported or available.


tbzatek (approved):
Nothing wrong jumps at me while reading this so I guess it's fine.

PR #1218, opened 2024-03-28
drop python<=3.7 support

kloczek:
Filter all python code over `pyuprade --py38-plus` and mark in setup.py that this module is only for python>=3.8.


StorageGhoul:

vojtechtrefny (commented):
I guess this looks OK to me in general. I don't see a reason for dropping the <= 3.7 support, but I also don't see a reason for not doing that, so if it's going to help you (or someone else) we can merge this. But please fix the pep8/pycodestyle failures. Also the `translation-canary` folder is a git subtree so the changes to it should be separate and go through https://github.com/rhinstaller/translation-canary

kloczek:
>  I don't see a reason for dropping the <= 3.7 support

`blivet` module is used by `blivet-gui`.
At the moment it is not possible to build full GUI desktop without at least python 3.10 (mesa, some parts of gnome during build and rust rpm base tooling needs python>=3.10).

IIRC fc38 uses python 3.10 or 3.11 so with current CI definition it is not possible to test this module with python<=3.9.

In other words it is nothing more than just cutting some legacy tail 😋

vojtechtrefny:
> In other words it is nothing more than just cutting some legacy tail

Asi I said, I am not against doing this, but currently the tests are failing (static analysis is complaining about indentation) so if you want this merged, please fix the failures first.

kloczek:
Is it any problem with merge this PR an add indentation fixes?

PR #1229, opened 2024-04-25
Added support for PV grow

japokorn:
Storage role requires support for a case when PV has to be resized to fill all available space when its device's size changes (usually on VM).

A new flag `grow_to_fill` was added, which marks the device for size expansion (all available space it taken).
Proper size is determined by LVM, avoiding inaccurate size calculations in blivet.

### DIFF
--- a/blivet/formats/lvmpv.py
+++ b/blivet/formats/lvmpv.py
@@ -33,7 +33,7 @@ from ..devicelibs import lvm
 from ..tasks import availability, pvtask
 from ..i18n import N_
 from ..size import Size
-from ..errors import PhysicalVolumeError
+from ..errors import DeviceFormatError, PhysicalVolumeError
 from . import DeviceFormat, register_device_format
 from .. import udev
 from ..static_data.lvm_info import pvs_info, vgs_info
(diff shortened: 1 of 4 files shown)


vojtechtrefny (commented):
Looks good to me in general, but it needs a test.

japokorn:
- Added missing tests.
- Replaced `blockdev.lvm.pvresize(self.pv.device)` with `blockdev.lvm.pvresize(self.pv.device, 0)`
- Made `grow_to_fill` a property that cannot be set unless the device is resizable (raises `DeviceFormatError`)
- Added condition to `formats/__init__.py` that will prevent bypassing of resize when new size is the same as old one if device is type `lvmpv` with `grow_to_fill` set to `True`
- pylint and pep8 fixes

blivet/formats/lvmpv.py
```diff
@@ -107,6 +110,17 @@ def __repr__(self):
                "pe_start": self.pe_start, "data_alignment": self.data_alignment})
         return s
 
+    @property
+    def grow_to_fill(self):
```
vojtechtrefny, line 114: Please add a docstring here explaining how this works, e.g. what target size needs to be set to make this work and that `ActionResizeFormat` still needs to be created and executed.

japokorn: Fixed. Added docstring.

vojtechtrefny (approved):

PR #1247, opened 2024-06-17
Set log level to INFO for libblockdev

vojtechtrefny:
Default log level is WARNING and we want to log even INFO messages for libblockdev. INFO level contains for example also all commands and their output that we want to log to program.log.


dwlehman (approved):
Looks good to me.

PR #1256, opened 2024-07-10
devicetree: resolve devices also with the PARTUUID=.. naming

KKoukiou:
This identifier should be handled as well. It's met in debian fstab.


vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):
Thank you!

KKoukiou:
> I still need to verify that this works for anaconda.

Verified this fixes the issue spotted from anaconda.

PR #1266, opened 2024-08-05
LUKS HW-OPAL support

vojtechtrefny:
Work in progress support for creating and managing OPAL self encrypting drives with LUKS. This for now contains only support for recognizing pre-existing LUKS HW-OPAL devices, because this doesn't require any additional support from libblockdev (which creating and removing these devices will require).


vojtechtrefny:
Note: Tests are failing because https://github.com/storaged-project/libblockdev/pull/1047 needs to be merged first for the "full" LUKS OPAL support.

PR #1267, opened 2024-08-09
devices: catch exceptions where invalid access happens first

KKoukiou:
[storage.log](https://github.com/user-attachments/files/16558156/storage.log)

### DIFF
--- a/blivet/devices/partition.py
+++ b/blivet/devices/partition.py
@@ -377,10 +377,10 @@ class PartitionDevice(StorageDevice):
         if not self.exists:
             return self.part_type_uuid_req
         else:
-            if hasattr(parted.Partition, "type_uuid") and self.parted_partition.type_uuid:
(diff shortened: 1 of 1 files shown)


StorageGhoul:

blivet/devices/partition.py
```diff
@@ -377,9 +377,9 @@ def part_type_uuid(self):
         if not self.exists:
             return self.part_type_uuid_req
         else:
-            if hasattr(parted.Partition, "type_uuid") and self.parted_partition.type_uuid:
```
vojtechtrefny, line 380: I think checking whether the `parted_partition` is set with something like `hasattr(parted.Partition, "type_uuid") and and self.parted_partition and self.parted_partition.type_uuid` would be better than moving the try-except block.

KKoukiou: no problem, adjusted.

vojtechtrefny (approved):

PR #1301, opened 2024-10-03
free_space_estimate: adjust for compression on btrfs

AdamWill:
We have a problem with Fedora aarch64 KDE disk images ATM. We want to make them 16 power-of-ten gigabytes big, so they will fit on "16GB" SD cards / USB sticks. When we set them to this size, compose fails because the anaconda free space check fails, saying "An additional 436 MiB is needed". That check is ultimately backed by this estimate.

In these images, the main system partition is a btrfs volume, with compression enabled. If you examine the last successful image (which was built at size 18 power-of-two GB), thanks to compression, only ~5GB of space is actually occupied on that volume. The contents would easily fit if it were slightly smaller.

With this change we adjust the free space estimate by a compression factor, as well as a metadata factor, if the class defines one. For the btrfs class we set it to 1.4 when the compression flag is set. That's a pretty conservative estimate, we will usually achieve a much better compression ratio with a typical OS payload (as the numbers above show). AFAICT this codepath is only actually used by the anaconda size check, so the change should be fairly safe.

### DIFF
--- a/blivet/formats/fs.py
+++ b/blivet/formats/fs.py
@@ -193,6 +193,14 @@ class FS(DeviceFormat):
         return d
 
     @classmethod
+    def _compression_size_factor(cls):
+        """ Adjust free space estimate for transparent compression.
+        Classes for filesystems that implement transparent compression
+        can override this.
+        """
+        return 1
+
+    @classmethod
     def free_space_estimate(cls, device_size):
         """ Get estimated free space when format will be done on device
             with size ``device_size``.
@@ -207,7 +215,7 @@ class FS(DeviceFormat):
             :return: estimated free size after format
             :rtype: :class:`~.size.Size`
         """
-        return device_size * cls._metadata_size_factor
+        return device_size * cls._metadata_size_factor * cls._compression_size_factor()
 
     @classmethod
     def get_required_size(cls, free_space):
@@ -1107,6 +1115,16 @@ class BTRFS(FS):
             if content and not all(c in self._system_dirs + subvols for c in content):
                 return False
             return True
+
+    @classmethod
+    def _compression_size_factor(cls):
+        """ Adjust free space estimate for transparent compression.
+        We estimate 40% 'additional space' if it's enabled (this is
+        pretty conservative).
+        """
+        if flags.btrfs_compression:
+            return 1.40
+        return 1
 
 
 register_device_format(BTRFS)


AdamWill:
Note I think in theory we should do the same for filesystems on LVM with VDO compression enabled, but that looks a bit harder and we don't need it right now, so I skipped it. :P

Conan-Kudo (approved):

blivet/formats/fs.py
```diff
@@ -207,7 +207,13 @@ def free_space_estimate(cls, device_size):
             :return: estimated free size after format
             :rtype: :class:`~.size.Size`
         """
-        return device_size * cls._metadata_size_factor
+        estimate = device_size * cls._metadata_size_factor
+        try:
+            estimate = estimate * cls._compression_size_factor
+        except AttributeError:
+            # class doesn't have compression size factor, this is fine
+            pass
```
davide125, line 215: If the attribute is missing the statement on 210 will fail as well. Also, is it fine to just swallow exceptions here?

davide125 (changes requested):

blivet/formats/fs.py
AdamWill: line 210 already exists, every class already defines a `_metadata_size_factor`. It just felt weird to define a `_compression_size_factor` of 1 for every class except btrfs. But actually, now I look at it, I think they all inherit from `FS`, so I can just set a factor of 1 there and we shouldn't need this.

Yes, swallowing exceptions is fine, since it just means we return the estimate without adjusting for compression. But anyway it's academic if I change it as described above.

davide125: Got it, thanks for clarifying

AdamWill:
ehhh, I worry there might be an ordering problem between the class definition and the flag setting here, actually. mmmf.

davide125 (approved):

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny (approved):

AdamWill:
huh. is the centos 9 failure real or some kinda test system blip? it does seem like it failed in a related test, but the log shows no info on the failure and seems like it just sorta blew up in a cloud of java?

vojtechtrefny:
> seems like it just sorta blew up in a cloud of java?

That's just the VM dying during the test execution, it sometimes happens, our testing infrastructure has some random issues unfortunately. The Java exception is Jenkins losing connection in the middle of execution.

PR #1313, opened 2024-11-04
Translate vendor id 0x1af4 to Virtio Block Device (#1242117)

vojtechtrefny:
The vendor information used to come from parted which reports virtio blk devices as 'Virtio Block Device'. The sysfs vendor entry has always been 0x1af4 which isn't very descriptive. So in disk.description check for the exception and return "Virtio Block Device". disk.vendor is left as-is, only the user friendly description is modified.

Resolves: rhbz#1242117

### DIFF
--- a/blivet/devices/disk.py
+++ b/blivet/devices/disk.py
@@ -137,7 +137,11 @@ class DiskDevice(StorageDevice):
 
     @property
     def description(self):
-        return " ".join(s for s in (self.vendor, self.model, self.wwn) if s)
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Ok @jkonecny12, you win :-)

jkonecny12:
I hope we are all winners here :). Thanks a lot for looking into this!

jkonecny12 (approved):
Thanks a lot!

PR #1317, opened 2024-11-14
Make GPT default label type on all architectures

vojtechtrefny:
Exceptions are DASD drives on s390 and 32bit ARM. Everywhere else GPT will be default.

### DIFF
--- a/blivet/formats/disklabel.py
+++ b/blivet/formats/disklabel.py
@@ -220,12 +220,13 @@ class DiskLabel(DeviceFormat):
 
     @classmethod
     def get_platform_label_types(cls):
-        label_types = ["msdos", "gpt"]
+        # always prefer gpt except for configurations below
+        label_types = ["gpt", "msdos"]
         if arch.is_pmac():
             label_types = ["mac"]
-        # always prefer gpt on aarch64, x86_64, and EFI plats except 32-bit ARM
-        elif arch.is_aarch64() or arch.is_x86(bits=64) or (arch.is_efi() and not arch.is_arm()):
-            label_types = ["gpt", "msdos"]
+        # prefet msdos on 32-bit ARM
+        elif arch.is_arm():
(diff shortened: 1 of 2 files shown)


blivet/formats/disklabel.py
```diff
@@ -220,12 +220,13 @@ def parted_device(self):
...
     def get_platform_label_types(cls):
-        label_types = ["msdos", "gpt"]
+        # always prefer gpt except for configurations below
+        label_types = ["gpt", "msdos"]
         if arch.is_pmac():
             label_types = ["mac"]
-        # always prefer gpt on aarch64, x86_64, and EFI plats except 32-bit ARM
-        elif arch.is_aarch64() or arch.is_x86(bits=64) or (arch.is_efi() and not arch.is_arm()):
-            label_types = ["gpt", "msdos"]
+        # prefet msdos on 32-bit ARM
```
cgwalters, line 227: prefer

blivet/formats/disklabel.py
```diff
@@ -220,12 +220,13 @@ def parted_device(self):
...
+        elif arch.is_arm():
```
cgwalters, line 228: It's confusing to me that `is_arm` means 32 bit, is this because the other arch string is `aarch64`?

Maybe worth a followup to rename `is_arm` to `is_arm32` or so

cgwalters (commented):
Looks sane to me, thanks for doing this!

For bootc we'll need to handle msdos for upgrade scenarios for the forseeable future but I think GPT is the right new default.

blivet/formats/disklabel.py
vojtechtrefny: The name is unfortunate, I had to check myself what `is_arm` actually means. We have `arch.is_arm` which means 32bit ARM and `arch.is_aarch64` for 64bit ARM.

Issue #1242, opened 2024-06-02
Get filesystem size and free space from udev when possible


vojtechtrefny:
For some filesystems blkid can now read the filesystem size and free space from the filesystem metadata and these are also available in udev:

```
E: ID_FS_BLOCKSIZE=4096
E: ID_FS_LASTBLOCK=262144
E: ID_FS_SIZE=1073741824
E: ID_FS_TYPE=ext4
```

we should use these values when possible instead of getting the information manually.

PR #1321, opened 2024-11-27
Get filesystem size from udev database

vojtechtrefny:
We can now get filesystem size for all filesystems supported by us directly from udev, no need to run the filesystem specific tools.

Fixes: #1242


vojtechtrefny:
Unfortunately we still need to keep the old code running the command line tools for CentOS 9 which doesn't have the latest version of libblkid. But on most system we should be able to get the information from udev without depending on the fallback.

Issue #1320, opened 2024-11-25
LVM resize fails with calculation error - pv metadata unaccounted for


plimptm:
I have been troubleshooting the use of this library in the ansible [linux-system-roles collection](https://github.com/linux-system-roles/storage) and have traced an issue with the resizing capabilities to what I believe to be a size calculation bug in blivet. I searched the issues and didn't see anything related, so I'm opening this here.

Note: I first encountered this bug while trying to use blivet to resize a logical volume (after resizing the underlying disk) but have been able to reproduce the issue by simply calling the `LVMLogicalVolumeDevice.resize()` method even when there is no extra space available.

Expected behaviors:

- calling the `resize` method with unallocated PEs in the parent volume group will expand the logical volume to fill those unallocated PEs
- calling the `resize` method with no unallocated PEs in the parent volume group will see that the logical volume and volume groups sizes already match and return something indicating there are no available extents

Actual behavior:

```python
>>> b = blivet.Blivet()
>>> lv = b.lvs[9]
>>> lv
LVMLogicalVolumeDevice instance (0x7f630064f880) --
  name = vg01-lv_app01  status = True  id = 354
  children = [] 
  parents = ['existing 10 GiB lvmvg vg01 (350)']
  uuid = Kgmu9m-dPFg-PdEJ-6KmH-5FPs-CRBx-Q2lVxa  size = 9.99 GiB
  format = existing ext4 filesystem
  major = 0  minor = 0  exists = True  protected = False
  sysfs path = /sys/devices/virtual/block/dm-9
  target size = 9.99 GiB  path = /dev/mapper/vg01-lv_app01
  format args = []  original_format = ext4  target = None  dm_uuid = None  VG device = LVMVolumeGroupDevice instance (0x7f62ff5018b0) --
  name = vg01  status = True  id = 350
  children = ['existing 9.99 GiB lvmlv vg01-lv_app01 (354) with existing ext4 filesystem']
  parents = ['existing 10 GiB disk sdb (341) with existing lvmpv']
  uuid = 9pDOvI-CXzn-0Xce-XRka-0TtE-Ut7E-1qJNFE  size = 10 GiB  
  format = existing None
  major = 0  minor = 0  exists = True  protected = False
  sysfs path =
  target size = 9.99 GiB  path = /dev/vg01
  format args = []  original_format = None  free = 0 B  PE Size = 4 MiB  PE Count = 2558
  PE Free = 0  PV Count = 1
[...]

PR #1334, opened 2025-01-21
LVMPV format size

vojtechtrefny:
This makes the LVM code to check the actual LVM PV format size instead of assuming it's always the same size as the underlying block device.

Fixes: #1320


vojtechtrefny:
I ran some LVM and autopart kickstart tests on this and together with the new tests I added, I am pretty sure this shouldn't break anything, so marking this a s ready for review.

PR #1335, opened 2025-01-22
Include additional information in PartitioningError

vojtechtrefny:
The generic 'Unable to allocate requested partition scheme' is not very helpful, we should try to include additional information if possible.

### DIFF
--- a/blivet/partitioning.py
+++ b/blivet/partitioning.py
@@ -34,7 +34,7 @@ from .errors import DeviceError, PartitioningError, AlignmentError
 from .flags import flags
 from .devices import Device, PartitionDevice, device_path_to_name
 from .size import Size
-from .i18n import _
+from .i18n import _, N_
 from .util import compare
 
 import logging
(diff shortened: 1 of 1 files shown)


vojtechtrefny:
Slight issue with this is that the error message can get quite long with more disks -- up to one extra error line for each disk if you specify multiple disks when creating a partition (which for example Anaconda uses to tell blivet to choose the best disk to allocate the partition). So an extreme example (8 disks all with 4 primary partitions) can look like this:

image

vojtechtrefny:
> We should show the error, but I think a generic detail message, `No free partition slots on the requested devices` is also acceptable`.

The problem with that is that we can have different reasons for different disks. Right now this PR covers only the cases when a partition cannot be added because of limits for number of partitions on give partition table, but I plan to add other error messages so it would be possible that with 4 disks some could be excluded because there isn't enough free space, one because of the primary partition number limit and one because the partition would start at sector >= 2 TiB. There is unfortunately too many reasons for a partition not being able to be placed on a disk.

KKoukiou:
@vojtechtrefny in that case you either create a more sophisticated error set, where there are pairs:
``` python
# Create a dictionary to group errors by reason
errors = {}
if reason in grouped_set:
        errors[reason].append(disk)
    else:
        errors[reason] = [disk]

# Show combined errors for disks
final_errors = [reason .format(disks)} for reason, disks in errors.items()]
msg = _("Unable to allocate requested partition scheme:\n%s") % "\n".join(final_errors)
```

Or just keep your current code, better than not error details.

jkonecny12:
I think having list of the disks would be great, however, if we can join these errors together that would be even better.

vojtechtrefny:
New version:
image

PR #1354, opened 2025-03-21
Allow using multiple "contexts" for LUKS

vojtechtrefny:
Currently only one passphrase and one key file can be specified when creating new LUKS, this change allows specifying multiple different "contexts" for when creating a new or unlocking an existing LUKS device. For now only passphrase and key file contexts are supported, but this change will also allow adding support for new types of contexts (kernel keyring, TPM etc.) in the future.

The change is made to be backwards compatible so "passphrase" can still be specified for the LUKS format, it will just internally create a new passphrase context so no changes are needed for existing code.

### DIFF
--- a/blivet/devicelibs/crypto.py
+++ b/blivet/devicelibs/crypto.py
@@ -106,6 +106,17 @@ class KeyslotContextList(object):
         else:
             return self.__contexts
 
+    def clear_contexts(self, ctype=None):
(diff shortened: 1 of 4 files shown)


vojtechtrefny:
Marking as ready for review. The anaconda tests are now passing and I also run some luks kickstart tests which also passed.

PR #1365, opened 2025-04-08
Drop parted device cache during reset v2

vojtechtrefny:
Parted caches information about about devices internally. The cache is based on device name so when recreating a device using the same name but different size (e.g. when removing RAID0 array and creating RAID1 array with the same name) the cached information, including device size, is used by parted. This means that during next reset() call parted will think the device is bigger and tries to fix the partition table on the array breaking it instead.

Resolves: rhbz#2357214


KKoukiou (approved):
I verified that it resolves rhbz#2357214

PR #1375, opened 2025-04-17
tests: Add udev trigger call after creating MD array for tests

vojtechtrefny:
When manually creating the array using mdadm sometimes one of the "legs" doesn't have the linux_raid_memeber format set in udev DB.


vojtechtrefny:
Jenkins, test this please.

PR #1397, opened 2025-06-23
Group DEVICE_TYPE_* constants in an Enum

abadger:
For other code using blivet, it is useful to know all of the device types that blivet supports. This is better accomplished by grouping all of the device type constants into an Enum type so that the consuming code can import just the Enum and then retrieve all the values it needs from there.

This code makes the Enum the canonical source for that information but also recreates the variables that have existed up to now from the new Enum.


tests/pylint/runpylint.py
```diff
@@ -27,7 +27,13 @@ def __init__(self):
...
             FalsePositive(r"Instance of 'GError' has no 'message' member"),  # overriding currently broken local pylint disable
             FalsePositive(r"No name '.*' in module 'libmount'"),
-            FalsePositive(r"Unknown option value for 'disable', expected a valid pylint message and got 'possibly-used-before-assignment'")
+            FalsePositive(r"Unknown option value for 'disable', expected a valid pylint message and got 'possibly-used-before-assignment'"),
+            # These are created dynamically so pylint doesn't detect them
+            FalsePositive(r"Module 'blivet.devicefactory' has no 'DEVICE_TYPE_[A-Z_]+' member"),
+            FalsePositive(r"No name 'DEVICE_TYPE_[A-Z_]+' in module 'blivet.devicefactory'"),
+            # Only ignore the undefined variable inside of devicefactory.py, where it
+            # is dynamically defined
+            FalsePositive(r"E0602\(undefined-variable\):blivet/devicefactory.py:.*Undefined variable 'DEVICE_TYPE_[A-Z_]+'"),
```
vojtechtrefny, line 36: This looks like a too broad false-positive, it might actually mask some real bugs. If we decide to move forward with this, I'd probably prefer to list the `DEVICE_TYPE...` constant manually in the `devicefactory` module just to make pylint happy instead of generating it.

blivet/devicefactory.py
```diff
@@ -52,14 +54,27 @@
...
-DEVICE_TYPE_BTRFS = 3
-DEVICE_TYPE_DISK = 4
-DEVICE_TYPE_LVM_THINP = 5
-DEVICE_TYPE_LVM_VDO = 6
-DEVICE_TYPE_STRATIS = 7
+
+@unique
+class DEVICE_TYPES(Enum):
+    """
+    Types of devices that blivet is aware of.
```
vojtechtrefny, line 61: These are device types supported by the device factory, not all devices blivet supports.

vojtechtrefny (commented):
Do you have a specific use case for this in Anaconda?
Blivet is generally not doing a good job with exposing supported technologies, so something like this is not a bad idea, but I think we might want to broader design to make sure we cover all the potential use cases. There's for example the `tasks.availability` module which can answer whether the technology supported by blivet is actually supported by the operating system. This is something that is also not very well exposed in the API and we might want to create a new "system" to answer both these questions. (And I don't even want to start with filesystems where "supported" can have very different meaning based on what operations you want.)
So I am not against this in general, but I think we need to talk about this more before we start to make more changes.

abadger:
> There's for example the `tasks.availability` module which can answer whether the technology supported by blivet is actually supported by the operating system. This is something that is also not very well exposed in the API and we might want to create a new "system" to answer both these questions.

Yeah, this use case would need something more because it is dynamic in nature.  It has to be calculated as the intersection of what code blivet has and the capabilities of the OS.

> (And I don't even want to start with filesystems where "supported" can have very different meaning based on what operations you want.) So I am not against this in general, but I think we need to talk about this more before we start to make more changes.

I think what makes filesystems tricky is that present consumers want to display something like `btrfs` and `xfs` to the users but talking about what the filesystem supports is really a matrix of capabilities.  For that we would need an API that makes the simple, "I just want to label these buttons in my UI" case simple but the "I need to only enable this feature when the filesystem supports this feature" possible.

Considering those two cases... I wonder if this PR belongs in the same framework as those at all?  Those are both dealing with blivet telling external code about the features and capabilities of blivet and the system it's running on.  This PR is really about easily telling calling code about the language to speak when interacting with blivet (this symbolic type name maps to this type code).  What do you think?

tests/pylint/runpylint.py
abadger: Commit that updates the code is pushed.  I haven't squashed yet so it's easier for you to see what the new changes are.

blivet/devicefactory.py
abadger: Docstr5ing updated.

blivet/dbus/blivet.py
```diff
@@ -79,11 +78,11 @@ def interface(self):
     @property
     def properties(self):
         props = {"Devices": self.ListDevices(),
-                 "DEVICE_TYPE_LVM": DEVICE_TYPE_LVM,
-                 "DEVICE_TYPE_LVM_THINP": DEVICE_TYPE_LVM_THINP,
-                 "DEVICE_TYPE_PARTITION": DEVICE_TYPE_PARTITION,
-                 "DEVICE_TYPE_MD": DEVICE_TYPE_MD,
-                 "DEVICE_TYPE_BTRFS": DEVICE_TYPE_BTRFS}
+                 "DEVICE_TYPES.LVM": DEVICE_TYPES.LVM,
```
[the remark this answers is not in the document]
abadger: Change made in: https://github.com/storaged-project/blivet/pull/1397/commits/f5752852a17f729dccabc6ba364fca09f788779d

blivet/devicefactory.py
```diff
@@ -52,14 +54,36 @@
...
-DEVICE_TYPE_MD = 1
-DEVICE_TYPE_PARTITION = 2
-DEVICE_TYPE_BTRFS = 3
-DEVICE_TYPE_DISK = 4
-DEVICE_TYPE_LVM_THINP = 5
-DEVICE_TYPE_LVM_VDO = 6
-DEVICE_TYPE_STRATIS = 7
+
+@unique
+class DEVICE_TYPES(IntEnum):
```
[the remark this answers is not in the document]
abadger: Change made in: https://github.com/storaged-project/blivet/pull/1397/commits/e6bfeda7e38b0ea5b65a41681a203b42c30faf55

abadger:
I've made the requested changes, they're the two unsquashed commits at the end for easy reviewing.  I have a feeling that CI will be broken because Fedora Infrastructure is moving copr to a new colo.  I'll take a look to see tomorrow.

vojtechtrefny (approved):

PR #1398, opened 2025-06-26
Shorten the safe_device_name length to 55 characters

pastalian:
Anaconda may use the hostname for the auto-generated vgname. However, since the hostname can exceed 55 characters, which is considered an invalid length by is_lvm_name() [1], we need to ensure that these auto-generated names are properly truncated.

[1] https://github.com/storaged-project/blivet/blob/b070f4cf918d5cb80bb4f84d1963bbdc5fbfbe3c/blivet/devicelibs/lvm.py#L273-L278

Downstream report: https://forums.almalinux.org/t/alma-linux-10-hyper-v-is-not-a-valid-name-for-this-device/6200

## Summary by Sourcery

Shorten the maximum allowed length for safe_device_name to 55 characters to prevent invalid LVM names when using long hostnames.

Bug Fixes:
- Truncate auto-generated VG names based on long hostnames to avoid exceeding LVM’s name length limits.

Enhancements:
- Reduce safe_device_name’s max_len from 96 to 55 to align with is_lvm_name() restrictions.

### DIFF
--- a/blivet/blivet.py
+++ b/blivet/blivet.py
@@ -1013,7 +1013,7 @@ class Blivet(object, metaclass=SynchronizedMeta):
     def safe_device_name(self, name, device_type=None):
         """ Convert a device name to something safe and return that.
 
-            LVM limits lv names to 128 characters. I don't know the limits for
+            LVM limits vgname + lvname to 126 characters. I don't know the limits for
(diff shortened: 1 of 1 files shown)


vojtechtrefny (approved):
Looks good to me, thank you.

PR #1464, opened 2026-04-08
Fix crash on systems with broken stratis pools

vojtechtrefny:
A broken or partially created stratis pool can be reported as stopped without a name. In this case we fail to gather the info about it and crash. We can't work with devices without a name so we need to completely ignore these pools.

Related: https://github.com/stratis-storage/project/issues/861

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation and robustness when processing stopped storage pools: the system now detects missing pool names, logs warnings, and skips creating or registering pool devices when required metadata is absent, preventing invalid configurations and related errors.

### DIFF
--- a/blivet/populator/helpers/stratis.py
+++ b/blivet/populator/helpers/stratis.py
@@ -176,6 +176,11 @@ class StratisFormatPopulator(FormatPopulator):
             # possibly stopped pool
             pool_info = next((pool_info for pool_info in stratis_info.stopped_pools if self.device.path in pool_info.devices), None)
             if pool_info:
+                if not pool_info.name:
+                    # no name -> probably broken or partially constructed pool
+                    # we cannot add a device without name
+                    log.warning("Ignoring stopped stratis pool %s without name", pool_info.uuid)
+                    return
                 pool_device = self._add_stopped_pool_device(pool_info)
             else:
                 # no stopped pool info, no bd info -> nothing we can do
--- a/blivet/static_data/stratis_info.py
+++ b/blivet/static_data/stratis_info.py
@@ -211,8 +211,14 @@ class StratisInfo(object):
                 else:
                     encrypted = bool(feats["encryption"])
 
+            if "name" in pools_info[pool_uuid].keys():
+                pool_name = pools_info[pool_uuid]["name"].get_string()
+            else:
+                log.warning("Stopped Stratis pool %s does not have name", pool_uuid)
+                pool_name = None
+
             info = StratisStoppedPoolInfo(uuid=pool_uuid,
-                                          name=pools_info[pool_uuid]["name"].get_string(),
+                                          name=pool_name,
                                           devices=[d["devnode"] for d in pools_info[pool_uuid]["devs"]],
                                           encrypted=encrypted)
             stopped_pools.append(info)


blivet/static_data/stratis_info.py
```diff
@@ -211,8 +211,14 @@ def _get_stopped_pools_info(self):
...
+            else:
+                log.warning("Stopped Stratis pool %s does not have name", pool_uuid)
+                pool_name = None
+
             info = StratisStoppedPoolInfo(uuid=pool_uuid,
-                                          name=pools_info[pool_uuid]["name"].get_string(),
+                                          name=pool_name,
                                           devices=[d["devnode"] for d in pools_info[pool_uuid]["devs"]],
                                           encrypted=encrypted)
             stopped_pools.append(info)
```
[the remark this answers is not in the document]
vojtechtrefny: These pools are not added to the devicetree, but we still want the information about their existence in the static data because there is "something" in the system with this UUID.

blivet/static_data/stratis_info.py
```diff
@@ -211,8 +211,14 @@ def _get_stopped_pools_info(self):
                 else:
                     encrypted = bool(feats["encryption"])
 
+            if "name" in pools_info[pool_uuid].keys():
+                pool_name = pools_info[pool_uuid]["name"].get_string()
+            else:
+                log.warning("Stopped Stratis pool %s does not have name", pool_uuid)
+                pool_name = None
```
tasleson, line 218: Previously, the namedtuple `StratisStoppedPoolInfo.pool_name` had a type of `string`.  With this change it can now be `None`.  This could possibly cause a user of blivet to run into an error if they assume it will always be a valid string.  Thus, it's kind of a subtle public API change.  One could argue that changing this to an empty string would be more appropriate as it preserves the data type.  However, the counter argument could be made that `None` forces caller to specifically address when their code blows up instead of silently displaying an empty string.

blivet/populator/helpers/stratis.py
```diff
@@ -176,6 +176,11 @@ def _add_pool_device(self):
             # possibly stopped pool
             pool_info = next((pool_info for pool_info in stratis_info.stopped_pools if self.device.path in pool_info.devices), None)
             if pool_info:
+                if not pool_info.name:
+                    # no name -> probably broken or partially constructed pool
+                    # we cannot add a device without name
+                    log.warning("Ignoring stopped stratis pool without name")
```
tasleson: Please add the pool uuid to the warning, so that someone debugging can know which pool this is, for correlation.

```python
log.warning("Ignoring stopped stratis pool %s without name", pool_info.uuid))
```

tasleson (commented):
My initial reaction was `LGTM`, but after a more detailed review, I think there is some ideas that should be considered before committing which I've noted in the review.

Also, what is the state of the `devices` key?  Does it have valid entries or empties when this occurs?  Do we need code guards for this or is it just the name that is absent?  Do we want to be able to handle total garbage in the future if stratis breaks in different ways?

I also think it would be helpful to reference the stratis bug issue in the commit message.  So that someone visiting this commit in the future can see what happened with stratis.  I'm assuming that there is a plan for fixing it.

blivet/static_data/stratis_info.py
vojtechtrefny: Fortunately the `static_data` module is not part of the [public API](https://storaged.org/blivet/docs/api.html).

blivet/populator/helpers/stratis.py
vojtechtrefny: Good point, fixed.

PR #1483, opened 2026-06-09
Fix original_format not set on MD arrays due to sysfs race

vojtechtrefny:
When the MDRaidMember populator helper creates an MD array, it calls handle_device(update_orig_fmt=True) to detect and record the array's format as original_format. However, this call is gated on md_array.status which reads md/array_state from sysfs. Due to a race condition, the array state may not yet be active, causing the call to be skipped entirely.

When the main scan loop later processes the array device, it calls handle_device without update_orig_fmt, so original_format is never updated from its initial empty DeviceFormat(None).

This causes failures during device destruction: the LUKS format teardown (which calls cryptsetup close) relies on original_format to close the dm-crypt mapping before mdadm --stop runs. With original_format stuck at None, the LUKS device is never closed, and mdadm --stop fails with "Cannot get exclusive access".

Update the condition in handle_device to also set original_format when it was never properly initialized (type is None) and a real existing format has been detected by handle_format.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device handling so the app preserves the correct original format even when format detection happens later than device discovery.
  * Prevents incomplete or missing format snapshots from causing incorrect format history, especially in race-prone storage setups.
  * Added coverage for late-detected formats to ensure the stored original format is set correctly and not overwritten once established.

### DIFF
--- a/blivet/populator/populator.py
+++ b/blivet/populator/populator.py
@@ -323,7 +323,7 @@ class PopulatorMixin(object, metaclass=SynchronizedMeta):
 
         # now handle the device's formatting
         self.handle_format(info, device)
-        if device_added or update_orig_fmt:
+        if device_added or update_orig_fmt or (not device.original_format.type and device.format.type and device.format.exists):
             device.original_format = copy.deepcopy(device.format)
         device.device_links = udev.device_get_symlinks(info)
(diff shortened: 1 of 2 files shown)


vojtechtrefny:
@coderabbitai review

Issue #1500, opened 2026-07-22
Deadlock between blivet_lock and logging handler lock due to @exclusive wrapping __str__


zhaoyonghao10:
Problem                                                                                                                                                                                             
                                                                                                                                                                                                      
  When running Anaconda with automatic partitioning, a deadlock can occur between the Storage DBus service's main thread and a background task thread. The deadlock involves two locks acquired in    
  opposite order:                                                                                                                                                                                   
                                                                                                                                                                                                      
  1. blivet_lock (global RLock in blivet/threads.py)                                                                                                                                                  
  2. logging handler's self.lock (RLock in logging/__init__.py)
                                                                                                                                                                                                      
  Tracebacks                                                                                                                                                                                          
                                                                                                                                                                                                      
  Thread 1 — GLib MainLoop (DBus service thread):                                                                                                                                                     
                                                                                                                                                                                                    
  Thread 1 holds: blivet_lock                                                                                                                                                                         
  Thread 1 waits: handler.lock                                                                                                                                                                        
                                                                                                                                                                                                      
  ```
 Traceback (most recent call first):
File /usr/lib64/python3.1i/logging/_init_.py", line 927, in acquire
 self.lock,acquire()
File/usr/lib64/python3.11/logging/_ init_.py", line 976, in handle
 self,acquire()
 File/usr/lib64/python3.11/logging/_ init_.py", line 1706, in callHandlers
 hdlr.handle(record)
File */usr/lib64/python3.11/logging/_ init__.py", line 1644, in handle
 self.callHandlers(record)
File*/usr/lib64/python3.1i/logging/_init_.py", line 1634, in _log
 self.handle(record)
File /usr/lib64/python3.11/logging/_ init_py", line 1477, in debug 
self.log(DEBUG, msg, args, kwargs)
 File "/usr/lib/python3.11/site-packages/blivet/storage_log.py", line 46, in log_method_call
log.debug(fmt, *fmt_args)
File/usr/lib/python3.11/site-packages/blivet/devicetree.py", line 551, in get_device_by_name
logmethod call(self,name=name,incompleteincomlete,hidden=hien)
File*/usr/lib/python3.11/site-packages/blivet/threads.py", line 53, in run_with_lock
return m(*args, **kwargs)
Filev/usr/lib64/python3.11/site-packages/pyanaconda/modules/storage/devicetree/viewer.py", line 236, in _get_device
device = self.storage.devicetree.get_device by name(
[...]

PR #1502, opened 2026-07-24
Fix deadlock between blivet_lock and logging lock

century6:
When installing openEuler via Anaconda, the process `python3 -m pyanaconda.modules.storage` can hang due to a classic lock ordering violation deadlock between two threads:

Thread 1 (D-Bus service):
  Holds blivet_lock -> waits for logging lock
  (get_device_by_name holds blivet_lock via the exclusive decorator,
  then calls log_method_call/log.debug which acquires the logging
  handler's lock)

Thread 2 (task worker):
  Holds logging lock -> waits for blivet_lock
  (device_matches calls log.debug which acquires the logging lock,
  then formats the message via msg % self.args, which triggers
  \_\_str\_\_ on a devicetree argument that internally calls run_with_lock
  to acquire blivet_lock)

The root cause is that SynchronizedMeta wraps ALL methods and properties with the exclusive decorator, including \_\_str\_\_ which is a read-only operation that does not modify any state. When the logging framework holds its lock and defers string formatting to emit time, calling \_\_str\_\_ on a DeviceTree/Device object acquires blivet_lock, creating a circular dependency with code that holds blivet_lock first and then logs.

Fix this by adding __str__ and its read-only call chain to _unsynchronized_methods for both DeviceTreeBase and Device, so that these methods do not acquire blivet_lock. This is safe because:

- All exempted methods/properties are purely read-only operations
- Write operations on properties (e.g. children, parents setters) are still called from within synchronized methods that already hold the lock
- The worst case without the lock is reading a slightly stale snapshot, which is acceptable for \_\_str\_\_ (used only for debugging/logging)

#1500 

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented potential deadlocks when converting device details to strings and when rendering the device tree.
  * Improved reliability of device string details by consistently using safe access to device identity and relationship information.
  * Enhanced device tree rendering by more accurately detecting subtree roots based on current parent relationships.

### DIFF
--- a/blivet/devices/device.py
+++ b/blivet/devices/device.py
@@ -75,8 +75,8 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
     # deadlock when __str__ is called during logging format (which holds
     # the logging handler lock). These are safe to call without the lock
     # since they only read state and do not modify it.
-    _unsynchronized_methods = ['__str__', '_to_string', '_get_name',
-                               'type', '_get_parents', 'children']
+    _unsynchronized_methods = ['__str__', '_to_string', '_get_parents',
+                               'children']
 
     def __init__(self, name, parents=None):
         """
@@ -119,7 +119,7 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
 
     # Force str and unicode types in case type or name is unicode
     def _to_string(self):
-        s = "%s %s (%d)" % (self.type, self._get_name(), self.id)
+        s = "%s %s (%d)" % (self._type, self._name, self.id)
         return s
 
     def __str__(self):
--- a/blivet/devices/device.py
+++ b/blivet/devices/device.py
@@ -71,6 +71,13 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
     _packages = []
     _external_dependencies = []
 
+    # Read-only methods/properties exempted from blivet_lock to prevent
+    # deadlock when __str__ is called during logging format (which holds
+    # the logging handler lock). These are safe to call without the lock
+    # since they only read state and do not modify it.
+    _unsynchronized_methods = ['__str__', '_to_string', '_get_name',
+                               'type', '_get_parents', 'children']
+
     def __init__(self, name, parents=None):
         """
             :param name: the device name (generally a device node's basename)
@@ -112,7 +119,7 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
 
     # Force str and unicode types in case type or name is unicode
     def _to_string(self):
-        s = "%s %s (%d)" % (self.type, self.name, self.id)
+        s = "%s %s (%d)" % (self.type, self._get_name(), self.id)
         return s
 
     def __str__(self):
@@ -155,6 +162,9 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
         self._init_parent_list()
         for parent in parents:
             self._parents.append(parent)
+
+    def _get_parents(self):
+        return self._parents[:]
 
     @property
     def children(self):
--- a/blivet/devicetree.py
+++ b/blivet/devicetree.py
@@ -72,6 +72,9 @@ class DeviceTreeBase(object, metaclass=SynchronizedMeta):
         :class:`~.deviceaction.DeviceAction` instances can only be registered
         for leaf devices, except for resize actions.
     """
+
+    _unsynchronized_methods = ['__str__']
+
     def __init__(self, ignored_disks=None, exclusive_disks=None):
         """
             :keyword ignored_disks: ignored disks
@@ -122,7 +125,7 @@ class DeviceTreeBase(object, metaclass=SynchronizedMeta):
                     s += show_subtree(child, depth + 1)
             return s
 
-        roots = [d for d in self._devices if not d.parents]
+        roots = [d for d in self._devices if not d._get_parents()]
         tree = ""
         for root in roots:
             tree += show_subtree(root, 0)


century6:
@coderabbitai review

century6:
@vojtechtrefny Could you take a look when you have time?

century6:
Friendly reminder on this PR. It's been a week since I opened it, and I'm keen to get some initial feedback to make sure I'm on the right track. I'm happy to make any adjustments needed. Could a maintainer or committer please review when possible? Thanks in advance!
@vojtechtrefny @tasleson @tbzatek

vojtechtrefny:
Jenkins, ok to test.

vojtechtrefny:
I'll try to look at this next week. Sorry for delays.

tasleson:
Thanks for the thorough deadlock analysis, the problem description is excellent and the fix works for the base Device class. However, I think there's a gap with subclasses that override type as a property.

 The issue: SynchronizedMeta.__new__ only checks _unsynchronized_methods from the class being defined (dct), not from parent classes. So when LVMLogicalVolumeDevice (and several LVM mixins) override type as a @property, that override gets wrapped with exclusive() because the subclass doesn't have its own _unsynchronized_methods. The call chain __str__ → _to_string → self.type still acquires blivet_lock on those subclasses.

  Here's a test that demonstrates both cases. you can drop it in as tests/unit_tests/deadlock_test.py:

```python
  """
  Demonstrate the deadlock between blivet_lock and Python's logging lock.

  The deadlock is a lock-ordering inversion between two threads:

    Thread A: holds blivet_lock -> calls log.debug() -> needs logging lock
    Thread B: holds logging lock -> handler calls str(device) -> needs blivet_lock

  test_device_str_logging_deadlock:
    Uses the base Device class.  FAILS on unfixed code, PASSES with the
    _unsynchronized_methods fix.

  test_subclass_type_override_deadlock:
    Uses a Device subclass that overrides `type` as a @property (the way
    LVMLogicalVolumeDevice does).  FAILS even WITH the fix, because
    SynchronizedMeta wraps the subclass's `type` property independently --
    the parent's _unsynchronized_methods list is not consulted.
  """

  import logging
  import subprocess
  import sys
  import textwrap
  import threading
  import unittest

  TIMEOUT = 5


  def _run_deadlock_scenario(device_class_path):
      """Run the deadlock scenario in-process.

      :param device_class_path: dotted path to a Device class, e.g.
          "blivet.devices.device.Device" or
          "tests.unit_tests.deadlock_test.SubclassWithTypeOverride".
      :returns: True if it completed without deadlocking.
      """
      import importlib

      from blivet.threads import blivet_lock

      mod_path, cls_name = device_class_path.rsplit(".", 1)
      mod = importlib.import_module(mod_path)
      device_cls = getattr(mod, cls_name)

      dev = device_cls("testdev")

      # Two events coordinate the threads so they both hold their
      # respective locks before attempting the conflicting acquisition.
      thread_a_has_lock = threading.Event()
      thread_b_in_emit = threading.Event()

      class DeviceStrHandler(logging.Handler):
          """Handler whose first emit() call synchronises with Thread A
          then calls str(dev), reproducing the deadlock window.
          Subsequent emit() calls (re-entry from Thread A) are no-ops."""

          def __init__(self):
              super().__init__()
              self._first = True

          def emit(self, record):
              if not self._first:
                  return
              self._first = False
              # We are inside Handler.handle(), which holds self.lock.
              # Signal Thread A that we are in position.
              thread_b_in_emit.set()
              # Wait until Thread A confirms it holds blivet_lock.
              if not thread_a_has_lock.wait(timeout=TIMEOUT):
                  return
              # This is the critical call: str(dev) calls Device.__str__
              # which, if wrapped by SynchronizedMeta, tries to acquire
              # blivet_lock -- deadlock.
              str(dev)

      logger = logging.getLogger("blivet.deadlock_test")
      logger.setLevel(logging.DEBUG)
      handler = DeviceStrHandler()
      logger.addHandler(handler)

      def thread_a():
          with blivet_lock:
              thread_a_has_lock.set()
              # Wait until Thread B is inside emit (holding handler lock).
              if not thread_b_in_emit.wait(timeout=TIMEOUT):
                  return
              # This tries to acquire handler.lock -- Thread B holds it.
              logger.debug("thread_a: holding blivet_lock, now logging")

      def thread_b():
          # Enters Handler.handle() which acquires handler.lock, then
          # calls emit() where synchronization happens.
          logger.debug("thread_b: triggers DeviceStrHandler.emit")

      ta = threading.Thread(target=thread_a, daemon=True)
      tb = threading.Thread(target=thread_b, daemon=True)
      ta.start()
      tb.start()
      ta.join(timeout=TIMEOUT)
      tb.join(timeout=TIMEOUT)

      logger.removeHandler(handler)
      return not (ta.is_alive() or tb.is_alive())


  # ---------------------------------------------------------------------------
  # A minimal Device subclass that overrides `type` as a @property, exactly
  # the way LVMLogicalVolumeDevice (and several LVM mixin classes) do.
  # SynchronizedMeta wraps this property with exclusive() because the
  # subclass does NOT declare its own _unsynchronized_methods.
  # ---------------------------------------------------------------------------
  from blivet.devices.device import Device            # noqa: E402
  from blivet.threads import SynchronizedMeta         # noqa: E402


  class SubclassWithTypeOverride(Device, metaclass=SynchronizedMeta):
      """Mimics LVMLogicalVolumeDevice's pattern of overriding `type`."""

      @property
      def type(self):
          return "fake_subclass"

      thread_b_in_emit = threading.Event()

      class DeviceStrHandler(logging.Handler):
          """Handler whose first emit() call synchronises with Thread A
          then calls str(dev), reproducing the deadlock window.
          Subsequent emit() calls (re-entry from Thread A) are no-ops."""

          def __init__(self):
              super().__init__()
              self._first = True

          def emit(self, record):
              if not self._first:
                  return
              self._first = False
              # We are inside Handler.handle(), which holds self.lock.
              # Signal Thread A that we are in position.
              thread_b_in_emit.set()
              # Wait until Thread A confirms it holds blivet_lock.
              if not thread_a_has_lock.wait(timeout=TIMEOUT):
                  return
              # This is the critical call: str(dev) calls Device.__str__
              # which, if wrapped by SynchronizedMeta, tries to acquire
              # blivet_lock -- deadlock.
              str(dev)

      logger = logging.getLogger("blivet.deadlock_test")
      logger.setLevel(logging.DEBUG)
      handler = DeviceStrHandler()
      logger.addHandler(handler)

      def thread_a():
          with blivet_lock:
              thread_a_has_lock.set()
              # Wait until Thread B is inside emit (holding handler lock).
              if not thread_b_in_emit.wait(timeout=TIMEOUT):
                  return
              # This tries to acquire handler.lock -- Thread B holds it.
              logger.debug("thread_a: holding blivet_lock, now logging")

      def thread_b():
          # Enters Handler.handle() which acquires handler.lock, then
          # calls emit() where synchronization happens.
          logger.debug("thread_b: triggers DeviceStrHandler.emit")

      ta = threading.Thread(target=thread_a, daemon=True)
      tb = threading.Thread(target=thread_b, daemon=True)
      ta.start()
      tb.start()
      ta.join(timeout=TIMEOUT)
      tb.join(timeout=TIMEOUT)

      logger.removeHandler(handler)
      return not (ta.is_alive() or tb.is_alive())


  # ---------------------------------------------------------------------------
  # A minimal Device subclass that overrides `type` as a @property, exactly
  # the way LVMLogicalVolumeDevice (and several LVM mixin classes) do.
  # SynchronizedMeta wraps this property with exclusive() because the
  # subclass does NOT declare its own _unsynchronized_methods.
  # ---------------------------------------------------------------------------
  from blivet.devices.device import Device            # noqa: E402
  from blivet.threads import SynchronizedMeta         # noqa: E402


  class SubclassWithTypeOverride(Device, metaclass=SynchronizedMeta):
      """Mimics LVMLogicalVolumeDevice's pattern of overriding `type`."""

      @property
      def type(self):
          return "fake_subclass"


  # ---------------------------------------------------------------------------

  DEADLOCK_MSG = (
      "Device.__str__ under SynchronizedMeta causes a lock-ordering "
      "inversion with Python's logging handler lock."
  )


  def _subprocess_deadlock_test(device_class_path):
      """Spawn a subprocess that runs the deadlock scenario and return
      (passed: bool, detail: str)."""
      script = textwrap.dedent(f"""\
          import sys
          sys.path.insert(0, ".")
          from tests.unit_tests.deadlock_test import _run_deadlock_scenario
          ok = _run_deadlock_scenario("{device_class_path}")
          sys.exit(0 if ok else 1)
      """)

      try:
          result = subprocess.run(
              [sys.executable, "-c", script],
              timeout=TIMEOUT * 3,
              capture_output=True,
              text=True,
          )
      except subprocess.TimeoutExpired:
          return False, f"subprocess did not exit within {TIMEOUT * 3}s"

      if result.returncode != 0:
          return False, f"threads timed out.  stderr: {result.stderr}"

      return True, ""


  class DeadlockTest(unittest.TestCase):

      def test_device_str_logging_deadlock(self):
          """str(Device) inside a logging handler must not deadlock."""
          ok, detail = _subprocess_deadlock_test(
              "blivet.devices.device.Device"
          )
          if not ok:
              self.fail(f"Deadlock (base Device): {detail}\n{DEADLOCK_MSG}")

      def test_subclass_type_override_deadlock(self):
          """str() on a subclass that overrides `type` must not deadlock.

          LVMLogicalVolumeDevice (and several LVM mixins) override `type`
          as a @property in their own class body.  SynchronizedMeta wraps
          that override with exclusive() independently of the parent's
          _unsynchronized_methods list.  So _to_string -> self.type still
          acquires blivet_lock on those subclasses, re-introducing the
          deadlock even after __str__ itself is exempted.
          """
          ok, detail = _subprocess_deadlock_test(
              "tests.unit_tests.deadlock_test.SubclassWithTypeOverride"
          )
          if not ok:
              self.fail(
                  f"Deadlock (subclass with type override): {detail}\n"
                  f"{DEADLOCK_MSG}\n"
                  "The _unsynchronized_methods exemption on Device does not "
                  "propagate to subclass property overrides."
              )
```

With your current fix applied, test_device_str_logging_deadlock passes but test_subclass_type_override_deadlock still deadlocks.

  Suggested fix: In _to_string, access the backing fields directly instead of going through property accessors:

```python
  def _to_string(self):
      s = "%s %s (%d)" % (self._type, self._name, self.id)
      return s
 ```

This avoids blivet_lock regardless of what subclasses do with their properties. The trade-off is that LVM devices with mixin-based type overrides would show their base _type (e.g. "lvmlv" instead of "lvmthinpool") in __str__ output, but that seems fine for logging/debugging.

**Comment by century6** [comment]
> Thanks for the thorough deadlock analysis, the problem description is excellent and the fix works for the base Device class. However, I think there's a gap with subclasses that override type as a property.
> 
> The issue: SynchronizedMeta.**new** only checks _unsynchronized_methods from the class being defined (dct), not from parent classes. So when LVMLogicalVolumeDevice (and several LVM mixins) override type as a @Property, that override gets wrapped with exclusive() because the subclass doesn't have its own _unsynchronized_methods. The call chain **str** → _to_string → self.type still acquires blivet_lock on those subclasses.
> 
> Here's a test that demonstrates both cases. you can drop it in as tests/unit_tests/deadlock_test.py:
> 
> ```python
>   """
>   Demonstrate the deadlock between blivet_lock and Python's logging lock.
> 
>   The deadlock is a lock-ordering inversion between two threads:
> 
>     Thread A: holds blivet_lock -> calls log.debug() -> needs logging lock
>     Thread B: holds logging lock -> handler calls str(device) -> needs blivet_lock
> 
>   test_device_str_logging_deadlock:
>     Uses the base Device class.  FAILS on unfixed code, PASSES with the
>     _unsynchronized_methods fix.
> 
>   test_subclass_type_override_deadlock:
>     Uses a Device subclass that overrides `type` as a @property (the way
>     LVMLogicalVolumeDevice does).  FAILS even WITH the fix, because
>     SynchronizedMeta wraps the subclass's `type` property independently --
>     the parent's _unsynchronized_methods list is not consulted.
>   """
> 
>   import logging
>   import subprocess
>   import sys
>   import textwrap
>   import threading
>   import unittest
> 
>   TIMEOUT = 5
> 
> 
>   def _run_deadlock_scenario(device_class_path):
>       """Run the deadlock scenario in-process.
> 
>       :param device_class_path: dotted path to a Device class, e.g.
>           "blivet.devices.device.Device" or
>           "tests.unit_tests.deadlock_test.SubclassWithTypeOverride".
>       :returns: True if it completed without deadlocking.
>       """
>       import importlib
> 
>       from blivet.threads import blivet_lock
> 
>       mod_path, cls_name = device_class_path.rsplit(".", 1)
>       mod = importlib.import_module(mod_path)
>       device_cls = getattr(mod, cls_name)
> 
>       dev = device_cls("testdev")
> 
>       # Two events coordinate the threads so they both hold their
>       # respective locks before attempting the conflicting acquisition.
>       thread_a_has_lock = threading.Event()
>       thread_b_in_emit = threading.Event()
> 
>       class DeviceStrHandler(logging.Handler):
>           """Handler whose first emit() call synchronises with Thread A
>           then calls str(dev), reproducing the deadlock window.
>           Subsequent emit() calls (re-entry from Thread A) are no-ops."""
> 
>           def __init__(self):
>               super().__init__()
>               self._first = True
> 
>           def emit(self, record):
>               if not self._first:
>                   return
>               self._first = False
>               # We are inside Handler.handle(), which holds self.lock.
>               # Signal Thread A that we are in position.
>               thread_b_in_emit.set()
>               # Wait until Thread A confirms it holds blivet_lock.
>               if not thread_a_has_lock.wait(timeout=TIMEOUT):
>                   return
>               # This is the critical call: str(dev) calls Device.__str__
>               # which, if wrapped by SynchronizedMeta, tries to acquire
>               # blivet_lock -- deadlock.
>               str(dev)
> 
>       logger = logging.getLogger("blivet.deadlock_test")
>       logger.setLevel(logging.DEBUG)
>       handler = DeviceStrHandler()
>       logger.addHandler(handler)
> 
>       def thread_a():
>           with blivet_lock:
>               thread_a_has_lock.set()
>               # Wait until Thread B is inside emit (holding handler lock).
>               if not thread_b_in_emit.wait(timeout=TIMEOUT):
>                   return
>               # This tries to acquire handler.lock -- Thread B holds it.
>               logger.debug("thread_a: holding blivet_lock, now logging")
> 
>       def thread_b():
>           # Enters Handler.handle() which acquires handler.lock, then
>           # calls emit() where synchronization happens.
>           logger.debug("thread_b: triggers DeviceStrHandler.emit")
> 
>       ta = threading.Thread(target=thread_a, daemon=True)
>       tb = threading.Thread(target=thread_b, daemon=True)
>       ta.start()
>       tb.start()
>       ta.join(timeout=TIMEOUT)
>       tb.join(timeout=TIMEOUT)
> 
>       logger.removeHandler(handler)
>       return not (ta.is_alive() or tb.is_alive())
> 
> 
>   # ---------------------------------------------------------------------------
>   # A minimal Device subclass that overrides `type` as a @property, exactly
>   # the way LVMLogicalVolumeDevice (and several LVM mixin classes) do.
>   # SynchronizedMeta wraps this property with exclusive() because the
>   # subclass does NOT declare its own _unsynchronized_methods.
>   # ---------------------------------------------------------------------------
>   from blivet.devices.device import Device            # noqa: E402
>   from blivet.threads import SynchronizedMeta         # noqa: E402
> 
> 
>   class SubclassWithTypeOverride(Device, metaclass=SynchronizedMeta):
>       """Mimics LVMLogicalVolumeDevice's pattern of overriding `type`."""
> 
>       @property
>       def type(self):
>           return "fake_subclass"
> 
>       thread_b_in_emit = threading.Event()
> 
>       class DeviceStrHandler(logging.Handler):
>           """Handler whose first emit() call synchronises with Thread A
>           then calls str(dev), reproducing the deadlock window.
>           Subsequent emit() calls (re-entry from Thread A) are no-ops."""
> 
>           def __init__(self):
>               super().__init__()
>               self._first = True
> 
>           def emit(self, record):
>               if not self._first:
>                   return
>               self._first = False
>               # We are inside Handler.handle(), which holds self.lock.
>               # Signal Thread A that we are in position.
>               thread_b_in_emit.set()
>               # Wait until Thread A confirms it holds blivet_lock.
>               if not thread_a_has_lock.wait(timeout=TIMEOUT):
>                   return
>               # This is the critical call: str(dev) calls Device.__str__
>               # which, if wrapped by SynchronizedMeta, tries to acquire
>               # blivet_lock -- deadlock.
>               str(dev)
> 
>       logger = logging.getLogger("blivet.deadlock_test")
>       logger.setLevel(logging.DEBUG)
>       handler = DeviceStrHandler()
>       logger.addHandler(handler)
> 
>       def thread_a():
>           with blivet_lock:
>               thread_a_has_lock.set()
>               # Wait until Thread B is inside emit (holding handler lock).
>               if not thread_b_in_emit.wait(timeout=TIMEOUT):
>                   return
>               # This tries to acquire handler.lock -- Thread B holds it.
>               logger.debug("thread_a: holding blivet_lock, now logging")
> 
>       def thread_b():
>           # Enters Handler.handle() which acquires handler.lock, then
>           # calls emit() where synchronization happens.
>           logger.debug("thread_b: triggers DeviceStrHandler.emit")
> 
>       ta = threading.Thread(target=thread_a, daemon=True)
>       tb = threading.Thread(target=thread_b, daemon=True)
>       ta.start()
>       tb.start()
>       ta.join(timeout=TIMEOUT)
>       tb.join(timeout=TIMEOUT)
> 
>       logger.removeHandler(handler)
>       return not (ta.is_alive() or tb.is_alive())
> 
> 
>   # ---------------------------------------------------------------------------
>   # A minimal Device subclass that overrides `type` as a @property, exactly
>   # the way LVMLogicalVolumeDevice (and several LVM mixin classes) do.
>   # SynchronizedMeta wraps this property with exclusive() because the
>   # subclass does NOT declare its own _unsynchronized_methods.
>   # ---------------------------------------------------------------------------
>   from blivet.devices.device import Device            # noqa: E402
>   from blivet.threads import SynchronizedMeta         # noqa: E402
> 
> 
>   class SubclassWithTypeOverride(Device, metaclass=SynchronizedMeta):
>       """Mimics LVMLogicalVolumeDevice's pattern of overriding `type`."""
> 
>       @property
>       def type(self):
>           return "fake_subclass"
> 
> 
>   # ---------------------------------------------------------------------------
> 
>   DEADLOCK_MSG = (
>       "Device.__str__ under SynchronizedMeta causes a lock-ordering "
>       "inversion with Python's logging handler lock."
>   )
> 
> 
>   def _subprocess_deadlock_test(device_class_path):
>       """Spawn a subprocess that runs the deadlock scenario and return
>       (passed: bool, detail: str)."""
>       script = textwrap.dedent(f"""\
>           import sys
>           sys.path.insert(0, ".")
>           from tests.unit_tests.deadlock_test import _run_deadlock_scenario
>           ok = _run_deadlock_scenario("{device_class_path}")
>           sys.exit(0 if ok else 1)
>       """)
> 
>       try:
>           result = subprocess.run(
>               [sys.executable, "-c", script],
>               timeout=TIMEOUT * 3,
>               capture_output=True,
>               text=True,
>           )
>       except subprocess.TimeoutExpired:
>           return False, f"subprocess did not exit within {TIMEOUT * 3}s"
> 
>       if result.returncode != 0:
>           return False, f"threads timed out.  stderr: {result.stderr}"
> 
>       return True, ""
> 
> 
>   class DeadlockTest(unittest.TestCase):
> 
>       def test_device_str_logging_deadlock(self):
>           """str(Device) inside a logging handler must not deadlock."""
>           ok, detail = _subprocess_deadlock_test(
>               "blivet.devices.device.Device"
>           )
>           if not ok:
>               self.fail(f"Deadlock (base Device): {detail}\n{DEADLOCK_MSG}")
> 
>       def test_subclass_type_override_deadlock(self):
>           """str() on a subclass that overrides `type` must not deadlock.
> 
>           LVMLogicalVolumeDevice (and several LVM mixins) override `type`
>           as a @property in their own class body.  SynchronizedMeta wraps
>           that override with exclusive() independently of the parent's
>           _unsynchronized_methods list.  So _to_string -> self.type still
>           acquires blivet_lock on those subclasses, re-introducing the
>           deadlock even after __str__ itself is exempted.
>           """
>           ok, detail = _subprocess_deadlock_test(
>               "tests.unit_tests.deadlock_test.SubclassWithTypeOverride"
>           )
>           if not ok:
>               self.fail(
>                   f"Deadlock (subclass with type override): {detail}\n"
>                   f"{DEADLOCK_MSG}\n"
>                   "The _unsynchronized_methods exemption on Device does not "
>                   "propagate to subclass property overrides."
>               )
> ```
> 
> With your current fix applied, test_device_str_logging_deadlock passes but test_subclass_type_override_deadlock still deadlocks.
> 
> Suggested fix: In _to_string, access the backing fields directly instead of going through property accessors:
> 
> ```python
>   def _to_string(self):
>       s = "%s %s (%d)" % (self._type, self._name, self.id)
>       return s
> ```
> 
> This avoids blivet_lock regardless of what subclasses do with their properties. The trade-off is that LVM devices with mixin-based type overrides would show their base _type (e.g. "lvmlv" instead of "lvmthinpool") in **str** output, but that seems fine for logging/debugging.

Thanks for the suggestion! I've made the changes in a new commit.

**Comment by century6** [comment]
Please let me know if there are any further changes or clarifications needed. I am more than happy to make additional modifications or provide any further information if required. : )
@vojtechtrefny @tasleson

**Comment by century6** [comment]
ping @vojtechtrefny @tasleson

File: blivet/devices/device.py [hunk 1]
```diff
@@ -71,6 +71,13 @@ class Device(util.ObjectID, metaclass=SynchronizedMeta):
     _packages = []
     _external_dependencies = []
 
+    # Read-only methods/properties exempted from blivet_lock to prevent
+    # deadlock when __str__ is called during logging format (which holds
+    # the logging handler lock). These are safe to call without the lock
+    # since they only read state and do not modify it.
+    _unsynchronized_methods = ['__str__', '_to_string', '_get_name',
+                               'type', '_get_parents', 'children']
```
vojtechtrefny: So `_get_name` and `_type` are no longer needed (replaced by using `_type` and `_name`), correct?

century6: I had considered removing `_get_name` and `type` from `_unsynchronized_methods`, but kept them for safety reasons:
- They might be called elsewhere during logging formatting
- Keeping read-only methods in the unsynchronized list has no side effects
﻿

However, since `_to_string` no longer uses them, removing them is also reasonable.
I'm fine with either approach. What do you prefer?

vojtechtrefny: I'd prefer keeping the list as short as possible, thanks.

century6: OK, I've updated the commit of [bd6681c](https://github.com/storaged-project/blivet/pull/1502/changes/bd6681c885feb4edff0c4c9c6ad9e7b5becaf7ff) in this PR. : )

vojtechtrefny: Thank you!

vojtechtrefny (approved):

